PR #17, opened 2020-03-25
feat: suggest UI5 namespaces in xmlns attributes values

allaVolkov:


bd82:
The test needs to be part of this PR for the CHANGELOG to be updated as expected

allaVolkov:
integration test added

PR #33, opened 2020-04-01
test: add library version 1.71.14

tal-sapan:
- Add script for downloading ui5 resources for version
- Output validation errors in tests that check the consistency of ui5 libs
- Enhance the json schema to support 1.71.14
- Manual fix for class sap.ui.vk.BaseNodeProxy in library sap.ui.vk - remove the same class from its "implements" (as this fails in strict mode)


tal-sapan:
Wanted to add this before changing the library version and adding the dist layer libraries

test-packages/test-utils/src/utils/download-ui5-resources.ts
```diff
@@ -0,0 +1,115 @@
...
+// It is run from the download-resources script in the package.json.
+// To get a different version change the version number in the last line of this file.
+import { TestModelVersion } from "../../api";
+import { reduce, forEach, keys } from "lodash";
+import { dirname, resolve } from "path";
+import { writeFileSync, mkdirsSync, readFileSync } from "fs-extra";
+import fetch from "node-fetch";
+
+async function getLibs(): Promise<string[]> {
+  // To fetch the metadata.json file dynamically by version do the following:
```
bd82, line 11: I don't understand "coding in comments"
If you want to support a dynamic flow, add such support as an optional flow directly why the complexity?

test-packages/test-utils/src/utils/download-ui5-resources.ts
```diff
@@ -0,0 +1,115 @@
+// This file can be used to download api.json files for UI5 version we want to add to the test directory.
+// It is run from the download-resources script in the package.json.
+// To get a different version change the version number in the last line of this file.
+import { TestModelVersion } from "../../api";
+import { reduce, forEach, keys } from "lodash";
+import { dirname, resolve } from "path";
+import { writeFileSync, mkdirsSync, readFileSync } from "fs-extra";
```
bd82, line 7: do we need **sync** variants here if all functions in this file are async?

bd82:
### Process
There are about 3 different topics in a single commit (2 commits, one of which is a merge commit) here. This can be confusing, particularly when one of the commits adds 700K+ LOC.

These are test utils so its not horrible... yet something to keep in mind.

### Contents
If we can now download relevant API.jsons on the fly, we can consider removing them from the source control and download those as needed.
 - OFC with caching to some local folder...

### Legal
We appear to have some metadata.json file in the source code ?
- Is this an openui5 metadata? is this file orignally under an open source license? if not it cannot be part of this repo.

packages/semantic-model/src/api-json.d.ts
```diff
@@ -66,12 +66,14 @@ export type Exceptions = {
  * Namespaces can have the common symbol properties and extend another symbol
  */
 export type NamespaceSymbol = SymbolBase & {
-  kind?: "namespace";
+  kind?: "namespace" | "member";
```
bd82, line 69: This is strange, what is this "member"? is this a mistake in the JSDocs?
Maybe a pre-processor is needed for the API.json files so the schema would not contain  invalid data?

bd82: This is particularly suspicious as we did not need it for version 1.74, so it seems like a bug that was fixed.

test-packages/test-utils/src/utils/download-ui5-resources.ts
tal-sapan: Mostly because I didn't want to change the `tsconfig.base.json` to support dom APIs for a test utility script, and I couldn't find a way to compile it when changing only the `tsconfig.json` of the test package. It's only required for the `jszip` usage but I couldn't find another library with nice API for reading zip files (I didn't spend a lot of time on it though).

test-packages/test-utils/src/utils/download-ui5-resources.ts
tal-sapan: Changed

packages/semantic-model/src/api-json.d.ts
tal-sapan: I agree that it's probably a bug in the jsdoc. The question is if it's necessary to pre-fix it or if it's better to handle it in the schema and conversion code like now. There are other likely bugs in the jsdoc that we probably aren't going to pre-fix (like extra fields on some types that I also added to the schema).

tal-sapan:
> ### Process
> There are about 3 different topics in a single commit (2 commits, one of which is a merge commit) here. This can be confusing, particularly when one of the commits adds 700K+ LOC.
> 
> These are test utils so its not horrible... yet something to keep in mind.

I don't consider the json files to be lines of code, so it's not really a lot of code. You aren't going to review the json files. The topic is adding support for version 1.71.14 (currently in tests only), and for this I had to add the test files and adapt the schema. The rest of it is support for adding ui5 versions to the tests: the validation errors output help find out schema problems in the downloaded libraries, and the script makes it easier to add new versions (and if we use the script for downloading test files on the fly instead of using it as a development time tool, it would be necessary for the tests themselves).

> ### Contents
> If we can now download relevant API.jsons on the fly, we can consider removing them from the source control and download those as needed.
> 
> * OFC with caching to some local folder...

Do you mean in the tests?

> ### Legal
> We appear to have some metadata.json file in the source code ?
> 
> * Is this an openui5 metadata? is this file orignally under an open source license? if not it cannot be part of this repo.

Actually now that the metadata.json is in npmjs.com we can fetch it directly without submitting the file to our repository (and also without unzipping it). If it's not in our repo, would it solve this issue? It does make me wonder about the api.json files though. Maybe they can't be submitted either (even as test code)?

test-packages/test-utils/src/utils/download-ui5-resources.ts
bd82: If we keep the api.json files as part of the source code, then perhaps this code
should not be part of the test-utilities/src and instead be located as part of test-utilities/scripts and then you can **write it in JavaScript** and avoid the TypeScript overhead.

Even if we do want to use it in automated test flows (download api.json on demand in tests) rather then manual execution. We can still put it in a **different** package under test-packages folder, and implement said packages purely in JavaScript.

btw we seem to be throwing everything into a single monolithic test-utilities package
which already created a cyclic package dependency... (unrelated to this PR)

packages/semantic-model/src/api-json.d.ts
bd82: What are the capabilities of schema merging or union types and schema --> d.ts generation?

I'd like to have some distinction between valid things and "ignored issues", e.g:

```typescript
interface namespace {
  kind?: "namespace" | InvalidNamespaceKind
}

export type InvalidNamespaceKind = "member"
```

Basically we need to distinguish between what is valid and what is not, otherwise our Schema will no longer be very useful in the long term.

bd82:
> Do you mean in the tests?

Yes.

> Actually now that the metadata.json is in npmjs.com we can fetch it directly without submitting the file to our repository (and also without unzipping it). If it's not in our repo, would it solve this issue? It does make me wonder about the api.json files though. Maybe they can't be submitted either (even as test code)?

API.json of openui5 can be submitted as they are derived from another OSS project.
But of the dist layer should not be submitted.
Which is another point in favor of downloading as needed and caching locally in a gitignored folder.

bd82:
> Process

O.k. so it is only 2 semi related topics and not 3 (version + download script)

packages/semantic-model/src/api-json.d.ts
tal-sapan: Everything under `definitions` gets a named type as far as I can tell. It might be possible to add an enum there with a single value and reference it from the namespace definition.
The question is what you define as "not valid" in the schema. Because while this looks like an unintended bug, it does actually exist in the api.json files (at least in some versions), and there is no "real" or "valid" version of the schema anywhere. The original schema had json syntax errors and even after fixing them I had to change it so it will capture what's really in the api.json files (not just the member thing, there were many other changes). Who's to say that not all of these fixes are errors in the jsdoc?

tal-sapan:
> API.json of openui5 can be submitted as they are derived from another OSS project.
> But of the dist layer should not be submitted.
> Which is another point in favor of downloading as needed and caching locally in a gitignored folder.

This is an interesting approach. I'll have to add code that fixes the issue I had with an invalid `implements` in one of the classes, but on the other hand we could delete the json files while not adding too much additional run time to the tests (though it will increase the run time by a few seconds for each tested version). Seeing as we don't have much choice with the dist layer libraries I will add this to this PR.

tal-sapan:
The new commit doesn't remove the json files, I will do that later (in this PR)

test-packages/test-utils/src/utils/download-ui5-resources.ts
tal-sapan: There are no comments now (also this is no longer a standalone script)

tal-sapan:
> ### Process
> There are about 3 different topics in a single commit (2 commits, one of which is a merge commit) here. This can be confusing, particularly when one of the commits adds 700K+ LOC.
 
There are still a few topics in this PR but instead of adding json files it now removes them, so I think it's worth it.

> ### Contents
> If we can now download relevant API.jsons on the fly, we can consider removing them from the source control and download those as needed.
> 
> * OFC with caching to some local folder...

Done

> ### Legal
> We appear to have some metadata.json file in the source code ?
> 
> * Is this an openui5 metadata? is this file orignally under an open source license? if not it cannot be part of this repo.

Removed (also the api.json files)

tal-sapan:
@bd82 I did the changes we talked about (download api.json files on the fly during the tests and fix them). All tests now run on the full sapui5 libraries so they had to be adapted.
Please note that `readTestLibraryFile` test utility function isn't used yet (I could remove it from this PR if you want, I don't want to add the test that uses it yet since this PR is big enough already).

.gitignore
```diff
@@ -78,4 +78,7 @@ lib
 # tsc incremental build artifacts
 tsconfig.tsbuildinfo
 
-*.vsix
\ No newline at end of file
+*.vsix
+
+# Cached test resources
+test-packages/test-utils/.model-cache
```
bd82, line 84: not really important or needs to be fixed...
I would consider a nested gitignore for such things in the future though.

packages/language-server/test/completion-items-spec.ts
```diff
@@ -9,13 +9,21 @@ import {
...
+  GEN_MODEL_TIMEOUT
+} from "@ui5-language-assistant/test-utils";
 
 import { getCompletionItems, computeLSPKind } from "../src/completion-items";
 
-const ui5SemanticModel: UI5SemanticModel = generateModel("1.74.0"); //TODO: use 1.71.x
-
 describe("the UI5 language assistant Code Completion Services", () => {
+  let ui5SemanticModel: UI5SemanticModel;
+  before(async function() {
```
bd82, line 21: Good application of best practice to avoid heavy logic outside test flow 👍

packages/language-server/test/completion-items-spec.ts
```diff
@@ -9,13 +9,21 @@ import {
...
+    this.timeout(GEN_MODEL_TIMEOUT);
+    //TODO: use 1.71.x
+    ui5SemanticModel = await generateModel("1.74.0");
```
bd82, line 24: switching to 1.71 now breaks tests?

packages/language-server/test/completion-items-spec.ts
```diff
@@ -172,7 +180,12 @@ describe("the UI5 language assistant Code Completion Services", () => {
       "upload",
       "util",
       "ux3",
-      "uxap"
+      "uxap",
+      "ubc",
+      "ui",
+      "ui", // This "ui" is in a different parent
```
bd82, line 186: is there more information on the suggestion to distinguish those? 
Basically make an assertion on more then just the label as you have implemented
in other tests?

packages/semantic-model/package.json
```diff
@@ -34,7 +34,7 @@
     "lint": "eslint . --ext .ts --max-warnings=0 --ignore-path=../../.gitignore",
     "test": "mocha",
     "coverage": "nyc mocha",
-    "generate-json-api": "json2ts -i ./resources/sap-ui-library-api.json -o ./src/api-json.d.ts --bannerComment \"/* eslint-disable @typescript-eslint/no-explicit-any */ /* This file is automatically generated. DO NOT MODIFY IT BY HAND. Instead, modify the source JSONSchema file (sap-ui-library-api.json) and run generate-json-api from the package.json to regenerate this file. */\""
+    "generate-json-api": "json2ts -i ./resources/sap-ui-library-api.json -o ./src/api-json.d.ts --bannerComment \"/* eslint-disable @typescript-eslint/no-explicit-any */ /* This file is automatically generated. DO NOT MODIFY IT BY HAND. Instead, modify the source JSONSchema file (sap-ui-library-api.json) and run generate-json-api from the package.json to regenerate this file. */\" && prettier --write ./src/api-json.d.ts"
```
bd82, line 37: At some point we will need to extract this command to a separate script file.

packages/semantic-model/src/api-json.d.ts
bd82: The problem is:
1. Overtime we will add more and more "exceptions" due to human errors and more UI5 versions supported.
2. We won't know what parts of the schema are errors and what are not.

Handling this issue can be deferred to another time, however we should not forget this technical debt. I think we may need to do a push to move this schema to the UI5 build
to fail fast on human errors...

@codeworrior FYI

bd82: I started an email discussion about this. lets defer it to the future.

packages/semantic-model/src/api.ts
```diff
@@ -16,15 +16,22 @@ export function generate({
...
 }: {
   version: string;
   libraries: Record<string, Json>;
   typeNameFix: TypeNameFix;
   strict?: boolean;
+  printValidationErrors?: boolean;
 }): UI5SemanticModel {
   const jsonSymbols = newMap<ConcreteSymbol>();
-  const model = convertToSemanticModel(libraries, jsonSymbols, strict);
+  const model = convertToSemanticModel(
```
bd82, line 29: if/when we need to add another param (in future) to `convertToSemanticModel` we should probably use "config object pattern"

packages/semantic-model/src/convert.ts
```diff
@@ -40,7 +41,9 @@ export function convertToSemanticModel(
   reduce(
     sortedLibs,
     (model, { libraryName, fileContent }) => {
-      if (isLibraryFile(libraryName, fileContent, strict)) {
+      if (
+        isLibraryFile(libraryName, fileContent, strict, printValidationErrors)
```
bd82, line 45: all these state parameters make me re-think if we may want to convert these to class **state** instead of a pure functional flow (in the future), particularly if we find ourselves adding more flags.

It may also be possible to pass the flags under a single "complex" object to reduce complexity (if we will need more flags...)

packages/semantic-model/test/api-spec.ts
```diff
@@ -233,41 +235,45 @@ context("The ui5-language-assistant semantic model package API", () => {
...
+      });
+
+      it(`is created successfully in strict mode`, async () => {
+        const model = await generateModel(version, false);
         expect(model).to.exist;
       });
-      it(`has correct parent on root symbols`, () => {
-        const model = generateModel(version);
+      it(`has correct parent on root symbols`, async () => {
+        const model = await generateModel(version, false);
```
bd82, line 248: should the model generation be extracted to `before` clause?

packages/vscode-ui5-language-assistant/test/suite/extension-spec.ts
```diff
@@ -74,7 +74,17 @@ describe("the Language Server Client Integration Tests", () => {
...
                           xmlns:u⇶`;
-    const completionsList = ["unified", "upload", "util", "ux3"];
+    const completionsList = [
+      "unified",
+      "upload",
+      "util",
+      "ux3",
+      "ubc",
+      "ui",
+      "ui", // This "ui" is in a different parent
```
bd82, line 84: as mentioned earlier, do we have enough info to make a better assertion that distinguish between `ui` and `ui`?

test-packages/test-utils/api.d.ts
```diff
@@ -68,19 +69,23 @@ export function buildUI5Model(
...
+ * If downloadLibraries is true (default), increase the timeout of the test/hook to GEN_MODEL_TIMEOUT.
+ *
  * @param version
+ * @param downloadLibraries By default, download the library files before creating the model.
+ *        If you call downloadLibraries explicitly before this function, send false (it will improve the performance).
  * @param disableCache By default caching is used. This improves performance and is safe (since the model is immutable).
  */
 export function generateModel(
   version: TestModelVersion,
+  downloadLibraries?: boolean,
```
bd82, line 86: **Important:**

Multiple optional flags (of same type) in public (even if test-packages) `API --> Config Object Pattern`

test-packages/test-utils/api.d.ts
```diff
@@ -119,3 +124,17 @@ export function expectSuggestions(
...
+): Promise<FetchResponse>;
+
+export function downloadLibraries(version: TestModelVersion): Promise<void>;
+
+/**
+ * generateModel can take a long time in some cases now that it might download the libraries.
+ * Use this constant to set the timeout on the test/hook that calls it.
+ * Note: this.timeout(...) cannot be called from within an arrow function.
+ */
+export const GEN_MODEL_TIMEOUT = 3000;
```
bd82, line 140: we may need even more to avoid random failures, e.g on slow containers in circle-ci
or slow developer machines.

test-packages/test-utils/api.d.ts
```diff
@@ -119,3 +124,17 @@ export function expectSuggestions(
   suggestions: UI5XMLViewCompletion[],
   expected: string[]
 ): void;
+
+export function readTestLibraryFile(
```
bd82, line 128: we should consider (in future) having multiple `test-packages` this one is becoming a bit of a monolith

test-packages/test-utils/src/utils/download-ui5-resources.ts
```diff
@@ -0,0 +1,78 @@
+import { TestModelVersion } from "../../api";
+import { reduce, keys, map } from "lodash";
+import { resolve } from "path";
+import { writeFile, mkdirs, pathExists } from "fs-extra";
+import fetch from "node-fetch";
+
+async function getLibs(version: TestModelVersion): Promise<string[]> {
+  // The metadata.json seems to have been added only very recently :(
+  let versionInURL: string = version;
```
bd82, line 9: The name `versionInURL` could be more clear if it indicated it is used in **metadata** url.
As I am sure more URLs are built somewhere along this flow.

Also we should explicitly document the assumption that the list of libraries has not changed, (maybe even check with UI5 team?)

test-packages/test-utils/src/utils/download-ui5-resources.ts
```diff
@@ -0,0 +1,78 @@
...
+  const librariesMetadata = JSON.parse(fileContent);
+  return keys(librariesMetadata.libraries);
+}
+
+export async function addUi5Resources(
+  version: TestModelVersion,
+  folder: string
+): Promise<void> {
+  // CDN libraries (example URL):
+  // https://sapui5-sapui5.dispatcher.us1.hana.ondemand.com/test-resources/sap/m/designtime/api.json
```
bd82, line 29: is this comment still relevant? (the `us1` based url)

test-packages/test-utils/src/utils/download-ui5-resources.ts
```diff
@@ -0,0 +1,78 @@
...
+  // Older versions:
+  // https://sapui5.hana.ondemand.com/1.71.14/test-resources/sap/m/designtime/api.json
+  const baseUrl = `https://sapui5.hana.ondemand.com/${version}/test-resources/`;
+  const libs = await getLibs(version);
+  const nameToFile = reduce(
```
bd82, line 34: This is **subjective**:

Because the size of the input and output is the same
this **reduce** is not much more than a glorified forEach.

What do you think of combining Map and ZipObject
- https://lodash.com/docs/4.17.15#zipObject

To obtain he same result? 
would it be clearer?

test-packages/test-utils/src/utils/semantic-model-provider.ts
```diff
@@ -99,3 +213,71 @@ export function generateModel(
   }
   return model;
 }
+
+// Fix functions for model issues unfixable by schema changes
+type LibraryFix = (content: Json) => void;
```
bd82, line 218: how do we handle this in none test flows?

test-packages/test-utils/src/utils/semantic-model-provider.ts
```diff
@@ -80,15 +188,21 @@ export function loadLibraries(version: TestModelVersion): Record<string, Json> {
...
+  downloadLibs = true,
   disableCache = false
-): UI5SemanticModel {
+): Promise<UI5SemanticModel> {
   if (has(MODEL_CACHE, version) && !disableCache) {
     return MODEL_CACHE[version];
   }
 
+  if (downloadLibs) {
+    await downloadLibraries(version);
```
bd82, line 201: Are we assuming here that the test infrastructure never executes test N+1 before test N has finished ?

Could subsequent calls to `generateModel` be executed before the first one finished?
and the libraries be downloaded multiple times?

I am not sure this can happen in Mocha, but perhaps we should take care of this regardless? (e.g return a **singleton** promise to the result (per ui5 version) instead (?) of a **new** promise each time this function is called

.gitignore
tal-sapan: The problem is that eslint and prettier both use the top level gitignore file and don't support gitignore hierarchy.

packages/language-server/test/completion-items-spec.ts
tal-sapan: I'm not sure but I didn't want to add it to this PR (the comment was already there from before, and I think if we change the version here we should do it in all such tests)

packages/language-server/test/completion-items-spec.ts
tal-sapan: There is, we can add the `detail`. Didn't want to change the tests right now.

packages/semantic-model/src/convert.ts
tal-sapan: I think an object for the options makes sense

packages/semantic-model/test/api-spec.ts
tal-sapan: The first test tests it... I could add another describe for the other 3 tests

packages/language-server/test/completion-items-spec.ts
bd82: o.k, I wonder if we should change test version or not, perhaps it does not matter because we want  to support multiple versions in the future.

test-packages/test-utils/api.d.ts
tal-sapan: Actually I think I'll remove the third flag since it's not in use (I thought I'll use it... but then I didn't so it's unnecessary)

test-packages/test-utils/api.d.ts
tal-sapan: I could increase it to 5000

test-packages/test-utils/api.d.ts
tal-sapan: We should probably at least separate the model build utilities and the model generation utilities

test-packages/test-utils/src/utils/download-ui5-resources.ts
tal-sapan: I'm pretty sure it has changed since some of the URLs return an error for older versions. Hopefully they didn't remove any libraries but only added, but yes, we should check with them.

test-packages/test-utils/src/utils/download-ui5-resources.ts
tal-sapan: I prefer to leave it here until we know for sure what to do in case there is no defined library version (right now we use a default but when we take it from the project, I think that if it's not defined, it will use the latest which is the CDN url in runtime... but I'm not sure).

test-packages/test-utils/src/utils/download-ui5-resources.ts
tal-sapan: Yes, I'll change it

test-packages/test-utils/src/utils/semantic-model-provider.ts
tal-sapan: We don't, but in productive code `strict` is false so all these issues are ignored.
I'll add tests for this.

tal-sapan: Do you think we should run all the tests in non-strict mode and without these fixes?

test-packages/test-utils/src/utils/semantic-model-provider.ts
tal-sapan: Changed in `downloadLibraries`

test-packages/test-utils/api.d.ts
tal-sapan: I'll change it to config object

packages/semantic-model/test/api-spec.ts
bd82: your choice

test-packages/test-utils/api.d.ts
bd82: It is just a test-package, but config-objects are a best practice for public APIs 😄

test-packages/test-utils/src/utils/download-ui5-resources.ts
bd82: I meant why does the CDN link has a specific **region** (US) included?

test-packages/test-utils/src/utils/semantic-model-provider.ts
bd82: Hmm, running tests in similar way to productive flows is generally a good things

test-packages/test-utils/src/utils/download-ui5-resources.ts
tal-sapan: I think it has to have some region... if you know a URL without a region we can change the comment (but then I assume it will be the EU region or something)

codeworrior: @petermuessig shouldn't this be the region-free entry URL so that the automatic server selection is active?

tal-sapan:
@bd82 If there are no more comments could you please approve the PR?

PR #39, opened 2020-04-02
feat: support SAP UI5 Distribution libraries version 1.71.14

tal-sapan:
Set the version on the model directly since the versions in the distribution libraries aren't the same across all libraries.

### DIFF
--- a/packages/language-server/src/server.ts
+++ b/packages/language-server/src/server.ts
@@ -14,8 +14,7 @@ import { getCompletionItems } from "./completion-items";
 
 const connection = createConnection(ProposedFeatures.all);
 const documents = new TextDocuments(TextDocument);
-let model: UI5SemanticModel;
-let isSemanticModelCreated = false;
+let getSemanticModelPromise: Promise<UI5SemanticModel> | undefined = undefined;
 
 connection.onInitialize(() => {
   return {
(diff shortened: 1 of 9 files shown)


packages/language-server/src/server.ts
```diff
@@ -32,17 +31,15 @@ connection.onInitialize(() => {
...
-  }
+  getSemanticModelPromise = getSemanticModel();
 });
 
 connection.onCompletion(
   async (
     textDocumentPosition: TextDocumentPositionParams
   ): Promise<CompletionItem[]> => {
-    if (isSemanticModelCreated) {
+    if (getSemanticModelPromise !== undefined) {
```
bd82, line 41: 👍

packages/language-server/src/ui5-model.ts
```diff
@@ -7,28 +7,44 @@ import {
   Json,
   TypeNameFix
 } from "@ui5-language-assistant/semantic-model";
+import { Fetcher } from "api";
```
bd82, line 10: This import is strange, could we import with a **relative** path instead?

packages/language-server/src/ui5-model.ts
```diff
@@ -7,28 +7,44 @@ import {
...
 
-// TODO: use 1.71.x
-const DEFAULT_UI5_VERSION = "1.74.0";
+const DEFAULT_UI5_VERSION = "1.71.14";
+
+let fetcher: Fetcher = fetch;
+
+// These functions are for testing purposes
+/* istanbul ignore next */
+export function getFetch(): Fetcher {
```
bd82, line 18: - What is the advantage of file(module) level state instead of optional parameter on  `getSemanticModel` with default value?

- In what scenario would you like to use these custom fetchers in the tests?

tal-sapan: 1. I don't want it to be used outside the test for this file so I'd rather not add a parameter to a function used somewhere else. I could however add a function that receives this parameter and will not be used outside this file and its test.
1. I intended to use it in the test to fetch the files from the local tests directory instead of off the internet. The solution I wanted to do doesn't exactly work with the changes I'll do in the PR that adds tests for 1.71.14 but it could still be useful - I'll write the files to the local ignored directory so it will only happen once for each version for all tests. It will be part of the other PR anyway.

bd82: o.k I'll look into the dev flows and downloading of resources again after its all done.

BTw: you may want to pre-download before the tetsts start running to avoid timeouts.

PR #44, opened 2020-04-05
feat(xml-views-completion): do not suggest abstract classes in XML tags

bd82:


packages/semantic-model/src/convert.ts
```diff
@@ -162,9 +162,10 @@ function convertClass(
   symbol: apiJson.ClassSymbol
 ): model.UI5Class {
   const base = convertSymbol(libName, symbol);
-  const classs: model.UI5Class = {
+  const classes: model.UI5Class = {
```
tal-sapan, line 165: It's one class... I couldn't call it `class` because it's a reserved word so I added an extra `s`

bd82: ah I thought its a typo and created a new typo, I will rename to `clazz`

bd82:
> The rest looks ok

I fixed the typo (renamed to `clazz`) and will merge.
Let me know if you prefer another name...

PR #49, opened 2020-04-06
fix(vscode-ui5-language-assistant): full schema v1.19

sapirpol:


tal-sapan:
Could you move the files to an inner folder inside "resources" like "manifest-schema"? We might want to use the resources for more things in the future

PR #50, opened 2020-04-06
perf(language-server): cache downloaded resources

tal-sapan:
Download and cache UI5 resources in the plugin's globalStoragePath instead of downloading it again on every plugin activation.

### DIFF
--- a/packages/language-server/api.d.ts
+++ b/packages/language-server/api.d.ts
@@ -4,6 +4,10 @@
  */
 export declare const SERVER_PATH: string;
 
+export type ServerInitializationOptions = {
+  modelCachePath: string;
+};
+
 export type FetchResponse = {
   ok: boolean;
   json: () => Promise<unknown>;
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -34,7 +34,11 @@
   "devDependencies": {
     "@types/node-fetch": "2.5.5",
     "@ui5-language-assistant/semantic-model-types": "^1.0.0",
-    "vscode-languageserver-types": "3.15.1"
+    "vscode-languageserver-types": "3.15.1",
+    "tmp-promise": "2.0.2",
+    "@types/tmp": "0.1.0",
+    "rimraf": "3.0.2",
+    "@types/rimraf": "3.0.0"
   },
   "scripts": {
     "ci": "npm-run-all clean compile lint coverage",
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -36,9 +36,7 @@
     "@ui5-language-assistant/semantic-model-types": "^1.0.0",
     "vscode-languageserver-types": "3.15.1",
     "tmp-promise": "2.0.2",
-    "@types/tmp": "0.1.0",
-    "rimraf": "3.0.2",
-    "@types/rimraf": "3.0.0"
+    "@types/tmp": "0.1.0"
   },
   "scripts": {
     "ci": "npm-run-all clean compile lint coverage",
--- a/packages/language-server/src/ui5-model.ts
+++ b/packages/language-server/src/ui5-model.ts
@@ -36,6 +36,7 @@ export async function getSemanticModelWithFetcher(
   // a warning to the user
   if (modelCachePath !== undefined) {
     cacheFolder = getCacheFolder(modelCachePath, version);
+    console.log(`${cacheFolder} will be used to cache UI5 resources`);
     try {
       await mkdirs(cacheFolder);
     } catch (err) {
@@ -49,39 +50,15 @@ export async function getSemanticModelWithFetcher(
 
   await Promise.all(
     map(libs, async libName => {
-      let filePath;
-      let apiJson;
-      if (cacheFolder !== undefined) {
-        filePath = getCacheFilePath(cacheFolder, libName);
-        try {
-          if (
-            (await pathExists(filePath)) &&
-            (await lstat(filePath)).isFile()
-          ) {
-            apiJson = await readJson(filePath);
-          }
-        } catch (err) {
-          console.warn(
-            `Could not read file ${filePath} from UI5 resources cache`,
-            err
-          );
-        }
-      }
+      const cacheFilePath = getCacheFilePath(cacheFolder, libName);
+      let apiJson = await readFromCache(cacheFilePath);
+      // If the file doesn't exist in the cache (or we couldn't read it), fetch it from the network
       if (apiJson === undefined) {
         const url = baseUrl + libName.replace(/\./g, "/") + suffix;
         const response = await fetcher(url);
         if (response.ok) {
           apiJson = await response.json();
-          if (filePath !== undefined) {
-            try {
-              await writeJson(filePath, apiJson);
-            } catch (err) {
-              console.warn(
-                `Could not cache UI5 resources to file ${filePath}`,
-                err
-              );
-            }
-          }
+          await writeToCache(cacheFilePath, apiJson);
         } else {
           console.error(`Could not read UI5 resources from ${url}`);
         }
@@ -100,6 +77,35 @@ export async function getSemanticModelWithFetcher(
   });
 }
 
+async function readFromCache(filePath: string | undefined): Promise<unknown> {
+  if (filePath !== undefined) {
+    try {
+      if ((await pathExists(filePath)) && (await lstat(filePath)).isFile()) {
+        return await readJson(filePath);
+      }
+    } catch (err) {
+      console.warn(
+        `Could not read file ${filePath} from UI5 resources cache`,
+        err
+      );
+    }
+  }
+  return undefined;
+}
+
+async function writeToCache(
+  filePath: string | undefined,
+  apiJson: unknown
+): Promise<void> {
+  if (filePath !== undefined) {
+    try {
+      await writeJson(filePath, apiJson);
+    } catch (err) {
+      console.warn(`Could not cache UI5 resources to file ${filePath}`, err);
+    }
+  }
+}
+
 // Exported for test purposes
 export function getCacheFolder(
   modelCachePath: string,
@@ -108,7 +114,13 @@ export function getCacheFolder(
   return resolve(modelCachePath, "ui5-resources-cache", version);
 }
 // Exported for test purposes
-export function getCacheFilePath(cacheFolder: string, libName: string): string {
+export function getCacheFilePath(
+  cacheFolder: string | undefined,
+  libName: string
+): string | undefined {
+  if (cacheFolder === undefined) {
+    return undefined;
+  }
   return resolve(cacheFolder, libName + ".json");
 }
 
--- a/packages/language-server/src/ui5-model.ts
+++ b/packages/language-server/src/ui5-model.ts
@@ -101,7 +101,10 @@ export async function getSemanticModelWithFetcher(
 }
 
 // Exported for test purposes
-export function getCacheFolder(modelCachePath: string, version: string) {
+export function getCacheFolder(
+  modelCachePath: string,
+  version: string
+): string {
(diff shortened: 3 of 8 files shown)


packages/language-server/api.d.ts
```diff
@@ -4,6 +4,10 @@
  */
 export declare const SERVER_PATH: string;
 
+export type ServerInitializationOptions = {
```
bd82, line 7: Could you add a comment that these params are passed via the connection?

packages/language-server/package.json
```diff
@@ -34,7 +34,11 @@
   "devDependencies": {
     "@types/node-fetch": "2.5.5",
     "@ui5-language-assistant/semantic-model-types": "^1.0.0",
-    "vscode-languageserver-types": "3.15.1"
+    "vscode-languageserver-types": "3.15.1",
+    "tmp-promise": "2.0.2",
+    "@types/tmp": "0.1.0",
+    "rimraf": "3.0.2",
```
bd82, line 40: - rimraf is already a common dev dep at the root of the repo.
- `@types/rimraf` could also be added to the root package.

packages/language-server/src/ui5-model.ts
```diff
@@ -11,29 +13,81 @@ import { Fetcher } from "../api";
...
   const version = DEFAULT_UI5_VERSION;
   const jsonMap: Record<string, Json> = {};
   const baseUrl = `https://sapui5.hana.ondemand.com/${version}/test-resources/`;
   const suffix = "/designtime/api.json";
   const libs = getLibs();
+  let cacheFolder: string | undefined;
+
+  // Note: all cache handling (reading, writing etc) is optional from the user perspective but
+  // impacts performance, therefore if any errors occur when handling the cache we ignore them but output
+  // a warning to the user
```
bd82, line 36: - Where do these console.logs go? (The extension host terminal?)
- Could we get these logs to the output channel?
- I think I saw some APIs to log via the connection somewhere...

packages/language-server/src/ui5-model.ts
```diff
@@ -11,29 +13,81 @@ import { Fetcher } from "../api";
...
+        } catch (err) {
+          console.warn(
+            `Could not read file ${filePath} from UI5 resources cache`,
+            err
+          );
+        }
+      }
+      if (apiJson === undefined) {
+        const url = baseUrl + libName.replace(/\./g, "/") + suffix;
+        const response = await fetcher(url);
```
bd82, line 72: we may want to log successful flows as well as caching can add complexity and strange bugs.

packages/language-server/test/ui5-model-spec.ts
```diff
@@ -29,5 +41,110 @@ describe("the UI5 language assistant ui5 model", () => {
...
-  }).timeout(5000); // The default timeout is 2000ms and this test takes ~3000ms
+  }
+
+  it("will get UI5 semantic model", async () => {
+    const ui5Model = await getSemanticModel(undefined);
+    assertSemanticModel(ui5Model);
+  }).timeout(GET_MODEL_TIMEOUT);
+
+  it("doesn't fail if a file cannot be fetched", async () => {
+    const ui5Model = await getSemanticModelWithFetcher(async (url: string) => {
```
bd82, line 52: why won't this fail? does it not fail because the previous test populated a cache?

packages/language-server/test/ui5-model-spec.ts
```diff
@@ -29,5 +41,110 @@ describe("the UI5 language assistant ui5 model", () => {
...
+          throw new Error(`Cannot read from ${url}`);
+        }
+      };
+    }, undefined);
+    expect(ui5Model).to.exist;
+  });
+
+  describe("with cache", () => {
+    describe("cache in temp dir", () => {
+      let cachePath: string;
```
bd82, line 65: const?

packages/language-server/test/ui5-model-spec.ts
```diff
@@ -29,5 +41,110 @@ describe("the UI5 language assistant ui5 model", () => {
...
+      beforeEach(async () => {
+        ({ path: cachePath } = await tempDir());
```
bd82, line 67: very useful tmp library

packages/language-server/test/ui5-model-spec.ts
```diff
@@ -29,5 +41,110 @@ describe("the UI5 language assistant ui5 model", () => {
...
+        rimrafSync(cachePath);
+      });
+
+      it("caches the model the first time getSemanticModel is called", async () => {
+        const ui5Model = await getSemanticModel(cachePath);
+        assertSemanticModel(ui5Model);
+
+        // Check the files were created in the folder
+        const files = await readdir(cachePath);
+        expect(files.length).to.be.greaterThan(0);
```
bd82, line 82: `to.not.be.empty` ?
- https://www.chaijs.com/api/bdd/#method_empty

packages/language-server/test/ui5-model-spec.ts
```diff
@@ -29,5 +41,110 @@ describe("the UI5 language assistant ui5 model", () => {
...
+            throw new Error(
+              `The files should be taken from the cache, got call for ${url}`
+            );
+          },
+          cachePath
+        );
+        expect(fetcherCalled).to.be.false;
+        // Make sure it's not the model itself that is cached
+        expect(ui5ModelFromCache).to.not.equal(ui5Model);
+        assertSemanticModel(ui5ModelFromCache);
```
bd82, line 98: maybe a small assertion that the cached model is identical (deep values) to the none cached one?
I know we cannot do full deep equals here due to the size of the model but perhaps we can check the number of classes/enums/namespaces/....

packages/language-server/test/ui5-model-spec.ts
```diff
@@ -29,5 +41,110 @@ describe("the UI5 language assistant ui5 model", () => {
...
+      }).timeout(GET_MODEL_TIMEOUT);
+
+      it("doesn't fail when file cannot be written to the cache", async () => {
+        // Create a folder with the file name so the file will not be written
+        await mkdirs(
+          getCacheFilePath(getCacheFolder(cachePath, VERSION), "sap.m")
+        );
+        const ui5Model = await getSemanticModel(cachePath);
+        expect(ui5Model).to.exist;
```
bd82, line 107: assert the model contains a class from sap.m ?

packages/language-server/test/ui5-model-spec.ts
```diff
@@ -29,5 +41,110 @@ describe("the UI5 language assistant ui5 model", () => {
...
+      }).timeout(GET_MODEL_TIMEOUT);
+
+      it("doesn't fail when file cannot be read from the cache", async () => {
+        // Create a file with non-json content so the file will not be deserialized
+        const cacheFolder = getCacheFolder(cachePath, VERSION);
+        await mkdirs(cacheFolder);
+        await writeFile(getCacheFilePath(cacheFolder, "sap.m"), "not json");
+        const ui5Model = await getSemanticModel(cachePath);
+        expect(ui5Model).to.exist;
```
bd82, line 116: again assert something from sap.m is actually contained in the model

packages/language-server/test/ui5-model-spec.ts
```diff
@@ -29,5 +41,110 @@ describe("the UI5 language assistant ui5 model", () => {
...
+      }).timeout(GET_MODEL_TIMEOUT);
+    });
+
+    describe("cache path is a file", async () => {
```
bd82, line 120: **when** cache path is a file ?

packages/language-server/src/ui5-model.ts
```diff
@@ -11,29 +13,81 @@ import { Fetcher } from "../api";
...
   await Promise.all(
     map(libs, async libName => {
-      const url = baseUrl + libName + suffix;
-      const response = await fetcher(url);
-      if (response.ok) {
-        const apiJson = await response.json();
+      let filePath;
+      let apiJson;
+      if (cacheFolder !== undefined) {
+        filePath = getCacheFilePath(cacheFolder, libName);
```
bd82, line 55: could this be flow be extracted to two sub functions?
- Cache One
- Network one.

packages/language-server/api.d.ts
tal-sapan: I'll add it where it's read

packages/language-server/package.json
tal-sapan: I'll move it

packages/language-server/src/ui5-model.ts
tal-sapan: They go to the output channel in vscode with the defined extension name (`UI5 Language Assitant`). Same as `connection.console`.

packages/language-server/src/ui5-model.ts
tal-sapan: I didn't want to write too much to the output since I don't think the user can turn off info messages. Maybe we should add messages when reading from the URL instead? Since it will happen less times (hopefully). And when it doesn't output anything we can assume it's taken from the cache.

packages/language-server/test/ui5-model-spec.ts
tal-sapan: The tests don't rely on the order - the previous test doesn't populate the cache because we don't send a cache folder. It doesn't fail because the code is fault-tolerant. It will write the error to the output and do its best with what it could get instead of not working at all.

tal-sapan: I'll make it more explicit that we don't populate the cache in these 2 tests

packages/language-server/test/ui5-model-spec.ts
tal-sapan: Will change

packages/language-server/test/ui5-model-spec.ts
tal-sapan: I'll add a check that `sap.m` namespace exists and taken from the correct library

packages/language-server/test/ui5-model-spec.ts
tal-sapan: Done

packages/language-server/test/ui5-model-spec.ts
tal-sapan: Does `when` exist in mocha? I couldn't find it in the documentation or types definition

packages/language-server/src/ui5-model.ts
tal-sapan: Not really because if we couldn't read from the cache we'll do the network call anyway, but I'll extract the cache handling to 2 functions so the flow is clearer

tal-sapan:
> * We will need a way to force clean the cache (e.g VSCode command).

Probably. For now I'll add a console log that outputs the folder so if there are issues we can tell the user to delete it.

> * Does the extension resources path API works in Theia too?

I'll check, it's a good question.

tal-sapan:
> > * Does the extension resources path API works in Theia too?
> 
> I'll check, it's a good question.

It seems to work and hopefully it's also persisted (it's in `/home/user/.theia/globalStorage`).

packages/language-server/test/ui5-model-spec.ts
tal-sapan: Done

packages/language-server/src/ui5-model.ts
bd82: In the future we will need configurable log levels controlled via the user settings in VSCode.

packages/language-server/test/ui5-model-spec.ts
bd82: I suggested adding "when" prefix to the description but maybe the full suggestion is. `content("when cache path is a file")`, but this is **completely subjective**

tal-sapan:
Seems to be persisted (I stopped and started the workspace and it read the files from the cache).

PR #52, opened 2020-04-07
feat(xml-views-completion): suggest namespaces on non-class elements

tal-sapan:


tal-sapan:
After discussion with Frank and Peter it seems that `xmlns` attributes should work everywhere in the xml file and not only on class elements.

PR #55, opened 2020-04-19
feat(semantic-model): support special attributes - xml completion

sapirpol:


packages/semantic-model/src/convert.ts
```diff
@@ -193,7 +193,15 @@ function convertClass(
       symbol["ui5-metadata"].properties,
       partial(convertProperty, libName, clazz)
     );
+    // Add special settings (for example: "id" from ManagedObject)
+    clazz.properties = clazz.properties.concat(
```
bd82, line 197: Should we add these properties on **every** class? or should we instead add these only on the top level class[es]?

@tal-sapan WDYT?

packages/xml-views-completion/test/providers/attributeName/prop-event-assoc-spec.ts
```diff
@@ -37,12 +37,23 @@ const radioButtonGroupProperties = [
...
+  "Type"
+];
+
 const radioButtonGroupEvents = ["select"];
 const uiCoreControlAssociations = ["ariaDescribedBy", "ariaLabelledBy"];
 const allPropsEventsAssociations = uiCoreControlProperties
   .concat(uiCoreControlEvents)
   .concat(uiCoreControlAssociations)
   .concat(radioButtonGroupProperties)
+  .concat(radioButtonGroupSpecialProperties)
```
bd82, line 56: it would be nice to have a separated test that only checks these specials properties.
That will give useful indication if this feature every breaks.

packages/semantic-model/src/convert.ts
tal-sapan: They only exist on a handful of classes (like `ManagedObject`, which `Element` inherits from so all controls and elements inherit from transitively). So there's no need to check where it's coming from. They're like regular properties.

bd82: But won't this mean we are breaking the model because, it may include classes that do not inherit from ManagedObject and yet we add these properties to **all** the classes...

Remember that we may want this model for other scenarios as well...

tal-sapan: We're not adding them on classes that don't have them... they will only be added for classes that have `specialSettings` under `ui5-metadata`

bd82: Ah I see now.

PR #56, opened 2020-04-19
feat(semantic-model): add experimental information

sapirpol:


packages/language-server/src/completion-items.ts
```diff
@@ -54,6 +54,9 @@ function transformToLspSuggestions(
   const lspSuggestions = map(suggestions, suggestion => {
     const lspKind = computeLSPKind(suggestion);
     let detailText = getNodeDetail(suggestion.ui5Node);
+    if (suggestion.ui5Node.experimentalInfo?.isExperimental) {
+      detailText = `(experimental) ${detailText}`;
```
tal-sapan, line 58: Did you consult with Yuval on whether we want this? I'm not sure we do...
Also, if we do, the "(experimental)" should be after the "(deprecated)" (or maybe not be there at all for deprecated APIs - ask Yuval this, too, please).

packages/language-server/src/documentation.ts
```diff
@@ -39,8 +39,23 @@ export function getNodeDocumentation(
       : EMPTY_STRING;
     contents += NL;
   }
-  // TODO add experimental when it's added to the model in this format: Experimental API [since version <since>]
 
+  if (node.experimentalInfo?.isExperimental) {
+    // Since and Experimental are both present
```
tal-sapan, line 44: I think you can remove this comment

packages/language-server/src/completion-items.ts
sapirpol: It was verified with Yuval, "(experimental)" should be after the "(deprecated)"

packages/language-server/test/completion-items-spec.ts
```diff
@@ -244,6 +244,24 @@ describe("the UI5 language assistant Code Completion Services", () => {
     ]);
   });
 
+  it("will get completion values for UI5 class", () => {
```
tal-sapan, line 247: Please mention that it's experimental in the test description

packages/language-server/src/completion-items.ts
bd82: should not we use Tags for deprecated instead of modifying the description?
- 
image

packages/language-server/src/documentation.ts
```diff
@@ -39,8 +39,24 @@ export function getNodeDocumentation(
       : EMPTY_STRING;
     contents += NL;
   }
-  // TODO add experimental when it's added to the model in this format: Experimental API [since version <since>]
 
+  if (node.experimentalInfo?.isExperimental) {
```
bd82, line 43: Far too many istanbul ignore statements here,,,

packages/language-server/src/documentation.ts
```diff
@@ -39,8 +39,24 @@ export function getNodeDocumentation(
...
+    /* istanbul ignore else */
+    if (contents !== "") {
+      contents += NL;
+    }
+    contents += "Experimental";
+    /* istanbul ignore next */
+    contents += node.experimentalInfo.since
```
bd82, line 50: Use Nullish coalescing instead of the ternary operator.

packages/language-server/test/completion-items-spec.ts
```diff
@@ -244,6 +244,24 @@ describe("the UI5 language assistant Code Completion Services", () => {
     ]);
   });
 
+  it("will get completion values for UI5 experimental class", () => {
+    const xmlSnippet = `<mvc:View 
+                          xmlns:mvc="sap.ui.core.mvc" 
+                          xmlns="sap.m">
+                          <content> <ContentS⇶`;
```
bd82, line 251: why is this snippet request completion on the closing tag?

packages/language-server/test/completion-items-spec.ts
```diff
@@ -244,6 +244,24 @@ describe("the UI5 language assistant Code Completion Services", () => {
...
+    const suggestions = getSuggestions(xmlSnippet);
+    const suggestionNames = map(suggestions, suggestion => ({
+      label: suggestion.label,
+      text: suggestion.insertText
+    }));
+    expect(suggestionNames).to.deep.equalInAnyOrder([
```
bd82, line 257: should we assert the existence of the `experimental` info in the description?

packages/language-server/src/completion-items.ts
bd82: https://microsoft.github.io/language-server-protocol/specifications/specification-current/

tal-sapan: We would but it causes crashes in Theia (it's not only unsupported but actually causes the code assist to not work). There's a comment about it where we create the `CompletionItem` instance.
This is the comparison between VSCode and Theia features:
https://che-incubator.github.io/vscode-theia-comparator/status.html

packages/language-server/src/documentation.ts
tal-sapan: Well, the lines are actually covered, the ignore is for the branches (since there's no `istanbul ignore if` for ternary expressions). Kind of hard to find the case where there is no since/text for deprecated or experimental.

packages/language-server/src/documentation.ts
tal-sapan: It should go to the `EMPTY_STRING` case also when `since` is an empty string

packages/language-server/test/completion-items-spec.ts
tal-sapan: It's not on the closing tag?

sapirpol: It's not for closing tag, it's for "ContentSwitcher"

packages/language-server/src/documentation.ts
bd82: - The result would be the same.
- The verbosity would be reduced.
- The semantics of the branch flow can be defined in multiple ways the specific branch
  chosen when X is an empty string is less relevant as long as the result is correct.

```javascript
const EMPTY_STRING = "";
const X = "";
// Print X unless X is undefined / Null
// otherwise Print EMPTY_STRING
console.log(X ?? EMPTY_STRING) // ""

// Print X unless X is undefined / Null / Empty / Empty Array / 0 / and the PRINT
// otherwise Print EMPTY_STRING
console.log(X ? X : EMPTY_STRING) // ""
```

packages/language-server/test/completion-items-spec.ts
bd82: ah the indentation confused me

packages/language-server/src/documentation.ts
bd82: At a minimum this should be documented as it seems like liberally using `istanbul ignore` to avoid writing tests.

But it seems better that this string builder would be tested with a **unit-test** particularly as this is a string that is visible to the end customer in the UI.

tal-sapan: I agree that we should add tests for this. It's a technical debt.

packages/language-server/src/documentation.ts
tal-sapan: Right, but we're not concatenating X here but a string that contains X

tal-sapan: If we don't check the case where `since` is an empty string we'll concatenate "` since version `" (without a version number) to the content.

bd82: ah I see, so lets move this comment to the concatnating of the "text" field below :)

packages/language-server/src/documentation.ts
bd82: @sapirpol could you please add **unit** tests for this string builder so we won't drag this debt into the future?

packages/language-server/src/documentation.ts
tal-sapan: It's less visible but there's an extra space there that will be added in that case

bd82: o.k. I see the `??` is not suitable here 😄

PR #59, opened 2020-04-22
feat(language-server): allow to replace FQN on code assist

tal-sapan:
Use `textEdit` instead of `insertText` to specify the entire range of characters that should be replaced when accepting the suggestion. By default only characters after `.` and `:` are replaced if `.` or `:` exist in the prefix.
This allows to replace fully qualified names like "sap.m.Button" with the correct tag name, to make for an easier search.

This also makes sure we replace the whole name even if the user requested the code assist in the middle of it (instead of inserting it and keeping the rest of the string): for example, selecting "MyClass2" when the cursor is at "MyClass|1" will result in "MyClass2" instead of "MyClass21"
like before.

Bug fix: use prefix instead of the full attribute value in namespace completion inside attribute value.

### DIFF
--- a/packages/language-server/src/completion-items.ts
+++ b/packages/language-server/src/completion-items.ts
@@ -3,11 +3,19 @@ import {
   CompletionItem,
   CompletionItemKind,
   TextDocumentPositionParams,
-  InsertTextFormat
+  InsertTextFormat,
+  TextEdit,
+  Range
 } from "vscode-languageserver";
 import { TextDocument } from "vscode-languageserver-textdocument";
 import { parse, DocumentCstNode } from "@xml-tools/parser";
-import { buildAst, DEFAULT_NS } from "@xml-tools/ast";
+import {
+  buildAst,
+  DEFAULT_NS,
+  SourcePosition,
+  XMLAttribute,
+  XMLElement
+} from "@xml-tools/ast";
 import {
   UI5SemanticModel,
   BaseUI5Node,
@@ -27,6 +35,7 @@ import {
   typeToString
 } from "@ui5-language-assistant/logic-utils";
 import { getNodeDocumentation } from "./documentation";
+import { Position } from "vscode-languageserver-types";
 
 export function getCompletionItems(
   model: UI5SemanticModel,
@@ -44,12 +53,27 @@ export function getCompletionItems(
     tokenVector: tokenVector
   });
 
-  return transformToLspSuggestions(suggestions, model);
+  const completionItems = transformToLspSuggestions(
+    suggestions,
+    model,
+    textDocumentPosition.position
+  );
+  const first = completionItems[0];
+  if (first && first.textEdit) {
+    console.log(
+      `Document position: ${textDocumentPosition.position.line}:${textDocumentPosition.position.character}`
+    );
+    console.log(
+      `Suggestion range: ${first.textEdit.range.start.line}:${first.textEdit.range.start.character}-${first.textEdit.range.end.line}:${first.textEdit.range.end.character}`
+    );
+  }
+  return completionItems;
 }
 
 function transformToLspSuggestions(
   suggestions: UI5XMLViewCompletion[],
-  model: UI5SemanticModel
+  model: UI5SemanticModel,
+  originalPosition: Position
 ): CompletionItem[] {
   const lspSuggestions = map(suggestions, suggestion => {
     const lspKind = computeLSPKind(suggestion);
@@ -60,10 +84,11 @@ function transformToLspSuggestions(
     if (suggestion.ui5Node.deprecatedInfo?.isDeprecated) {
       detailText = `(deprecated) ${detailText}`;
     }
+
     const completionItem: CompletionItem = {
       label: suggestion.ui5Node.name,
-      // TODO use textEdit instead of insertText to support replacing "sap.m.Button" with "Button"
-      insertText: createInsertText(suggestion),
+      filterText: getNodeFilterText(suggestion),
+      textEdit: createTextEdit(suggestion, originalPosition),
       insertTextFormat: InsertTextFormat.Snippet,
       detail: detailText,
       documentation: getNodeDocumentation(suggestion.ui5Node, model),
@@ -101,11 +126,18 @@ export function computeLSPKind(
   }
 }
 
-function createInsertText(suggestion: UI5XMLViewCompletion): string {
+function createTextEdit(
+  suggestion: UI5XMLViewCompletion,
+  originalPosition: Position
+): TextEdit {
+  let position: SourcePosition | undefined = undefined;
   let insertText = suggestion.ui5Node.name;
   switch (suggestion.type) {
     // Attribute key
     case "UI5NamespacesInXMLAttributeKey": {
+      position = getXMLAttributeKeyPosition(suggestion.astNode);
+      insertText = `xmlns:${insertText}`;
+
       // Auto-insert the selected namespace
       /* istanbul ignore else */
       if (suggestion.astNode.syntax.value === undefined) {
@@ -116,6 +148,8 @@ function createInsertText(suggestion: UI5XMLViewCompletion): string {
     case "UI5AssociationsInXMLAttributeKey":
     case "UI5EventsInXMLAttributeKey":
     case "UI5PropsInXMLAttributeKey": {
+      position = getXMLAttributeKeyPosition(suggestion.astNode);
+
       // Auto-insert ="" for attributes
       /* istanbul ignore else */
       if (suggestion.astNode.syntax.value === undefined) {
@@ -125,6 +159,7 @@ function createInsertText(suggestion: UI5XMLViewCompletion): string {
     }
     // Tag name
     case "UI5AggregationsInXMLTagName": {
+      position = getXMLTagNamePosition(suggestion.astNode);
       // Auto-close tag
       /* istanbul ignore else */
       if (suggestion.astNode.syntax.closeBody === undefined) {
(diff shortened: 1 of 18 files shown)


packages/xml-views-completion/src/providers/attributeValue/namespace.ts
```diff
@@ -29,7 +29,7 @@ export function namespaceValueSuggestions(
 
   const ui5Model = opts.context;
 
-  const attributeValue = xmlAttribute.value ?? "";
+  const attributeValue = opts.prefix ?? "";
```
bd82, line 32: can we have this in a different PR? 
This is a separated fix right?

I think we will lose CHANGELOG info
once this is squashed

packages/language-server/src/completion-items.ts
```diff
@@ -190,13 +256,63 @@ function getClassNamespacePrefix(
...
+      if (
+        !isEmpty(suggestion.astNode.ns) ||
+        (suggestion.astNode.name !== null &&
+          includes(suggestion.astNode.name, ":"))
+      ) {
+        const prefix = getClassNamespacePrefix(suggestion);
+        // The else case should never happen - if the user entered a namespace in the tag name only classes from
+        // that namespace are returned, so the namespace will be returned in the prefix
+        /* istanbul ignore else */
+        if (prefix !== undefined) {
```
bd82, line 289: should we put the check `getClassNamespacePrefix(suggestion) !== undefined` as part of the conditions in lines 281-283 ?

I know it means we will call it twice, but maybe that will allow the typescript compiler to understand that it is never undefined?

packages/language-server/src/completion-items.ts
```diff
@@ -190,13 +256,63 @@ function getClassNamespacePrefix(
   return undefined;
 }
 
+function getNodeFilterText(suggestion: UI5XMLViewCompletion): string {
```
bd82, line 259: What happens if we have no filterText at all? is there a default value?

bd82:
I need online sync to understand this filterText behavior.

packages/language-server/test/completion-items-spec.ts
```diff
@@ -25,28 +29,245 @@ describe("the UI5 language assistant Code Completion Services", () => {
...
+      `${description}: range end character`
+    ).to.be.at.least(position.character);
+  }
+
+  function assertFilterMatches(
+    filterText: string,
+    text: string,
+    description: string
+  ): void {
+    // This is a simple matcher - all characters in text are found in filterText in the same order.
```
bd82, line 67: What do you think about implementing this by dynamically creating a regExp?
It seems more declarative to me.
- /^B.*U.*T.*$/.test("BUTTON) // --> true

```
> "BUT".split("").join(".*")
'B.*U.*T'
```

bd82:
I will finish the review tomorrow after we sync

packages/xml-views-completion/src/providers/attributeValue/namespace.ts
tal-sapan: Ok... I did it in the same PR because I found the bug while I was writing the tests. I can put the tests in skip and re-enable them in PR with this fix.

packages/language-server/src/completion-items.ts
tal-sapan: Yes, the label is used (and then suggestions that don't contain the entire text that is being replaced, like FQN or namespace, will be filtered out)

packages/language-server/test/completion-items-spec.ts
tal-sapan: It would be a problem if there are characters that should be escaped (like dots), which will make it more complex than the iterative approach I think. I tried to do it with an array matcher (`to.include.ordered.members`) but that ignores duplicate entries.

packages/language-server/src/completion-items.ts
tal-sapan: I can do the assignment in the `if` so we won't call it twice... not sure it will be better. It would be less readable. I prefer to keep the ignore. (Also it's not a compilation issue, it's a logical issue with a reasonable fallback in case we have a bug).

packages/xml-views-completion/src/providers/attributeValue/namespace.ts
tal-sapan: Done

packages/language-server/src/completion-items.ts
```diff
@@ -27,6 +35,7 @@ import {
   typeToString
 } from "@ui5-language-assistant/logic-utils";
 import { getNodeDocumentation } from "./documentation";
+import { Position } from "vscode-languageserver-types";
```
bd82, line 38: import order

packages/language-server/src/completion-items.ts
```diff
@@ -60,10 +75,11 @@ function transformToLspSuggestions(
     if (suggestion.ui5Node.deprecatedInfo?.isDeprecated) {
       detailText = `(deprecated) ${detailText}`;
     }
+
     const completionItem: CompletionItem = {
       label: suggestion.ui5Node.name,
-      // TODO use textEdit instead of insertText to support replacing "sap.m.Button" with "Button"
-      insertText: createInsertText(suggestion),
+      filterText: getNodeFilterText(suggestion),
```
bd82, line 81: What would happen if filterText === textEdit always? what won't work?

packages/language-server/test/completion-items-spec.ts
```diff
@@ -25,28 +29,245 @@ describe("the UI5 language assistant Code Completion Services", () => {
...
+      replacedText: getTextInRange(xmlSnippet, suggestion.textEdit?.range)
+    }));
+    const suggestionKinds = uniq(
+      map(suggestions, suggestion => suggestion.kind)
+    );
+
+    expect(suggestionNames).to.deep.equalInAnyOrder([
+      {
+        label: "BusyDialog",
+        tagName: "m:BusyDialog",
```
bd82, line 169: why do we assert on the `tagName`? could/should we assert directly on the textEdit to help
readers of the code understand textEdit scenario from literal examples?

packages/language-server/test/completion-items-spec.ts
```diff
@@ -151,41 +487,149 @@ describe("the UI5 language assistant Code Completion Services", () => {
...
-      "upload",
-      "util",
-      "ux3",
-      "uxap",
-      "ubc",
-      "ui",
-      "ui", // This "ui" is in a different parent
-      "ui5",
-      "ulc"
+      {
```
bd82, line 552: this is a good example of a test I would like to know the textEdit in
to understand what is actually inserted once the user selects a suggestion

packages/language-server/src/completion-items.ts
tal-sapan: The filterText should contain the prefix (replaced text until the cursor position) and is used for sorting too (for example results that have the prefix as a substring rank higher than results that have additional characters in the middle)

packages/language-server/test/completion-items-spec.ts
tal-sapan: We can assert on the entire text edit (its `newText` property) but that includes things like closing brackets and tab stops. I think maybe we should have only a few tests that check these things specifically and most tests can check only the tag name (so that if we change the tab stops for example we won't need to change a lot of tests).

packages/language-server/test/completion-items-spec.ts
tal-sapan: I'll add it here

packages/language-server/test/completion-items-spec.ts
bd82: can we represent the `textEdit` in some simple string literal form?
The reasoning is to help understand the flow from a technical perspective.

packages/language-server/src/completion-items.ts
bd82: > The filterText should contain the prefix (replaced text until the cursor position) 
 
- Does it contain anything else?
- Could we compute it more simply using offsets and substring on `insertText` in that case?

packages/language-server/test/completion-items-spec.ts
tal-sapan: Done

packages/language-server/src/completion-items.ts
tal-sapan: Moved the logic to `createTextEdit` according to what we discussed (it didn't add as much code as I thought so I didn't refactor it)

packages/language-server/src/completion-items.ts
tal-sapan: Moved

bd82: I still see a relative local import followed by an external import

packages/language-server/src/completion-items.ts
```diff
@@ -101,13 +118,28 @@ export function computeLSPKind(
...
     // Attribute key
     case "UI5NamespacesInXMLAttributeKey": {
+      position = getXMLAttributeKeyPosition(suggestion.astNode);
+      insertText = `xmlns:${insertText}`;
+      // Namespace in xmlns attribute key should contain the "xmlns:" prefix
+      filterText = insertText;
+
       // Auto-insert the selected namespace
-      /* istanbul ignore else */
       if (suggestion.astNode.syntax.value === undefined) {
```
bd82, line 143: out of scope for this PR but it looks like we need to model the "equals" in the ASTNode as well.

tal-sapan: Yes. It currently causes a bug when you have the `=` sign without a value (the `=` sign remains after accepting the proposal, which contains a value).

packages/language-server/src/completion-items.ts
tal-sapan: Will move it in another PR

PR #63, opened 2020-04-26
feat(language-server): auto-insert namespace when selecting class

tal-sapan:
If the namespace of the class is not defined.
In case of collision add number suffix to the inserted name.

### DIFF
--- a/packages/language-server/src/completion-items.ts
+++ b/packages/language-server/src/completion-items.ts
@@ -1,4 +1,4 @@
-import { map, findKey } from "lodash";
+import { map, findKey, find, forEachRight } from "lodash";
 import {
   CompletionItem,
   CompletionItemKind,
@@ -8,13 +8,15 @@ import {
   Range
 } from "vscode-languageserver";
 import { TextDocument } from "vscode-languageserver-textdocument";
+import { Position } from "vscode-languageserver-types";
 import { parse, DocumentCstNode } from "@xml-tools/parser";
 import {
   buildAst,
   DEFAULT_NS,
   SourcePosition,
   XMLAttribute,
-  XMLElement
+  XMLElement,
+  XMLDocument
 } from "@xml-tools/ast";
 import {
   UI5SemanticModel,
@@ -35,7 +37,6 @@ import {
   typeToString
 } from "@ui5-language-assistant/logic-utils";
 import { getNodeDocumentation } from "./documentation";
-import { Position } from "vscode-languageserver-types";
 
 export function getCompletionItems(
   model: UI5SemanticModel,
@@ -76,12 +77,13 @@ function transformToLspSuggestions(
       detailText = `(deprecated) ${detailText}`;
     }
 
-    const textEditDetails = createTextEdit(suggestion, originalPosition);
+    const textEditDetails = createTextEdits(suggestion, originalPosition);
     const completionItem: CompletionItem = {
       label: suggestion.ui5Node.name,
       filterText: textEditDetails.filterText,
       textEdit: textEditDetails.textEdit,
       insertTextFormat: InsertTextFormat.Snippet,
+      additionalTextEdits: textEditDetails.additionalTextEdits,
       detail: detailText,
       documentation: getNodeDocumentation(suggestion.ui5Node, model),
       kind: lspKind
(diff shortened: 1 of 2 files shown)


packages/language-server/src/completion-items.ts
```diff
@@ -118,11 +120,15 @@ export function computeLSPKind(
   }
 }
 
-function createTextEdit(
+function createTextEdits(
```
bd82, line 123: This function is now ~120 lines, We keep trying to avoid it buy I think it is time to split it up...

packages/language-server/src/completion-items.ts
```diff
@@ -281,6 +338,35 @@ function getClassNamespacePrefix(
...
+  let parent: XMLElement | XMLDocument = xmlElement;
+  while (parent.type !== "XMLDocument") {
+    parent = parent.parent;
+  }
+  let positionLine = 0;
+  let positionColumn = 0;
+  // The 'else' part will never happen because to get suggestions on class name the open tag must exist so the
+  // xml document will not be empty
+  /* istanbul ignore else */
+  if (parent.rootElement !== null) {
```
bd82, line 356: lets avoid doing any text edit at runtime if our data/xml/model is invalid.

packages/language-server/test/completion-items-spec.ts
```diff
@@ -40,6 +40,15 @@ describe("the UI5 language assistant Code Completion Services", () => {
     return result?.[1];
   }
 
+  /** Return the attributes from a tag name suggestion insert text  */
+  function getAttributes(textEdit: TextEdit | undefined): string[] {
```
bd82, line 44: if you ever need more accurate parsing you can
use the XML Parser from XML-Tools to parse only an attribute's text.
It is not limited to starting at the root

packages/language-server/test/completion-items-spec.ts
```diff
@@ -160,6 +169,7 @@ describe("the UI5 language assistant Code Completion Services", () => {
     const suggestionNames = map(suggestions, suggestion => ({
```
bd82, line 169: this is no longer `suggestionNames`

packages/language-server/test/completion-items-spec.ts
```diff
@@ -170,24 +180,52 @@ describe("the UI5 language assistant Code Completion Services", () => {
...
       }
     ]);
 
     expect(suggestionKinds).to.deep.equal([CompletionItemKind.Class]);
   });
 
+  it("will insert class namespace with a new name when another namespace with the short name is defined", () => {
+    const xmlSnippet = `<sap.m.BusyI⇶ xmlns:m="sap.ui.core.mvc"`;
+    const suggestions = getSuggestions(xmlSnippet);
+    const suggestionNames = map(suggestions, suggestion => ({
```
bd82, line 200: `suggestionsDescriptor`?

packages/language-server/test/completion-items-spec.ts
```diff
@@ -170,24 +180,52 @@ describe("the UI5 language assistant Code Completion Services", () => {
...
         tagName: "m:BusyIndicator",
+        firstAttribute: `xmlns:m="sap.m"`,
         replacedText: "sap.m.Busy"
       }
     ]);
 
     expect(suggestionKinds).to.deep.equal([CompletionItemKind.Class]);
   });
 
+  it("will insert class namespace with a new name when another namespace with the short name is defined", () => {
```
bd82, line 197: the complexity of what this test does definitively warrants snapshot testing (in the future)

packages/language-server/src/completion-items.ts
tal-sapan: Done

packages/language-server/src/completion-items.ts
tal-sapan: I extracted the handling for classes which was the most complex

packages/language-server/test/completion-items-spec.ts
tal-sapan: Chaned to `suggestionsDetails` in all tests

tal-sapan:
I'm merging so I can use the changes here in another change, let me know if you think there's still something that should be changed

PR #64, opened 2020-04-27
refactor(language-server): move classes code completion tests to other file

tal-sapan:
The tests did not change except for one test that uses a new utility function to get the range for the `additionalTextEdits`:
`"will insert the namespace when selecting completion for class in inner tag and namespace is not defined"`

Utility functions moved to their own file.
Changes to utility functions:
* Added assertion that text edits don't overlap (`assertRangesDoNotOverlap`)
* Added model parameter to `getSuggestions` (since it's no longer in a closure)
* Added handling of `additionalTextEdits` ranges in `getSuggestions` 
* Added function `getRanges`
* Extracted function `getXmlSnippetDocument` from `getSuggestions`

### DIFF
--- a/packages/language-server/test/completion-items-utils.ts
+++ b/packages/language-server/test/completion-items-utils.ts
@@ -1,0 +1,234 @@
+import { forEach } from "lodash";
+import { expect } from "chai";
+import {
+  TextDocument,
+  TextDocumentPositionParams,
+  Position,
+  TextDocumentIdentifier,
+  CompletionItem,
+  TextEdit,
+  Range
+} from "vscode-languageserver";
+import { UI5SemanticModel } from "@ui5-language-assistant/semantic-model-types";
+import { expectExists } from "@ui5-language-assistant/test-utils";
+import { getCompletionItems } from "../src/completion-items";
+
+/** Return the first part of a tag name suggestion insert text */
+export function getTagName(textEdit: TextEdit | undefined): string | undefined {
+  if (textEdit === undefined) {
+    return undefined;
+  }
+  const result = /^([^> ]*)/.exec(textEdit.newText);
+  return result?.[1];
+}
+
+/** Use ⇶ to mark the cursor position */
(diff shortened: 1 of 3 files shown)


packages/language-server/test/completion-items-utils.ts
```diff
@@ -0,0 +1,234 @@
...
+
+/** Use ⭲ to mark range start and ⭰ to mark range end */
+export function getRanges(xmlSnippet: string): Range[] {
+  const RANGE_START_CHAR = "⭲";
+  const RANGE_END_CHAR = "⭰";
+
+  const { document } = getXmlSnippetDocument(xmlSnippet);
+  const ranges: Range[] = [];
+  let xmlTextWithRanges = xmlSnippet.replace("⇶", "");
+  while (xmlTextWithRanges.indexOf(RANGE_START_CHAR) !== -1) {
```
bd82, line 72: I would have attempted to implement this with a **none greedy** and **sticky** regExp
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/sticky

- > https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Quantifiers (look for "greedy" word)

But it is clear enough so no need to change

packages/language-server/test/completion-items-utils.ts
```diff
@@ -0,0 +1,234 @@
...
+    range.start.character,
+    `${description}: range start character`
+  ).to.be.at.most(position.character);
+  expect(
+    range.end.character,
+    `${description}: range end character`
+  ).to.be.at.least(position.character);
+}
+
+function assertRangesDoNotOverlap(
```
bd82, line 150: would this logic be much simpler if we **first** converted all these positions to offsets instead of line/colums?

tal-sapan: Done

PR #65, opened 2020-04-27
feat(language-server): auto-replace closing tag name

tal-sapan:
Also refactor the classes code completion tests to have a function with most of the logic.


bd82:
I think we should first investigate solving the general case of aligning open and close tag names, perhaps this will make this solution redundant...

tal-sapan:
> I think we should first investigate solving the general case of aligning open and close tag names, perhaps it would make this solution redundant...

In any case we can't use the generic xml plugin for *.view.xml files until we remove the fiori xml file type which will only happen after we have the validations

packages/language-server/src/completion-items.ts
```diff
@@ -252,6 +257,43 @@ function createTextEditsForClassInTagName(
...
+function addClosingTagTextEdit(
+  xmlElement: XMLElement,
+  closingTagName: string,
+  additionalTextEdits: TextEdit[]
+): void {
+  // Note: if we implement syncronization between the opening and closing tag in the generic XML extension
+  // and it will cover changes done by code assist suggestions this logic should be removed from this extension
+  if (xmlElement.syntax.closeName !== undefined) {
+    // Replace name in closing tag.
+    // Note: in case of invalid xml where the current tag is closed but has no closing tag,
```
bd82, line 269: so you need to know if its self closing?

packages/language-server/src/completion-items.ts
```diff
@@ -252,6 +257,43 @@ function createTextEditsForClassInTagName(
   return { range, newText, filterText };
 }
 
+function addClosingTagTextEdit(
+  xmlElement: XMLElement,
+  closingTagName: string,
+  additionalTextEdits: TextEdit[]
```
bd82, line 263: Please avoid **mutating** `additionalTextEdits` in this function.
instead return an array of additional textEdits and concat it **outside** this function.

This is a more functional style like most of our other code here and will make it easier to follow the logical flow.

packages/language-server/src/completion-items.ts
bd82: https://github.com/SAP/xml-tools/blob/master/packages/ast/CHANGELOG.md#410-2020-05-05

packages/language-server/test/completion-items-spec.ts
```diff
@@ -251,6 +252,189 @@ describe("the UI5 language assistant Code Completion Services", () => {
...
+          }
+        ],
+        replacedText: "te"
+      }
+    ]);
+
+    expect(suggestionKinds).to.deep.equal([CompletionItemKind.Text]);
+  });
+
+  it("will replace the class closing tag name when the tag is closed and has a different name from the opening tag", () => {
```
bd82, line 316: maybe we should not replace in this case.
Basically be very careful to only replace when both tags are aligned to reduce
the chance we are replacing on an unrelated tag.

This is similar to how I think tag syncing in XML-Tools should happen.
I prefer to be 99.99% correct when replacing at the cost of only editing for 90% of use cases.

packages/language-server/test/completion-items-spec.ts
```diff
@@ -251,6 +252,189 @@ describe("the UI5 language assistant Code Completion Services", () => {
...
+          }
+        ],
+        replacedText: "te"
+      }
+    ]);
+
+    expect(suggestionKinds).to.deep.equal([CompletionItemKind.Text]);
+  });
+
+  it("will replace the class closing tag name when the tag is closed and does not have a name", () => {
```
bd82, line 377: replacing in this scenario is o.k. because there is obviously an error
so we are trying to fix it, we won't make the situation any worse.

packages/language-server/test/copmletion-items-classes-spec.ts
```diff
@@ -21,6 +25,62 @@ describe("the UI5 language assistant Code Completion Services - classes", () =>
     ui5SemanticModel = await generateModel({ version: "1.74.0" });
   });
 
+  /** The first (but not final) place the custor stops when inserting the completion. Pressing tab moves it to the next place. */
+  const TAB_STOP1 = "${1}";
+
+  interface ClassCompletionItem {
```
bd82, line 31: was not this code in another PR? does this PR need to be updated? I will try to do so

packages/language-server/test/completion-items-spec.ts
```diff
@@ -11,7 +11,8 @@ import { computeLSPKind } from "../src/completion-items";
 import {
```
bd82, line 11: What % of time did you work on the productive logic versus the tests?
Feels like we could use snapshot tests here.
Maybe I could evaluate those with the diagnostics scenario first.

packages/language-server/src/completion-items.ts
tal-sapan: No, I mean this use case:
```xml
<parentTag>
  <innerTag>
</parentTag>
```

`</parentTag>` is considered the closing tag of `<innerTag>`

packages/language-server/src/completion-items.ts
tal-sapan: Done

packages/language-server/test/completion-items-spec.ts
tal-sapan: Ok, changed

PR #66, opened 2020-04-30
feat: add support for code completion of boolean properties values

tal-sapan:


packages/xml-views-completion/api.d.ts
```diff
@@ -122,3 +129,38 @@ export interface UI5NamespacesInXMLAttributeValueCompletion
   extends BaseXMLViewCompletion<XMLAttribute, UI5Namespace> {
   type: "UI5NamespacesInXMLAttributeValue";
 }
+
+export type LiteralXMLViewCompletionBaseType = boolean | string | number;
```
bd82, line 133: why can't we simply represent the literal as a plain string ?
Why do we need to distinguish between boolean/string/number here?

packages/language-server/src/completion-items.ts
```diff
@@ -199,8 +201,18 @@ function createTextEdits(
       // the attribute value syntax exists
       /* istanbul ignore next */
       range = getXMLAttributeValueRange(suggestion.astNode) ?? range;
-      newText = `"${suggestion.ui5Node.name}"`;
       // Attribute values should contain quotation marks
+      newText = `"${newText}"`;
+      filterText = newText;
+      break;
+    }
+    case "BooleanValueInXMLAttributeValueCompletion": {
```
bd82, line 209: what would be the difference between suggesting boolean/number/string ? why is a special case needed?

bd82: If its the quotes they can be part of the literal suggestions could they not be?

bd82: actually they probably should be part of the literal suggestion as if I read this correctly we will always offer double quotes here even if the user started entering single quotes?

packages/xml-views-completion/src/api.ts
```diff
@@ -24,9 +29,40 @@ export function getXMLViewCompletions(
...
+    suggestions,
+    _ =>
+      isLiteralXMLViewCompletion(_) ||
+      includes(allowedVisibility, _.ui5Node.visibility)
   );
 
   return publicAndProtectedSuggestions;
 }
+
+export function isLiteralXMLViewCompletion(
```
bd82, line 42: could we model the boolean as a UI5 node and avoid having a different structure causing multiple new branches of the already complex code?

could not a boolean "true" be represented a "UI5PrimitiveBooleanLiteral" with `name` "true" ?
and no description / deprecated info / additional info.

packages/xml-views-completion/src/providers/attributeValue/literal.ts
```diff
@@ -0,0 +1,51 @@
...
+      name: "true",
+      value: true
+    },
+    {
+      name: "false",
+      value: false
+    }
+  ];
+  const prefix = opts.prefix ?? "";
+  const prefixMatchingValues = filterMembersForSuggestion(values, prefix, []);
```
bd82, line 42: would this work for single quotes?

packages/xml-views-completion/src/providers/attributeValue/literal.ts
```diff
@@ -0,0 +1,51 @@
...
+  const ui5Property = find(properties, ["name", xmlAttribute.key]);
+  const propType = ui5Property?.type;
+  if (propType?.kind !== "PrimitiveType" || propType.name !== "Boolean") {
+    return [];
+  }
+
+  const values = [
+    {
+      name: "true",
+      value: true
```
bd82, line 34: as mentioned earlier, I am uncertain why we care about the semantic expression value
in a syntactic (completion) context.

would the name suffice?

packages/xml-views-completion/src/providers/attributeValue/literal.ts
bd82: Do we need the filter utility or could we use a regExp to check? which would be simpler?

packages/xml-views-completion/api.d.ts
```diff
@@ -122,3 +129,38 @@ export interface UI5NamespacesInXMLAttributeValueCompletion
...
+}
+
+export type LiteralXMLViewCompletion = BooleanValueInXMLAttributeValueCompletion;
+
+export interface BooleanValueInXMLAttributeValueCompletion
+  extends BaseLiteralXMLViewCompletion<XMLAttribute, PrimitiveType, boolean> {
+  type: "BooleanValueInXMLAttributeValueCompletion";
+}
+
+export function isLiteralXMLViewCompletion(
```
bd82, line 161: If we need to expose public API utils to distinguish the suggestions types
we may have too many types of suggestions.

packages/xml-views-completion/src/api.ts
bd82: The UI5 node could be dynamically created.

bd82: Can you think of another way to model this so there are not two different structures?
Its fine to even refactor the existing flows.

Random thoughts:
- Could/Should suggestions include methods and not only data (I am not fond of this but maybe it would help?)
- Would we have need more than a simple string(attribute) or identifier(tag) literal suggestion? even in a binding expression you normally only offer the next identifier not a FQN (at least in most languages).

packages/xml-views-completion/src/api.ts
```diff
@@ -24,9 +29,40 @@ export function getXMLViewCompletions(
...
+  suggestion: UI5XMLViewCompletion
+): suggestion is UI5NodeXMLViewCompletion {
+  // Suggestions are either literals or UI5 nodes
+  if (isLiteralXMLViewCompletion(suggestion)) {
+    return false;
+  }
+  // This is here so we'll get compilation errors if the type changes
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const _: UI5NodeXMLViewCompletion = suggestion;
+  return true;
```
bd82, line 67: The logic here is only correct if you combine design time and runtime knowledge.
Anyone calling it from plain JavaScript (which they could because it is a public API)
is going to get unexpected results

packages/xml-views-completion/api.d.ts
tal-sapan: It's all strings at the end because they are an xml attribute values which are written to a text document, but string values for example should be escaped while boolean and numbers don't.
I think it's more accurate to use the original type. Technically they are converted to string afterwards but I don't think the conversion should be here. (Like we keep the original value for default values in properties.)

packages/language-server/src/completion-items.ts
tal-sapan: Are single quotes valid in xml attributes?
We put double quotes on all suggested attribute values (including enum member and namespaces), replacing the original quotes... I did this because the token of the value includes the quotes, but if we add a token for the inner string we can change this.

tal-sapan: > what would be the difference between suggesting boolean/number/string ? why is a special case needed?

No real difference (maybe except escaping string values when we have them), but it's like we have different cases when suggesting properties, events and aggregations.
Technically I think the icons might be different. Boolean values are keywords or constants (I used constants because I don't think they are considered keywords in xml), while strings and numbers are values.

packages/xml-views-completion/src/api.ts
tal-sapan: It doesn't sound right to me to force literals to be UI5 nodes... they aren't. They're not part of the semantic model, don't exist on a namespace or have a parent, don't have a corresponding definition or exist in the api.json in any way.
We can model the literal "node" to look a bit like a UI5 node (put the `name` , `type` and `value` in a structure instead of being separate) which could make the code a bit simpler (though not by much - all the documentation for example would still have to be handled separately). We'll have to change the property name on the suggestion from `ui5Node` because it would no longer be a ui5 node.
If we want to change the structure to be completely agnostic to UI5 nodes we could return `CompletionItem`s... (I don't think this is a good idea, but it would take care of the difference in structures).

bd82:
lets sync on this in person

packages/xml-views-completion/src/providers/attributeValue/literal.ts
tal-sapan: We need to filter by the prefix, and I prefer to do it with the same utility function in case we change the logic in the future (e.g. to be more vscode-like - all characters included instead of substring)

packages/xml-views-completion/src/providers/attributeValue/literal.ts
tal-sapan: I think there should be a separation between the suggestion name (label) and value. They might not always be the same. The name can be more meaningful than the value when it's not just arbitrary values.

packages/xml-views-completion/src/api.ts
tal-sapan: The compilation error mentioned is internal in this package. The logic will be correct no matter how you call it, but if we add a literal type this will make sure we don't forget to update this function.

tal-sapan:
> What other scenarios would there be for literal suggestions?

Nothing concrete that I know of currently, but maybe in the future:
* The original default value of the property
* Current date/time/file name/app name etc
* Generated GUID/random number
* Generated ID for the `id` attribute (so it will have a unique stable ID)
* Strings from i18n (though we might handle that differently)

tal-sapan:
@bd82 I think I covered everything we discussed, you can review now.

packages/language-server/src/completion-items.ts
```diff
@@ -230,8 +227,18 @@ function createTextEdits(
...
-      newText = `"${suggestion.ui5Node.name}"`;
       // Attribute values should contain quotation marks
+      newText = `"${newText}"`;
+      filterText = newText;
+      break;
+    }
+    case "BooleanValueInXMLAttributeValue": {
+      // The 'else' part will never happen because to get suggestions for attribute value, the "" at least must exist so
+      // the attribute value syntax exists
+      /* istanbul ignore next */
```
bd82, line 238: do we need ignore `next` here instead of ignore `else`?

packages/xml-views-completion/src/providers/attributeValue/boolean-literal.ts
```diff
@@ -0,0 +1,54 @@
...
+    return [];
+  }
+  const properties = flattenProperties(elementClass);
+  const ui5Property = find(properties, ["name", xmlAttribute.key]);
+  const propType = ui5Property?.type;
+  if (propType?.kind !== "PrimitiveType" || propType.name !== "Boolean") {
+    return [];
+  }
+
+  const values: BooleanValue[] = [
```
bd82, line 34: subjective:
This is kind of the "model" for us here, and the function does the logic.
Should we define it just above the function? as it never changes?

packages/xml-views-completion/src/providers/attributeValue/boolean-literal.ts
```diff
@@ -0,0 +1,54 @@
...
+    {
+      kind: "BooleanValue",
+      name: "false",
+      value: false,
+    },
+  ];
+  const prefix = opts.prefix ?? "";
+  const prefixMatchingValues = filterMembersForSuggestion(values, prefix, []);
+
+  return map(prefixMatchingValues, (_) => ({
```
bd82, line 49: subjective:

From debugging perspective it may be best to assign this result to a variable
and then return it in a separate statement.

packages/language-server/src/completion-items.ts
tal-sapan: It doesn't work with else

packages/xml-views-completion/src/providers/attributeValue/boolean-literal.ts
tal-sapan: Done

packages/xml-views-completion/src/providers/attributeValue/boolean-literal.ts
tal-sapan: Done

PR #67, opened 2020-05-03
feat(xml-views-validation): skeleton

bd82:


packages/xml-views-validation/api.d.ts
```diff
@@ -0,0 +1,26 @@
...
+export function validateXMLView(opts: {
+  model: UI5SemanticModel;
+  xmlView: XMLDocument;
+}): UI5XMLViewIssue[];
+
+export interface UI5XMLViewIssueStructure {
+  kind: string;
+  message: string;
+  severity: "hint" | "info" | "warn" | "error";
+  range: { start: number; end: number };
```
tal-sapan, line 13: Are `start` and `end` offsets? Could you name them as such (`startOffset` and `endOffset`)?

packages/xml-views-validation/api.d.ts
```diff
@@ -0,0 +1,26 @@
+import { XMLDocument } from "@xml-tools/ast";
+import { UI5SemanticModel } from "@ui5-language-assistant/semantic-model-types";
+
+export function validateXMLView(opts: {
+  model: UI5SemanticModel;
+  xmlView: XMLDocument;
+}): UI5XMLViewIssue[];
+
+export interface UI5XMLViewIssueStructure {
```
tal-sapan, line 9: Maybe rename this to `BaseUI5XMLViewIssue`

packages/xml-views-validation/src/validate-xml-views.ts
```diff
@@ -0,0 +1,55 @@
...
+  XMLAstVisitor,
+  XMLAttribute,
+  XMLElement,
+  accept
+} from "@xml-tools/ast";
+import { UI5SemanticModel } from "@ui5-language-assistant/semantic-model-types";
+import { UI5XMLViewIssue } from "../api";
+
+export interface UI5Validators {
+  element: ((
```
tal-sapan, line 13: Do we want also document-level validators (for example: to verify that there is a root element, or that the root element type is `mvc:View`)?

packages/xml-views-validation/src/validate-xml-views.ts
```diff
@@ -0,0 +1,55 @@
...
+  XMLDocument,
+  XMLAstVisitor,
+  XMLAttribute,
+  XMLElement,
+  accept
+} from "@xml-tools/ast";
+import { UI5SemanticModel } from "@ui5-language-assistant/semantic-model-types";
+import { UI5XMLViewIssue } from "../api";
+
+export interface UI5Validators {
```
tal-sapan, line 12: Do you think there might be a need for validators that use a visitor to check that something exists in the document? Not just in relation to a specific node (like that associations point to an existing element, or if a namespace declaration is unused - these should be in attribute value validators but will have to use a visitor or other means of finding something in the xml), but in general? I don't have a use case right now... maybe it's more something to consider when we have a use case.

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
```diff
@@ -0,0 +1,54 @@
...
+): UseOfDeprecatedClassIssue[] {
+  const elementTagFqn = xmlToFQN(xmlElement);
+  const ui5Class = model.classes[elementTagFqn];
+  if (ui5Class === undefined) {
+    return [];
+  }
+
+  if (ui5Class.deprecatedInfo !== undefined) {
+    const issues: UseOfDeprecatedClassIssue[] = [];
+    const commonIssueFields: UseOfDeprecatedClassIssue = {
```
tal-sapan, line 18: If you don't define it as `UseOfDeprecatedClassIssue` you won't have to cast to `any` and typescript will force you to add the range property

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
```diff
@@ -0,0 +1,54 @@
...
+      kind: "UseOfDeprecatedClass",
+      message: `Deprecated UI5 Class: ${elementTagFqn} used.`,
```
tal-sapan, line 20: Did you copy the message from FS? If not I think `Class ${elementTagFqn} is deprecated` sound better

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
```diff
@@ -0,0 +1,54 @@
...
+      issues.push({
+        ...commonIssueFields,
+        range: {
+          start: xmlElement.syntax.openName.startOffset,
+          end: xmlElement.syntax.openName.endOffset
+        }
+      });
+    }
+
+    if (xmlElement.syntax.closeName !== undefined) {
```
tal-sapan, line 40: Not sure we need the issue in the closing tag too... also the check in the beginning is for the name in the opening tag, we don't actually check there what the closing tag is (in case it's mismatched). And if it's the same the user will already get the issue for the opening tag.

packages/xml-views-validation/src/validators/attributes/unknown-enum-value.ts
```diff
@@ -0,0 +1,57 @@
...
+    return [];
+  }
+
+  const possibleEnumValues = map(propType.fields, _ => _.name);
+  if (find(possibleEnumValues, _ => _ === actualAttributeValue) === undefined) {
+    const possibleValuesWithQuotes = map(possibleEnumValues, _ => `"${_}"`);
+    return [
+      {
+        kind: "UnknownEnumValue",
+        message: `Unknown enum value: ${
```
tal-sapan, line 44: Did you copy the message from FS?

packages/xml-views-validation/api.d.ts
bd82: I thought it may be redundant but now I see in LSP it is a "Position" for start/end.
Will Change

packages/xml-views-validation/api.d.ts
bd82: o.k.

packages/xml-views-validation/test/validators/attributes/unknown-enum-value-spec.ts
```diff
@@ -0,0 +1,135 @@
...
+              end: 131
+            },
+            severity: "error"
+          }
+        ]);
+      }
+    });
+  });
+
+  context("negative edge cases", () => {
```
tal-sapan, line 50: Add a negative test where the attribute name is unknown (and it's inside a class tag)

packages/xml-views-validation/test/validators/element/use-of-deprecated-class-spec.ts
```diff
@@ -0,0 +1,144 @@
...
+
+describe("the use of deprecated class validation", () => {
+  let ui5SemanticModel: UI5SemanticModel;
+
+  before(async function() {
+    this.timeout(GEN_MODEL_TIMEOUT);
+    ui5SemanticModel = await generateModel({ version: "1.74.0" });
+  });
+
+  context("true positive scenarios", () => {
```
tal-sapan, line 18: You didn't add a context for the positive tests for enum... please remove this (or add one to the other tests to be consistent)

packages/xml-views-validation/src/validate-xml-views.ts
bd82: The base visitor does not distinguish between the root element and other elements.
It could be done, however the root node is actually an "XmlDocument" node
which we can easily add support for, this may be needed for "stable id" validation, but I don't know the scenario well enough for now.

Regarding "root **element** validation": I am not sure, an element validation could check if the element is the Root before performing its checks, so such an API may not be needed. 

Your two examples may(???) not be valid validations, e.g: 
- If there is no root element (empty xml), what are we validating?
- The root element could also be a fragment, or perhaps a fragment could start with any root element? 

I would wait for expending the API until we have more specific scenarios.

packages/xml-views-validation/test/validators/element/use-of-deprecated-class-spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    it("will detect usage of a deprecated class", () => {
```
tal-sapan, line 19: Add a test where there is no closing tag (so we know that the warning is returned for invalid syntax too). And maybe one where the tag itself isn't closed (`<Button`)?

packages/xml-views-validation/src/validate-xml-views.ts
bd82: I think there would be such scenarios, but it may be possible to model these with the existing API, e.g:
- unused namespace warning:
  - Start at attribute which is `xmlns`
  - go to the parent XMLDocument (.parent.parent.parent...parent)
  - Start a new Visitor for tags only to see a use.

Obviously this is not the most efficient, but it logically keeps the different validations very separated / de-coupled. I would like to see an actual performance issues before trying to optimize it (which may lose the separation).

Another possibility is to run a validator on every tag that only collects and fill some state, (e.g maps of all used + declared xmlns prefixes), and then run a post traversal validation on the collected state, that would be more efficient, and also keep the separation.

Seems easy enough to expend the API when we will need it...

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
bd82: There is a problem with inference of string sub types if I do that:
image

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
bd82: Good suggestion

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
bd82: - Jetbrains does line-through on deprecated XML Tags even on the closing tag.
  which is a kind of validation (I don't have a problem view example to check)
- The open and close tags may be far from each other visually so there is a benefit to showing both, especially after we implement sync of the tags.
- It may be best to only show one error in the problem view, but I am not sure if the LSP apis support an error with multiple locations I will check

packages/xml-views-validation/src/validators/attributes/unknown-enum-value.ts
bd82: No. I will have a look

packages/xml-views-validation/test/api-spec.ts
```diff
@@ -0,0 +1,43 @@
...
+  });
+
+  it("will detect semantic UI5 xml view issues (smoke)", () => {
+    const xmlSnippet = `
+          <mvc:View
+            xmlns:mvc="sap.ui.core.mvc"
+            xmlns="sap.m"
+            xmlns:commons="sap.ui.commons">
+            >            
+            <List showSeparators = "Inner💩">
```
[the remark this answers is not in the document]
bd82: Pile of Poo is very distinctive!

packages/xml-views-validation/test/validators/attributes/unknown-enum-value-spec.ts
```diff
@@ -0,0 +1,135 @@
...
+      },
+      assertion: issues => {
+        expect(issues).to.deep.equal([
+          {
+            kind: "UnknownEnumValue",
+            message:
+              'Unknown enum value: "Inner💩", expecting one of: ["All", "Inner", "None"].',
+            range: {
+              start: 123,
+              end: 131
```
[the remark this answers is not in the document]
tal-sapan: Actually that one is already merged. Could you use it? The offset is a bit different in xml utils than in vscode, I think you'll have to add 1 to the start offset or something like that (the start offset in vscode is before the character and end offset is after the character, and the indexes are 0-based). Add a util function to convert it.

packages/xml-views-validation/test/api-spec.ts
```diff
@@ -0,0 +1,43 @@
...
+describe("the ui5 xml views validations API", () => {
+  let ui5SemanticModel: UI5SemanticModel;
+
+  before(async function() {
+    this.timeout(GEN_MODEL_TIMEOUT);
+    ui5SemanticModel = await generateModel({ version: "1.74.0" });
+  });
+
+  it("will detect semantic UI5 xml view issues (smoke)", () => {
+    const xmlSnippet = `
```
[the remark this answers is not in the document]
bd82: Will add

packages/xml-views-validation/test/validators/attributes/unknown-enum-value-spec.ts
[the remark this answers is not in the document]
bd82: Good point, I wonder how to do it without making the xml snippet invalid

packages/xml-views-validation/test/validators/attributes/unknown-enum-value-spec.ts
bd82: Will add

packages/xml-views-validation/test/validators/element/use-of-deprecated-class-spec.ts
bd82: will do

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
tal-sapan: Ok... there is a workaround which looks weird, but I think might be a bit less bad than using undefined (whether cast to `any` or `never`) because the workaround is closed to the definition:
image

packages/xml-views-validation/src/validate-xml-views.ts
tal-sapan: > The base visitor does not distinguish between the root element and other elements.
> It could be done, however the root node is actually an "XmlDocument" node
> which we can easily add support for, 

Yes, I meant for `XMLDocument`. Since it holds the root element the validation could be on it.

> this may be needed for "stable id" validation, but I don't know the scenario well enough for now.

I think stable IDs validation is not related, it's a validation that an `id` attribute exists (and possibly has a unique value in the document - but this is probably better to do in a different validation) except in some specific cases.

> Regarding "root **element** validation": I am not sure, an element validation could check if the element is the Root before performing its checks, so such an API may not be needed.

It's possible, but less efficient and logically these validations belongs on the document root I think.

> Your two examples may(???) not be valid validations, e.g:

Considering I just made them up it's very possible :)

> * If there is no root element (empty xml), what are we validating?

That the document isn't empty, I guess. I would say that normally this belongs in the generic xml validations but if a regular xml can be empty then it's specific (assuming xml views can't be empty, which should be checked of course, before we add a validation like this).

> * The root element could also be a fragment, or perhaps a fragment could start with any root element?

Yes, you would need to check the file name. But we have the document so maybe we can implement it. I think for fragments the root element is `sap.ui.core.FragmentDefinition`.

> I would wait for expending the API until we have more specific scenarios.

Ok

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
tal-sapan: Actually I think they do and this might be the best solution (it will also be required for other validations - like unique IDs). It's called `relatedInformation` on the `Diagnostic` interface (the equivalent of `CompletionItem`. By the way, this also has a tags property where you can mark the diagnostic locations as deprecated or unnecessary/unreachable, and that seems to be supported in Theia.

packages/xml-views-validation/test/validators/attributes/unknown-enum-value-spec.ts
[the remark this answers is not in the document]
tal-sapan: In the code we remove the characters before we create the document (like the arrows symbol for code completion cursor location). If you mean in the xml editor for snapshot tests, it won't be valid, but I don't think it matters much.

packages/xml-views-validation/api.d.ts
bd82: I will change the property name to `offsetRange` to avoid confusion with LSP's range
and to avoid duplication.

packages/xml-views-validation/src/validate-xml-views.ts
bd82: > That the document isn't empty

An empty XML document, or one without a root element is actually a syntax error.
- https://github.com/SAP/xml-tools/blob/master/packages/parser/lib/parser.js#L33

> It's possible, but less efficient and logically these validations belongs on the document root I think.

If checking the top level `XMLDocument` is fine, we can easily add it to the visitor. I agree it is more logical.

> Yes, you would need to check the file name. But we have the document so maybe we can implement it. I think for fragments the root element is sap.ui.core.FragmentDefinition.

Interesting, perhaps additional context would be needed in this case, we need to know if its a fragment or a view, but we can easily compose a different set of check functions for each scenario.

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
bd82: Thanks this is indeed better, but still looks a bit like dark magic 🧙‍♀️

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
bd82: The message in the original validations is not precise:
https://github.wdf.sap.corp/DevX/xmlvalidator/blob/master/src/fioriAnalysis/rules/XML_DEPRECATION.js#L79-L83

As the deprecated tag is a UI5 Class not always a control (may also be an element).

I will try to add the additional info to the message as well.

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
tal-sapan: Looks like types of properties in object literals are widened if there is no contextual information... I guess this adds the contextual information. I just wish we would have gotten an error if the `as` didn't match the value, but we can't win them all. (It would be possible to get an error if we defined an interface just for this object literal but it looks like overkill).

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
tal-sapan: Maybe if we change "control" to "class". I like the idea of adding the `since` and `message` to the issue.

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
bd82: Currently I am changing "range" to "offsetRange**s**" to model multiple problems
on the same issue, this should suffice to translate it to LSP "relatedInformation" in the Language Server.

In the future we may(???) want slightly different information for each "duplicated" issue, if so we will expend the API at the point.

packages/xml-views-validation/src/validators/attributes/unknown-enum-value.ts
bd82: I don't see that we had such a check in the past:
- https://github.wdf.sap.corp/DevX/xmlvalidator/tree/master/src/fioriAnalysis/rules

packages/xml-views-validation/test/api-spec.ts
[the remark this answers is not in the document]
bd82: done, used TYPO instead

packages/xml-views-validation/test/validators/attributes/unknown-enum-value-spec.ts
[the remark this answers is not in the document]
bd82: It is not a huge thing, but I like the jetbrains XML Editor working in an XML Snippet which is part of a JS string literal.

Perhaps we could mark the ranges **inside** the strings using some unicode char which is also valid in identifiers and then compute the range including the quotes.

packages/xml-views-validation/test/validators/element/use-of-deprecated-class-spec.ts
bd82: I add a single test where the tag is not closed `<Button`
because it is a **functional scenario** where a user is typing manually (without content assist)
and we want to give the user feedback ASAP so its worth testing it.

However I don't want to test too many invalid syntax scenarios in validations flow, this is the concern of the AST package in XML-Tools

bd82:
Thanks for in-depth review. 👍 
I believe all the comments have been addressed.
some additional upgrades to the API may occur when we start implementing more validations...

packages/xml-views-validation/src/validators/attributes/unknown-enum-value.ts
tal-sapan: Okay.
We should make a list of all the strings displayed to the user and review them with UA. Should probably create an activity so we don't forget to do it.

packages/xml-views-validation/test/validators/attributes/unknown-enum-value-spec.ts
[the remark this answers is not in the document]
tal-sapan: Not all of them are inside identifiers, but if you find characters that make sense I don't have a problem with changing them.

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
```diff
@@ -13,40 +14,52 @@ export function validateUseOfDeprecatedClass(
...
+    const deprecatedInfo = ui5Class.deprecatedInfo;
     const issues: UseOfDeprecatedClassIssue[] = [];
-    const commonIssueFields: UseOfDeprecatedClassIssue = {
-      kind: "UseOfDeprecatedClass",
-      message: `Deprecated UI5 Class: ${elementTagFqn} used.`,
-      severity: "warn",
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      range: undefined as any
+
+    const commonIssueFields = {
```
tal-sapan, line 29: I think you don't need this anymore since you only use it for`deprecatedIssue` which is already of the correct type (and has the ranges property)

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
```diff
@@ -13,40 +14,52 @@ export function validateUseOfDeprecatedClass(
     return [];
   }
 
-  if (ui5Class.deprecatedInfo !== undefined) {
+  if (
+    ui5Class.deprecatedInfo !== undefined &&
+    // An issue without a position is not a useful issue...
+    (xmlElement.syntax.openName !== undefined ||
+      /* istanbul ignore next - defensive programing for (currently) none reproducible scenario */
```
tal-sapan, line 21: Does this ignore comment work in the middle of the if condition?

packages/xml-views-validation/src/utils/deprecated-message-builder.ts
```diff
@@ -0,0 +1,23 @@
...
+  deprecatedInfo
+}: {
+  // TODO: add more kinds here as needed (e.g property/aggregation/...)
+  ui5Kind: "Class";
+  fqn: string;
+  deprecatedInfo: UI5DeprecatedInfo;
+}): string {
+  const msgPrefix = `UI5 ${ui5Kind}: ${fqn} is deprecated`;
+  const sinceOptionalPart = deprecatedInfo.since
+    ? ` since: ${deprecatedInfo.since}`
```
tal-sapan, line 16: In the documentation I use `since version <since>` without the colon (they don't write "version" in the `@since` tag)

packages/xml-views-validation/src/utils/deprecated-message-builder.ts
```diff
@@ -0,0 +1,23 @@
...
+  ui5Kind,
+  fqn,
+  deprecatedInfo
+}: {
+  // TODO: add more kinds here as needed (e.g property/aggregation/...)
+  ui5Kind: "Class";
+  fqn: string;
+  deprecatedInfo: UI5DeprecatedInfo;
+}): string {
+  const msgPrefix = `UI5 ${ui5Kind}: ${fqn} is deprecated`;
```
tal-sapan, line 14: I think the `:` is unnecessary

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
bd82: Yes I had coverage error on specific IF condition line and had to use this to fix it.

bd82:
> I prefer that you do the CR fixes according to the comments but it can be in another PR if you want

I am guessing you mean the new comments? I will have a look at those

packages/xml-views-validation/src/validators/attributes/unknown-enum-value.ts
bd82: good point, done:
- https://sapjira.wdf.sap.corp/browse/DEVXKINGSTON-937

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
bd82: Yep thanks.

packages/xml-views-validation/src/utils/deprecated-message-builder.ts
bd82: I based the string on the previous error message (with corrections).
- https://github.wdf.sap.corp/DevX/xmlvalidator/blob/master/src/fioriAnalysis/rules/XML_DEPRECATION.js#L79-L83

I don't have a strong opinion either way, lets let UX decide on this when we resolve: https://sapjira.wdf.sap.corp/browse/DEVXKINGSTON-937

packages/xml-views-validation/src/utils/deprecated-message-builder.ts
bd82: Will remove

bd82:
Second CR fixes applied in:
- https://github.com/SAP/ui5-language-assistant/pull/67/commits/b5b890c9c0c4b27b9334e582024e4b01d2b524c3

bd82:
Oops, forgot to update the tests.
This is the right commit: https://github.com/SAP/ui5-language-assistant/pull/67/commits/07998642776ef3900d40ad4b3d91e53f3bc5b125

packages/xml-views-validation/src/utils/deprecated-message-builder.ts
tal-sapan: It didn't have `:` and it said "version"... it was `The control <fqn> is deprecated[ since version <since>].[\n\t<description>]` (where `[]` parts are optional)

packages/xml-views-validation/src/validators/elements/use-of-deprecated-class.ts
```diff
@@ -26,18 +26,14 @@ export function validateUseOfDeprecatedClass(
     const deprecatedInfo = ui5Class.deprecatedInfo;
     const issues: UseOfDeprecatedClassIssue[] = [];
 
-    const commonIssueFields = {
+    const deprecatedIssue: UseOfDeprecatedClassIssue = {
```
tal-sapan, line 29: Looks ok, but now you don't need the `as` workarounds

bd82: Right, I've removed them

packages/xml-views-validation/src/utils/deprecated-message-builder.ts
bd82: I honestly have difficulty noticing these small differences, Lets leave this for proper PR made by UX folk to decide on the ultimate error string format.

tal-sapan: The diff is: instead of 
`` ` since: ${deprecatedInfo.since}` ``
Use 
`` ` since version ${deprecatedInfo.since}` ``

bd82: Ah o.k. I will see if I can change this before UX review.

PR #69, opened 2020-05-03
refactor(language-server): add utility for testing classes suggestions

tal-sapan:


tal-sapan:
Hide whitespace changes when reviewing to review faster :)

bd82:
Pros:
- Much more concise, each test is now only ~half the LOC.

Cons:
- Reaching failing test via error stack trace is slightly harder.
- Debugging tests is slightly harder (need to step into the assertion utility).

It is subjective but because there are so many tests **I am leaning towards the benefits being greater than the cons**. Which is not always the case sometimes duplication is o.k in tests... (IMHO)

PR #70, opened 2020-05-03
feat(language-server): add default value for property attribute

tal-sapan:
String, enum, boolean and number values are inserted and auto-selected.


packages/language-server/src/completion-items.ts
```diff
@@ -167,6 +166,26 @@ function createTextEdits(
...
+      // Auto-insert ="<default value>" (or ="" if there is no default value) for property attributes
+      if (suggestion.astNode.syntax.value === undefined) {
+        const defaultValue = suggestion.ui5Node.default;
+        // Only simple values are added in the completion (there are default values which are objects, arrays and null)
+        if (
+          defaultValue !== undefined &&
+          includes(["number", "boolean", "string"], typeof defaultValue)
+        ) {
+          // The ${0:} makes the text between : and } selected
+          newText += `="\${0:${defaultValue}}"`;
```
bd82, line 181: in the **future** we may need a user config option which kind of quotes to use

packages/language-server/src/completion-items.ts
```diff
@@ -167,6 +166,26 @@ function createTextEdits(
...
+
+      // Auto-insert ="<default value>" (or ="" if there is no default value) for property attributes
+      if (suggestion.astNode.syntax.value === undefined) {
+        const defaultValue = suggestion.ui5Node.default;
+        // Only simple values are added in the completion (there are default values which are objects, arrays and null)
+        if (
+          defaultValue !== undefined &&
+          includes(["number", "boolean", "string"], typeof defaultValue)
+        ) {
+          // The ${0:} makes the text between : and } selected
```
bd82, line 180: Can you please use the LSP term in the comment so it would be easy to look for in the specs?
e.g:
> // The LSP **PlaceHolder** `${0:}` makes the text between : and } selected

Maybe even focus on the **why** (functional scenario) more and let future developers reference the LSP spec 
for **how** exactly to use this syntax.
> // The LSP **PlaceHolder** `${0:}` selects text after completion for easy editing if the default value is not needed by the user.

packages/language-server/src/completion-items.ts
```diff
@@ -167,6 +166,26 @@ function createTextEdits(
...
+          newText += `="\${0:${defaultValue}}"`;
+        } else {
+          // The ${0} puts the cursor in the marked place
```
bd82, line 183: > // The **tabstop** `${0}` will place the cursor between the value's quote for easy typing of the (currently empty) value by the user.

?

packages/language-server/test/completion-items-spec.ts
```diff
@@ -18,6 +18,11 @@ describe("the UI5 language assistant Code Completion Services", () => {
   // Cursor position after selecting the suggestion
   const CURSOR_POSITION = "${0}";
 
+  // Pre-selected snippet text
+  function getSelectedText(text: string): string {
```
bd82, line 22: Perhaps a slightly different name would make this utility act more like a DSL
we use while building our assertions? e.g:

```javascript
{
   newText: `showNoData="${getSelectedText("true")}"`
   newText: `showNoData="${SELECTED_TEXT("true")}"`
}
```
I found the previous version with `CURSOR_POSITION` easier to read.

packages/language-server/test/completion-items-spec.ts
```diff
@@ -138,19 +152,28 @@ describe("the UI5 language assistant Code Completion Services", () => {
     const xmlSnippet = `<mvc:View 
                           xmlns:mvc="sap.ui.core.mvc" 
                           xmlns="sap.m"> 
-                          <List update⇶Start`;
+                          <List update⇶Start="onUpdateStart"`;
```
bd82, line 155: if we added a value should not the description of the test (scenario) also change (IT text)?
Is

packages/language-server/src/completion-items.ts
tal-sapan: Done

packages/language-server/src/completion-items.ts
tal-sapan: Done

packages/language-server/test/completion-items-spec.ts
tal-sapan: Done

packages/language-server/test/completion-items-spec.ts
tal-sapan: Done (I added it because this use case used to be covered by a similar test for a property attribute, which is now in a different `case` in the code)

PR #72, opened 2020-05-05
feat(language-server): expose xml-view-validations as LSP diagnostics

bd82:


packages/language-server/src/server.ts
```diff
@@ -63,6 +64,22 @@ connection.onCompletionResolve(
   }
 );
 
+documents.onDidChangeContent(async changeEvent => {
+  // TODO: can we ensure the promise has always been initialized first (never be undefined)?
```
bd82, line 68: @tal-sapan  could you look at the TODO comments with questions here?

packages/language-server/src/xml-view-diagnostics.ts
```diff
@@ -0,0 +1,108 @@
...
+        start: document.positionAt(currIssue.offsetRanges[0].start),
+        // Chevrotain's end offsets are none inclusive
+        end: document.positionAt(currIssue.offsetRanges[0].end + 1)
+      },
+      severity: toLspSeverity(currIssue.severity),
+      source: "UI5 Language Assistant",
+      message: currIssue.message
+    };
+
+    // TODO: I think we should ditch the related issues for now, it does not work well for the only use case I can think of currently (deprecated on ending tag)
```
bd82, line 49: What do you think on this comment @tal-sapan

bd82:
This is still a darft.

Tests may be in a separate PR as I would like to investigate suitability of snapshot testing for LSP flows.

packages/language-server/src/server.ts
tal-sapan: It is always initialized because `onInitialized` happens first but there is no way to let typescript know that

packages/language-server/src/xml-view-diagnostics.ts
tal-sapan: What do you mean doesn't work well? What does it do?

packages/language-server/src/server.ts
```diff
@@ -63,6 +64,22 @@ connection.onCompletionResolve(
...
+  if (getSemanticModelPromise === undefined) {
+    return;
+  }
+  // TODO: how to debounce this? should we debounce this?
```
tal-sapan, line 72: I don't understand what you mean with this comment

packages/language-server/src/server.ts
```diff
@@ -63,6 +64,22 @@ connection.onCompletionResolve(
...
+  const ui5Model = await getSemanticModelPromise;
+  // TODO: should we check we are dealing with a *.[view|fragment].xml?
```
tal-sapan, line 74: It's done in the client... we could add it to the server too, not sure if we should

packages/language-server/src/xml-view-diagnostics.ts
bd82: Related issues seem to just be implemented a link to other ranges.
so it does not replace actually having multiple distinct separate issue.

We may need it in unique ID validations, but as discussed earlier today, not in deprecated tags scenario

packages/language-server/src/server.ts
bd82: - How many such events happen per second? 
- Does the changeEvent only gets triggered after N miliseconds after there has been no input?
- Does the changeEvent happen for every character cicked?
- Do we need to optimize this or does not LSP implementation does it?

packages/language-server/src/server.ts
bd82: maybe it is good as defensive coding if someone else implements a client and forgets to do so?

packages/language-server/src/server.ts
bd82: o.k. I will remove the comment and we will leave the defensive if condition in the code

packages/language-server/src/xml-view-diagnostics.ts
bd82: I will remove the current manner of dealing with `related issues` (in an upcoming PR) and we will redo this capability in a better manner once we implement the unique ID validation.

packages/language-server/src/server.ts
bd82: I am going to assume the Node LSP library handles this, otherwise everyone would have major performance issues.

packages/language-server/src/server.ts
bd82: I will update the comment and leave it in for now so we consider it later.

bd82:
I am going to merge this soon. We will have to do additional reviews async.
Feel free to leave any other comments on this PR and I will address it.

Topics to be handled in other PRs:
- Testing (snapshots?)
- Remove `relatedIssues` API.
- Fixes for the existing two validations.

In this case/time I prefer clean history and momentum to synced reviews.

packages/language-server/src/xml-view-diagnostics.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+function validationIssuesToLspDiagnostics(
+  issues: UI5XMLViewIssue[],
+  document: TextDocument
+): Diagnostic[] {
+  const diagnostics: Diagnostic[] = map(issues, currIssue => {
+    const commonDiagnosticPros: Diagnostic = {
+      range: {
+        start: document.positionAt(currIssue.offsetRanges[0].start),
```
tal-sapan, line 41: Shouldn't there be an offset-by-1 issue with the start and end? Since Chevrotain offest is 1-based and VSCode is 0-based

bd82: offset is 0 based for both, only the line/columns are 1 based in Chevrotain

packages/language-server/src/server.ts
```diff
@@ -63,6 +64,22 @@ connection.onCompletionResolve(
   }
 );
 
+documents.onDidChangeContent(async changeEvent => {
```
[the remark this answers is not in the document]
bd82: > `onDidChangeTextDocument` was never invoked for me.

The basic MS LSP example used `documents.onDidChangeContent` API so I used it as well.

PR #83, opened 2020-05-07
test(semantic-model): model generation timeouts not working workaround

bd82:
temp workaround, root problem is due to cyclic dep between semantic-model(devDep) and
test-utils(dep)


bd82:
Merging without review as this blocks merging other PR:
- https://github.com/SAP/ui5-language-assistant/pull/73

@tal-sapan  I think we need to split up the test-utils somehow.

PR #88, opened 2020-05-09
test(language-server): snapshots tests for LSP diagnostics

bd82:


packages/language-server/scripts/update-diagnostics-snapshots.js
```diff
@@ -0,0 +1,49 @@
...
+const { format } = require("prettier");
+
+const {
+  getActualXMLWithRIssueRangesMarked,
+  getActualDiagnosticLSPResponse,
+  INPUT_FILE_NAME,
+  OUTPUT_RANGES_FILE_NAME,
+  OUTPUT_LSP_RESPONSE_FILE_NAME,
+} = require("../lib/test/snapshots/xml-view-diagnostics/utils");
+const { parse } = require("../");
```
bd82, line 15: TODO: REMOVE LINE 15

packages/language-server/src/xml-view-diagnostics.ts
```diff
@@ -75,8 +74,10 @@ function toLspSeverity(
       return DiagnosticSeverity.Error;
     case "warn":
       return DiagnosticSeverity.Warning;
+    /* istanbul ignore next - no `information` validation yet */
     case "info":
       return DiagnosticSeverity.Information;
+    /* istanbul ignore next - no `information` validation yet */
```
bd82, line 80: TODO: `hint` instead of `information`

packages/language-server/test/snapshots/xml-view-diagnostics/unknown-enum-value/diagnostic-spec.ts
```diff
@@ -0,0 +1,6 @@
+import { snapshotTestLSPDiagnostic } from "../utils";
+describe(`The Language Server Diagnostics feature`, () => {
```
bd82, line 2: TODO: `feature` -> `capabilities` ?

packages/language-server/test/snapshots/xml-view-diagnostics/use-of-deprecated-class/diagnostic-spec.ts
```diff
@@ -0,0 +1,6 @@
+import { snapshotTestLSPDiagnostic } from "../utils";
+describe(`The Language Server Diagnostics feature`, () => {
+  it("Can create diagnostic for an unknown enum value", async () => {
```
bd82, line 3: TODO: test description is wrong

packages/language-server/scripts/update-diagnostics-snapshots.js
bd82: 👍

packages/language-server/src/xml-view-diagnostics.ts
bd82: 👍

packages/language-server/test/snapshots/xml-view-diagnostics/utils.ts
```diff
@@ -0,0 +1,106 @@
+import { resolve, sep } from "path";
```
tal-sapan, line 1: Shouldn't this file be named `snapshot-utils` if it contains utilities for the snapshot tests? It's more specific and will prevent adding unrelated functions to it

packages/language-server/test/snapshots/xml-view-diagnostics/utils.ts
```diff
@@ -0,0 +1,106 @@
...
+import { readJsonSync, readFileSync } from "fs-extra";
+import { expect } from "chai";
+import { sortBy, forEachRight } from "lodash";
+import { generateModel } from "@ui5-language-assistant/test-utils";
+import { UI5SemanticModel } from "@ui5-language-assistant/semantic-model-types";
+import { getXMLViewDiagnostics } from "../../../src/xml-view-diagnostics";
+
+export const INPUT_FILE_NAME = "input.xml";
+export const OUTPUT_LSP_RESPONSE_FILE_NAME = "output-lsp-response.json";
+export const OUTPUT_RANGES_FILE_NAME = "output-ranges.xml";
```
tal-sapan, line 13: I though we could have to ranges inside the `input.xml`. It's not part of the input but it would prevent copying the input file, specifying the ranges in the response json file (instead only specifying the range number) and syncing it between the 2 output files.

packages/language-server/test/snapshots/xml-view-diagnostics/utils.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  const xmlTextDoc = TextDocument.create(
+    `file://${inputPath}`,
+    "xml",
+    0,
+    xmlSnippetContents
+  );
+
+  const actualDiagnostics = await getActualDiagnosticLSPResponse(testDir);
+  // this assumes no overlapping ranges
```
tal-sapan, line 66: This is fine for the current tests/implemented diagnostics but I'm not sure we can keep this assumption in the future.
It can be improved in another PR, but I thought we could keep a list of position and insert character (which would be one of the arrows) and sort that, and then do the same logic you did here (go over the list backwards) but add the insert character instead of both arrows.

packages/language-server/test/snapshots/xml-view-diagnostics/utils.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export function getExpectedXMLWithRIssueRangesMarked(testDir: string): string {
+  const sourcesTestDir = toSourcesTestDir(testDir);
+  const lspResponsePath = resolve(sourcesTestDir, OUTPUT_RANGES_FILE_NAME);
+  const expectedMarkedRanges = readFileSync(lspResponsePath).toString("utf-8");
+  return expectedMarkedRanges;
+}
+
+export async function snapshotTestLSPDiagnostic(
```
tal-sapan, line 95: Shouldn't this also check the ranges file? Currently you aren't using `getExpectedXMLWithRIssueRangesMarked` at all unless I missed something

packages/language-server/test/snapshots/xml-view-diagnostics/utils.ts
bd82: 👍

packages/language-server/test/snapshots/xml-view-diagnostics/utils.ts
bd82: I think I still want to print the whole LSP response,but combining the expected range markers in the input is a good idea, I will try it.

packages/language-server/test/snapshots/xml-view-diagnostics/utils.ts
bd82: Yeah, we can upgrade it **if** needed. I'm not yet sure if we need to, lets see after we implement more validations

packages/language-server/test/snapshots/xml-view-diagnostics/utils.ts
bd82: Oops 😄

packages/language-server/test/snapshots/xml-view-diagnostics/snapshots-utils.ts
```diff
@@ -1,16 +1,36 @@
...
+  expect(actualDiagnostics, "Mismatch LSP Diagnostics response").to.deep.equal(
+    expectedDiagnostics
+  );
+
+  const expectedRanges = map(expectedDiagnostics, (_) => _.range);
+  const expectedXMLWithRangeMarkers = getExpectedXMLWithRIssueRangesMarked(
+    testDir,
+    expectedRanges
+  );
+  const actualXMLWithRangeMarkers = readInputXMLSnippet(testDir, false);
```
tal-sapan, line 28: This is the input file content - it's more "expected" than "actual". This is used to check that the ranges in the input xml and the ranges specified in the expected response are synchronized.
If we don't compare them to the actual ranges at any point, why do we need them in the input xml at all? Or, a better question would be, what do we expect to do with the ranges marked in the XML?
What I wanted was to specify them only in the xml, so that we can reference them in the response without actually writing the line and character numbers, which will make changing the expected response unnecessary when we change the input xml without changing the actual issue (even if its position changes).
What it seems you want to do is use it for better error messages. For this we don't actually need the ranges in the input file since we add the actual and expected ranges according to the LSP response.

packages/language-server/test/snapshots/xml-view-diagnostics/snapshots-utils.ts
```diff
@@ -32,75 +52,80 @@ export async function getActualDiagnosticLSPResponse(
...
 
-export async function snapshotTestLSPDiagnostic(
-  testDir: string
-): Promise<void> {
-  const expected = getExpectedDiagnosticsLSPResponse(testDir);
-  const actual = await getActualDiagnosticLSPResponse(testDir);
-  expect(actual).to.deep.equal(expected);
+function readInputXMLSnippet(
+  testDir: string,
+  stringRangeMarkers = true
```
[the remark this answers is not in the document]
bd82: Yes thanks.

packages/language-server/test/snapshots/xml-view-diagnostics/snapshots-utils.ts
```diff
@@ -32,75 +52,80 @@ export async function getActualDiagnosticLSPResponse(
...
 }
 
-export async function getActualXMLWithRIssueRangesMarked(
-  testDir: string
-): Promise<string> {
-  const sourcesTestDir = toSourcesTestDir(testDir);
-  const inputPath = resolve(sourcesTestDir, INPUT_FILE_NAME);
-  const xmlSnippetContents = readFileSync(inputPath).toString("utf-8");
-
+export function getExpectedXMLWithRIssueRangesMarked(
```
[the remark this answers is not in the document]
bd82: Will change

packages/language-server/test/snapshots/xml-view-diagnostics/snapshots-utils.ts
tal-sapan: > what do we expect to do with the ranges marked in the XML?

This is be used for reading the test easily. Since the test only passes if the json and xml are synced, we know that the ranges in the json file must be the same as in the xml, so we can see visually where the diagnostics will appear in the editor.

bd82:
@tal-sapan please have a look at the last commit, I think it would be much clearer now with better names.

packages/language-server/test/snapshots/xml-view-diagnostics/snapshots-utils.ts
```diff
@@ -47,7 +53,7 @@ export function getExpectedDiagnosticsLSPResponse(
 const ui5ModelPromise = generateModel({ version: "1.71.14" });
 let ui5Model: UI5SemanticModel | undefined = undefined;
 
-export async function getActualDiagnosticLSPResponse(
+export async function computeNewDiagnosticLSPResponse(
```
tal-sapan, line 56: I preferred the previous name (with `actual`) since you actually call the API here

packages/language-server/test/snapshots/xml-view-diagnostics/snapshots-utils.ts
```diff
@@ -14,25 +14,31 @@ export const OUTPUT_LSP_RESPONSE_FILE_NAME = "output-lsp-response.json";
...
+  // Note the original `input.xml` snippet acts as **both**:
+  //   - The sample input.
+  //   - A snapshot for marker ranges.
+  const originalXMLSnippet = readInputXMLSnippet(testDir, false);
   expect(
-    actualXMLWithRangeMarkers,
+    originalXMLSnippet,
     "The XML input snippet range markers are incorrect"
-  ).to.equal(expectedXMLWithRangeMarkers);
+  ).to.equal(newlyComputedXMLWithMarkedRanges);
```
tal-sapan, line 38: You changed this from being the snapshot ranges to the actual ranges, which changes the purpose of this assertion - instead of making sure the input xml and expected response json are in sync, it checks that the ranges in the actual response are as expected in the input xml, I assume for better error messages. But if they aren't (and assuming the input xml and resopnse json are in sync) it will already fail in line 24...

bd82: I was trying to resolve the confusion we had yesterday due to mixing up the input and the snapshot.

So I am treating them as two **distinct** unrelated snapshots. Which are tested separately and visually inspected seperatly.

The assumption: `the input xml and response json are in sync` is not a safe assumption as the input.xml can be edited by **human** developers, it is not necessarily auto-generated.

packages/language-server/test/snapshots/xml-view-diagnostics/snapshots-utils.ts
bd82: I too normally prefer the names with `actual|expected` prefixes in test cases.

However I felt there was a bit of confusion here, (perhaps due to the input **also** being used as a snapshot?). So I selected terms closer to the snapshots testing methodology concepts (e.g: `newlyComputed` vs `snapshot`) and tried using those terms **consistently** everywhere in this file.

packages/language-server/test/snapshots/xml-view-diagnostics/snapshots-utils.ts
tal-sapan: Could you add a check for the consistency of the ranges in the input xml and snapshot json? Preferably at the beginning of the function. It would make it easier to understand if a test fails due to a bug in the code or because there's a problem in the xml/snapshot json.
It would also make it easier to read the tests if you assume the 2 files are in sync. No need to figure out the ranges in the json.

packages/language-server/test/snapshots/xml-view-diagnostics/snapshots-utils.ts
```diff
@@ -0,0 +1,137 @@
...
+  const diagnosticsSnapshots = readSnapshotDiagnosticsLSPResponse(testDir);
+  const newlyComputedDiagnostics = await computeNewDiagnosticLSPResponse(
+    testDir
+  );
+  expect(
+    newlyComputedDiagnostics,
+    "Snapshot Mismatch in LSP Diagnostics response"
+  ).to.deep.equal(diagnosticsSnapshots);
+
+  const newlyCommutedRanges = map(newlyComputedDiagnostics, (_) => _.range);
```
tal-sapan, line 26: Rename to `newlyComputedRanges`

packages/language-server/test/snapshots/xml-view-diagnostics/snapshots-utils.ts
tal-sapan: Ok

packages/language-server/test/snapshots/xml-view-diagnostics/snapshots-utils.ts
bd82: I don't understand the benefits, lets sync in person on this (maybe tomorrow?)

tal-sapan: ok

PR #99, opened 2020-05-10
fix(language-server): suggest current attribute name

tal-sapan:


package.json
```diff
@@ -43,14 +43,14 @@
     "eslint-plugin-eslint-comments": "3.1.2",
     "husky": "4.2.5",
     "lerna": "3.20.2",
-    "lint-staged": "10.0.3",
+    "lint-staged": "10.2.2",
```
bd82, line 46: this should not be part of your change.

package.json
```diff
@@ -43,14 +43,14 @@
...
     "mocha": "7.1.2",
     "npm-run-all": "4.1.5",
     "nyc": "15.0.1",
     "prettier": "2.0.5",
     "source-map-support": "0.5.19",
     "typescript": "3.8.3",
-    "cz-conventional-changelog": "3.1.0",
+    "cz-conventional-changelog": "3.2.0",
```
bd82, line 53: i will udpate the branch

packages/xml-views-completion/test/providers/attributeName/prop-event-assoc-spec.ts
```diff
@@ -120,6 +120,62 @@ describe("The ui5-language-assistant xml-views-completion", () => {
         });
       });
 
+      it("will not filter out the current attribute when only attribute name exists", () => {
```
bd82, line 123: This is an **imperative** description of how it does it.
But the **functional** description is more like:
`will suggest attribute which is identical to the user's input prefix`

bd82: also I did not quite understand `when only attribute name exists`

tal-sapan: Other tests in this file also mention "filtering out", so I went with a similar description. My tests mention "not filtering out" so maybe it's different... I'll change it

bd82: I think other tests may mention filter in functional spec manner, e.g: `filter out none public classes` or `filter by prefix` but maybe some other descriptions could also be upgraded.

PR #100, opened 2020-05-10
fix(language-server): namespace added at wrong position for closed tag

tal-sapan:
For this xml:
`<>`
The code completion for tag name added the namespace at the wrong position.


packages/language-server/src/completion-items.ts
```diff
@@ -371,13 +371,34 @@ function rangeContains(range: Range, inner: Range): boolean {
...
+/**
+ * Create an insert range at the position right before the sent column.
+ * For exmaple, for text "12345" the character in position 1 is "1" and the
+ * insert position before it is the beginning of the text -> "|12345".
+ * The character in position 2 is "2" and the insert position before it is "1|2345".
+ * The character at position 5 is "5" and the insert position before it is "1234|5".
+ * */
+function createInsertRangeBefore(line: number, column: number): Range {
+  return {
+    start: Position.create(line - 1, column - 1),
```
bd82, line 397: comment that Chevrotain line/columns are `1 based` while VSCode line/columns are 0 based?

packages/language-server/src/completion-items.ts
```diff
@@ -371,13 +371,34 @@ function rangeContains(range: Range, inner: Range): boolean {
...
+/**
+ * Create an insert range at the position right after the sent column.
+ * For exmaple, for text "12345" the character in position 1 is "1" and the
+ * insert position after it is "1|2345".
+ * The character in position 2 is "2" and the insert position after it is "12|345".
+ * The character at position 5 is "5" and the insert position after it at the end of the text -> "12345|".
+ * */
+function createInsertRangeAfter(line: number, column: number): Range {
   return {
     start: Position.create(line - 1, column),
```
bd82, line 383: same as below, at minimum some comment about the magical `-1`

packages/language-server/src/completion-items.ts
```diff
@@ -454,10 +475,24 @@ function getAddNamespaceEdit(
...
-    // We want to insert, not replace - the actual position should be at the end of the range.
-    const range = createInsertRange(position.endLine, position.endColumn);
+    let range: Range;
+    if (parent.rootElement.syntax.openName !== undefined) {
+      const position = parent.rootElement.syntax.openName;
+      // We want to insert the namespace at the end of the range, after the tag name.
+      range = createInsertRangeAfter(position.endLine, position.endColumn);
+    } else {
+      const position =
+        parent.rootElement.syntax.openBody ?? parent.rootElement.position;
```
bd82, line 485: TODO: consider change XML-Tools/AST and adding the "<" token to the syntax.
This will reduce a branch in the code, and avoid the need for 4 lines comment.

Perhaps add it as a TODO / do it in another PR (later) due to time constraints ...

packages/language-server/test/completion-items-classes-spec.ts
```diff
@@ -332,6 +333,50 @@ describe("the UI5 language assistant Code Completion Services - classes", () =>
...
+      const tagName = getTagName(suggestion.textEdit);
+      expectExists(tagName, "tag name in suggestion");
+      const ns = tagName.split(":")[0];
+      const attributes = getAttributes(suggestion.textEdit);
+      expect(
+        attributes,
+        `attributes of ${suggestion.textEdit?.newText}`
+      ).to.have.lengthOf(2); // Namespace and tab stop
+      expect(
+        attributes[0],
```
bd82, line 347: do we need to assert on `attributes[1]` too?

packages/language-server/test/completion-items-classes-spec.ts
```diff
@@ -332,6 +333,50 @@ describe("the UI5 language assistant Code Completion Services - classes", () =>
...
+        `attributes[0] of ${suggestion.textEdit?.newText}`
+      ).to.match(
+        new RegExp(
+          `^xmlns:${ns}=`
+        ) /*, `attribute should start with xmlns:${ns}=`*/
+      );
+    });
+  });
+
+  it("will return valid class suggestions for empty tag with closing bracket", () => {
```
bd82, line 357: for future upgrades: snapshots testing may make writing these tests much simpler.

packages/language-server/src/completion-items.ts
tal-sapan: It's a bit more complicated than that... Chevrotain position is a position **of** a character while vscode position is a position **between** characters. That's why there's a need for `createInsertRangeBefore` and `createInsertRangeAfter`.

packages/language-server/src/completion-items.ts
tal-sapan: I'll add a comment

packages/language-server/test/completion-items-classes-spec.ts
tal-sapan: We can, not sure if it's necessary since we have some tests that already assert on it

packages/language-server/src/completion-items.ts
bd82: even a better reason to comment this or maybe even extract some utility/function with documentation.

packages/language-server/test/completion-items-classes-spec.ts
tal-sapan: These tests look at all the returned results, not a specific case. I'm not sure I would want a test that lists all possible classes in the root.

packages/language-server/test/completion-items-classes-spec.ts
bd82: lets leave it as is, we will get assertions on both attributes for **free* if/when we implement snapshots for completions as well.

PR #102, opened 2020-05-10
test(xml-views-validation): compute expected validation ranges

bd82:


packages/xml-views-validation/test/test-utils.ts
```diff
@@ -30,3 +31,28 @@ export function testValidationsScenario(opts: {
   });
   opts.assertion(issues);
 }
+
+export function computeExpectedRanges(markedXMLSnippet: string): OffsetRange[] {
```
bd82, line 35: should this be in test utils package?

packages/xml-views-validation/test/test-utils.ts
```diff
@@ -30,3 +31,28 @@ export function testValidationsScenario(opts: {
...
+  const expectedRanges = [];
+  const markedPattern = /🢂(.+?)🢀/gu;
+  let match;
+  // eslint-disable-next-line no-cond-assign
+  while ((match = markedPattern.exec(markedXMLSnippet))) {
```
tal-sapan, line 40: Compare to `null`?
```suggestion
  while ((match = markedPattern.exec(markedXMLSnippet)) !== null) {
```

packages/xml-views-validation/test/test-utils.ts
```diff
@@ -30,3 +31,28 @@ export function testValidationsScenario(opts: {
...
+    const start = match.index;
+    // Chevrotain ranges are inclusive (start) to exclusive (end)
+    const end = start + match[1].length - 1;
+    expectedRanges.push({ start, end });
```
tal-sapan, line 44: I think you have to change the string you match against after the match. It won't work if there is more than 1 range in the string because of the offset caused by the arrows (which aren't in the arrow-less xml).

packages/xml-views-validation/test/test-utils.ts
```diff
@@ -30,3 +31,28 @@ export function testValidationsScenario(opts: {
...
   opts.assertion(issues);
 }
+
+export function computeExpectedRanges(markedXMLSnippet: string): OffsetRange[] {
+  const expectedRanges = [];
+  const markedPattern = /🢂(.+?)🢀/gu;
+  let match;
+  // eslint-disable-next-line no-cond-assign
+  while ((match = markedPattern.exec(markedXMLSnippet))) {
+    const start = match.index;
```
tal-sapan, line 41: Doesn't the range-opening arrow cause an offset-by-1 issue here? For a string like `🢂a🢀` the `start` and `end` will both be in index 1 although they should be at `{0, 0}` because the actual string is `a`...

bd82: > /🢂(.+?)🢀/gu.exec("🢂a🢀").index // -> 0

index is at the start of the match, which is before the "🢂"

packages/xml-views-validation/test/test-utils.ts
bd82: Good point

packages/xml-views-validation/test/test-utils.ts
bd82: fixed

packages/xml-views-validation/test/test-utils.ts
bd82: fixed

packages/xml-views-validation/test/test-utils.ts
tal-sapan: Ah, ok. I thought it was the group index.

packages/xml-views-validation/test/test-utils.ts
```diff
@@ -34,11 +34,18 @@ export function testValidationsScenario(opts: {
 
 export function computeExpectedRanges(markedXMLSnippet: string): OffsetRange[] {
   const expectedRanges = [];
+
+  function previousMarkersOffset(): number {
+    // "🢂".length === 2 (due to surrogate pairs in UTF-8)
+    // Two markers per previous range.
+    return expectedRanges.length * 2 * 2;
```
tal-sapan, line 41: Instead of magic numbers 2 and 2 you can use
```suggestion
    return expectedRanges.length * ("🢂".length + "🢀".length);
```

bd82: will extract the marker signs to variables as well, it reduces readability (no more literal regExps), but makes it easy to update the marker arrows in the future.

tal-sapan: I'm not sure that's necessary, they will remain literal in tests

tal-sapan: And it's easy enough to find-and-replace these characters

PR #103, opened 2020-05-11
feat: add validation for unknown namespace name in xmlns attribute value

tal-sapan:
~~This doesn't use the expected ranges function in unit tests yet~~


tal-sapan:
Also pending Yuval's answer regarding the severity (currently I set it to warning because it doesn't cause an error in runtime by itself)

packages/language-server/src/xml-view-diagnostics.ts
```diff
@@ -48,18 +48,16 @@ function validationIssuesToLspDiagnostics(
...
-        };
       case "UseOfDeprecatedClass":
         return {
           ...commonDiagnosticPros,
           tags: [DiagnosticTag.Deprecated],
         };
-      /* istanbul ignore next - defensive programming */
       default:
-        assertNever(issueKind);
+        // Most issues don't need any special handling
```
bd82, line 57: we lose the exhaustiveness checking, is it not better to explicitly handle every state
and have the compiler fail if we forget to add/handle a special case?
then silently run with partial functionality?

- https://www.typescriptlang.org/docs/handbook/advanced-types.html#exhaustiveness-checking
- https://www.geeksforgeeks.org/scala-sealed-trait/

packages/language-server/test/snapshots/xml-view-diagnostics/unknown-xmlns-namespace/diagnostic-spec.ts
```diff
@@ -0,0 +1,6 @@
+import { snapshotTestLSPDiagnostic } from "../snapshots-utils";
+describe(`The language server diagnostics capability`, () => {
```
bd82, line 2: newline?

packages/logic-utils/src/utils/xml-ns-key.ts
```diff
@@ -1,6 +1,5 @@
-//we are only allowing word (\w+) characters in prefixes now (in completions)
-//TODO it should be aligned with the full XML spec
-const namespaceRegex = /^xmlns(:(?<prefix>\w*))?$/;
+// The prefix is an xml name minus the ":"
+const namespaceRegex = /^xmlns(:(?<prefix>[^:=]*))?$/;
```
bd82, line 2: by alignment I meant exporting the relevant `NAME` pattern from the XML Tokenizer.
- https://github.com/SAP/xml-tools/blob/master/packages/parser/lib/lexer.js#L35-L44

Your change is probably better so leave it, although instead of a subset it is now a superset so could even be incorrect in some cases.

But please keep the `TODO:` comment and maybe add the link:
> https://github.com/SAP/xml-tools/blob/master/packages/parser/lib/lexer.js#L35-L44

packages/logic-utils/api.d.ts
```diff
@@ -83,3 +83,15 @@ export function getRootSymbolParent(node: BaseUI5Node): BaseUI5Node | undefined;
  * Return a human-readable string representation of a UI5 type
  */
 export function typeToString(type: UI5Type | undefined): string;
+
+/**
+ * Check if an xml attribute key is an xmlns key
+ * @param key
+ */
+export function isXMLNamespaceKey(key: string): boolean;
```
bd82, line 91: Maybe this belong in XML-Tools?
We already do it there (incorrectly?)
- https://github.com/SAP/xml-tools/blob/master/packages/ast/lib/build-ast.js#L259

It is very syntax specific.
- https://www.w3.org/TR/xml-names/#NT-PrefixedAttName

packages/logic-utils/api.d.ts
```diff
@@ -83,3 +83,15 @@ export function getRootSymbolParent(node: BaseUI5Node): BaseUI5Node | undefined;
...
+
+/**
+ * Get the attribute name, without its "xmlns:" prefix, from an xmlns attribute key
+ * @param key
+ */
+export function getXMLNamespaceKeyPrefix(key: string): string;
```
bd82, line 97: same as above

packages/logic-utils/test/utils/xml-ns-key-spec.ts
```diff
@@ -0,0 +1,51 @@
+import { expect } from "chai";
```
bd82, line 1: its one thing to copy a one line regexp that should be elsewhere, but when we
also implement tests for that, does it become a larger tech debt?

packages/semantic-model/api.d.ts
```diff
@@ -32,3 +32,13 @@ export function forEachSymbol(
   model: UI5SemanticModel,
   iteratee: (symbol: BaseUI5Node, fqn: string) => boolean | void
 ): void;
+
+/**
+ * Return a root symbol according to its fully qualified name, or undefined if not found
+ * @param model
+ * @param fqn
+ */
+export function findSymbol(
```
bd82, line 41: How did we never need it before?

bd82: I mean should we refactor so this utility is used in previous use cases as well?

packages/xml-views-validation/api.d.ts
```diff
@@ -20,7 +20,10 @@ export interface OffsetRange {
   end: number;
 }
 
-export type UI5XMLViewIssue = UnknownEnumValueIssue | UseOfDeprecatedClassIssue;
+export type UI5XMLViewIssue =
+  | UnknownEnumValueIssue
+  | UseOfDeprecatedClassIssue
+  | UnknownNamespaceInXmlnsAttributeValueIssue;
```
bd82, line 26: Subjective / Discussion: 

Should this be named `UnknownNamespaceDefinitionIssue`? or something similar without the `AttribueValue` suffix?

Also should we remove the `Issue`  suffix in all these names?

packages/xml-views-validation/src/validators/attributes/unknown-xmlns-namespace.ts
```diff
@@ -0,0 +1,46 @@
+import { XMLAttribute } from "@xml-tools/ast";
+import { UI5SemanticModel } from "@ui5-language-assistant/semantic-model-types";
+import { findSymbol } from "@ui5-language-assistant/semantic-model";
+import { UnknownNamespaceInXmlnsAttributeValueIssue } from "../../../api";
```
bd82, line 4: import order (can you try integrating the eslint rule? I did not succde)

packages/language-server/src/xml-view-diagnostics.ts
tal-sapan: I don't think so in this case. In the completion items it makes sense because a lot of the parameters are LSP-specific and common, but in the validations we get all the information (except very specific things like the tags) from the inner logic. If we have more specific things (like if we decide to change the API to return the xml node instead of ranges) we can change it. But I don't see a need for it currently.

packages/language-server/test/snapshots/xml-view-diagnostics/unknown-xmlns-namespace/diagnostic-spec.ts
tal-sapan: I ran prettier... it doesn't have a problem with the line length

packages/logic-utils/src/utils/xml-ns-key.ts
tal-sapan: I figured that we can rely on a lower level package to validate the attribute key and we can do a "best guess" that might contain invalid keys as long as it contains all the valid keys... so there it is.
Since the attribute key is already checked I don't think we have to re-validate it here.
I will add a comment that invalid attribute names are handled by the parser. We only validate the attribute value here and we can assume that if the user started the attribute key with `xmlns:` they meant it to be an xmlns attribute (even if it contains invalid characters).

packages/logic-utils/api.d.ts
tal-sapan: It's not incorrect really, the spec says it's an attribute name that starts with `xmlns:` and doesn't have `:` after. The lower level validation of which characters can appear in the attribute name is handled by the parser in the link you put above.
This utility might belong there, yes. I don't know where to put it though. Maybe we can discuss it tomorrow?

packages/logic-utils/test/utils/xml-ns-key-spec.ts
tal-sapan: Wherever these functions are implemented they need tests...

packages/semantic-model/api.d.ts
tal-sapan: I think we looked for specific things before (class or namespace or enum etc) and not "any symbol with this name". Except while building the semantic model (which did use it but internally in the package it's defined).

packages/xml-views-validation/api.d.ts
tal-sapan: If we have other issues about unknown namespace definition (which aren't in the `xmlns:` attribute value) I prefer to have a different type for them.
About the `Issue` suffix I don't really care but it could help to understand what this is with the suffix since we don't use inheritance here.

packages/xml-views-validation/src/validators/attributes/unknown-xmlns-namespace.ts
tal-sapan: I want to try but I don't think I'll have time this sprint

packages/language-server/src/xml-view-diagnostics.ts
bd82: Lets sync on this tomorrow

packages/language-server/test/snapshots/xml-view-diagnostics/unknown-xmlns-namespace/diagnostic-spec.ts
bd82: I meant between import blocks and rest of code?

packages/logic-utils/src/utils/xml-ns-key.ts
bd82: o.k. we can assume the prefix is a valid XML Identifier.
If it was not it would not be usable anyhow in a tag name.

packages/language-server/test/snapshots/xml-view-diagnostics/unknown-xmlns-namespace/diagnostic-spec.ts
tal-sapan: It's the same as in the other snapshot tests. I think it's ok because it's basically a placeholder tepmlate... maybe we can even remove these files and generate the tests on the fly according to the folder names?

packages/logic-utils/api.d.ts
bd82: o.k lets sync on this.

packages/language-server/test/snapshots/xml-view-diagnostics/unknown-xmlns-namespace/diagnostic-spec.ts
tal-sapan: I'll try to generate them and see if it looks better

packages/logic-utils/test/utils/xml-ns-key-spec.ts
bd82: Yes the tests are needed, the point was that before we had a little duplicate regexp/logic now we have a full set of tests possibly in wrong place.

packages/xml-views-validation/api.d.ts
bd82: I see this difference / inconsistency:
- `InXmlnsAttributeValue` is described in XML **syntax** information
- `UnknownEnumValue` and `UseOfDeprecatedClassIssue` are described in UI5 VIew semantics.

packages/xml-views-validation/src/validators/attributes/unknown-xmlns-namespace.ts
bd82: Maybe I can find time to try again during some useless meeting 😄

packages/language-server/test/snapshots/xml-view-diagnostics/unknown-xmlns-namespace/diagnostic-spec.ts
tal-sapan: See #104

packages/xml-views-validation/api.d.ts
tal-sapan: I think if we allow the user to set the severity or turn off validations, it's possible that they won't want this validation in the xmlns attribute but will want it in other places (I don't know where... but the xmlns attribute it doesn't affect the runtime until you try to use it in a tag/attribute name).

packages/language-server/src/xml-view-diagnostics.ts
tal-sapan: Done

PR #104, opened 2020-05-11
refactor: auto-generate snapshot test cases from folders

tal-sapan:


packages/language-server/test/snapshots/xml-view-diagnostics/snapshots-diagnostics-spec.ts
```diff
@@ -0,0 +1,17 @@
+import * as klawSync from "klaw-sync";
+import { map, forEach } from "lodash";
+import { basename } from "path";
+import { snapshotTestLSPDiagnostic } from "./snapshots-utils";
+
+describe(`The language server diagnostics capability`, () => {
```
bd82, line 6: I intentionally created the spec files to allow `.only` and run only a single test.
How would we support this scenario if its all automatic?

tal-sapan: It can be done with code, besides adding `.only` also filter the returned folder after line 12:
```javascript
if (dirName !== "my-test-name") {
  return;
}
```
We can put it in a comment if you like to make it easier to use in debugging/local dev scenarios.
I think it's worth it instead of copy-pasting the same file for each snapshot test and only changing the description.
Even if we keep the copy-pasted file I think the description should still be created automatically from the folder name, because it's very easy to forget to change the description and it's not noticeable in code review - so we'll get many tests with copy-pasted names.

bd82: I wanted more detailed descriptions, but I agree that its easy to forget to update while copy/pasting.

In XML-Tools I used the folder name to auto generate the description which avoids this problem.
- https://github.com/SAP/xml-tools/blob/master/packages/ast/test/snapshots/invalid/attributes-no-value-last/sample.spec.js

Lets try your suggested alternative.
Please inspect the scenario when a test fails: 
- Is the trace stack usable? 
- Does the error easily lead back to the correct file?
  - e.g maybe we should not remove the hyphens the auto-generated description?
  - Maybe the `snapshotTestLSPDiagnostic` utility should include the folder names in its error messages?

tal-sapan: The stack trace doesn't contain much info, but I added the relevant file path(s) to the error messages and the real folder name to the test name. For example:
```
  1) The language server diagnostics capability
       Can create diagnostic for unknown enum value (unknown-enum-value):
      AssertionError: Snapshot Mismatch in LSP Diagnostics response. 
     Snapshot is at: packages\language-server\test\snapshots\xml-view-diagnostics\unknown-enum-value\output-lsp-response.json
```
The file path is clickable in the terminal opened in vscode.

PR #105, opened 2020-05-11
refactor(xml-views-validation): remove description in deprecation issue

bd82:
The description is often links which cannot be redered as there is no support for markdown, or
complex text which would become unreadable spam, again due to no "smart" rendering.

### DIFF
--- a/packages/language-server/src/xml-view-diagnostics.ts
+++ b/packages/language-server/src/xml-view-diagnostics.ts
@@ -14,6 +14,7 @@ import {
   validateXMLView,
   XMLViewIssueSeverity,
 } from "@ui5-language-assistant/xml-views-validation";
+import { convertDescriptionToMarkup } from "./documentation";
 
 export function getXMLViewDiagnostics(opts: {
   document: TextDocument;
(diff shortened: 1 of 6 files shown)


packages/xml-views-validation/src/utils/deprecated-message-builder.ts
```diff
@@ -13,11 +13,10 @@ export function buildDeprecatedIssueMessage({
 }): string {
   const msgPrefix = `UI5 ${ui5Kind} ${fqn} is deprecated`;
   const sinceOptionalPart = deprecatedInfo.since
-    ? ` since: ${deprecatedInfo.since}`
-    : "";
-  const detailsOptionalPart = deprecatedInfo.text
-    ? `\n\t${deprecatedInfo.text}.`
+    ? ` since version: ${deprecatedInfo.since}`
```
tal-sapan, line 16: Remove the `:`
```suggestion
    ? ` since version ${deprecatedInfo.since}`
```

packages/xml-views-validation/src/utils/deprecated-message-builder.ts
```diff
@@ -13,11 +13,10 @@ export function buildDeprecatedIssueMessage({
...
     : "";
 
-  return `${msgPrefix}${sinceOptionalPart}.${detailsOptionalPart}`;
+  // we are ignoring the deprecated description because there is no way to render
```
tal-sapan, line 19: Please return the description, it has value even without the markdown

packages/language-server/src/xml-view-diagnostics.ts
```diff
@@ -26,24 +27,30 @@ export function getXMLViewDiagnostics(opts: {
...
-        start: document.positionAt(currIssue.offsetRange.start),
+        start: opts.document.positionAt(currIssue.offsetRange.start),
         // Chevrotain's end offsets are none inclusive
-        end: document.positionAt(currIssue.offsetRange.end + 1),
+        end: opts.document.positionAt(currIssue.offsetRange.end + 1),
       },
       severity: toLspSeverity(currIssue.severity),
       source: "UI5 Language Assistant",
-      message: currIssue.message,
+      message: convertDescriptionToMarkup(currIssue.message, opts.ui5Model)
```
tal-sapan, line 52: If the markdown isn't displayed in the problems view, why convert it to markdown? Wouldn't it be better to remove the formatting (which would require a new function)?

bd82: this is a left over mistake.

Converting is also a problem when the message is huge.

packages/xml-views-validation/src/utils/deprecated-message-builder.ts
bd82: It does not always have value, Lets consult with UI5/Product, examples:
- image

- image

- image

bd82:
Will apply Frank's suggestions in another PR

PR #113, opened 2020-05-12
feat: validate boolean value in property

tal-sapan:
Small refactor: add util functions to get ui5 class and property from the xml element/attribute in logic-utils.


bd82:
Did you amend/push? looks like we lost all the previous discussion, or am I confusing this with another PR?

packages/logic-utils/api.d.ts
```diff
@@ -83,3 +83,23 @@ export function getRootSymbolParent(node: BaseUI5Node): BaseUI5Node | undefined;
  * Return a human-readable string representation of a UI5 type
  */
 export function typeToString(type: UI5Type | undefined): string;
+
+/**
+ * Return the UI5 Class for an XML Element
+ * @param element
+ * @param model
+ */
+export function getClassByElement(
```
bd82, line 92: question: is the verbose `getUI5ClassByXMLElement` name worth it for accuracy?

packages/logic-utils/api.d.ts
```diff
@@ -83,3 +83,23 @@ export function getRootSymbolParent(node: BaseUI5Node): BaseUI5Node | undefined;
...
+  element: XMLElement,
+  model: UI5SemanticModel
+): UI5Class | undefined;
+
+/**
+ * Return the UI5 Property for an XML Attribute
+ * @param attribute
+ * @param model
+ */
+export function getPropertyByAttributeKey(
```
bd82, line 102: same question as above

packages/xml-views-completion/src/providers/attributeValue/boolean-literal.ts
```diff
@@ -30,15 +27,7 @@ const allBooleanValues: BooleanValue[] = [
...
-  const xmlElement = opts.element;
-  const xmlAttribute = opts.attribute;
-
-  const elementClass = getClassByElement(xmlElement, opts.context);
-  if (elementClass === undefined) {
-    return [];
-  }
-  const properties = flattenProperties(elementClass);
-  const ui5Property = find(properties, ["name", xmlAttribute.key]);
+  const ui5Property = getPropertyByAttributeKey(opts.attribute, opts.context);
```
bd82, line 30: 👍

packages/xml-views-validation/src/utils/is-binding-attribute-value.ts
```diff
@@ -0,0 +1,7 @@
+export function isPossibleBindingAttributeValue(value: string): boolean {
+  // Check if an attribute value might be a binding expression.
+  // The current check is very naive - it only looks for matching curly brackets anywhere in the value.
```
bd82, line 3: maybe nitpicking on description but this regExp will find the longest substring surrounded by curly brackets (its a greedy match) not **matching** (you mean balanced?) brackets.

packages/xml-views-validation/src/validators/attributes/invalid-boolean-value.ts
```diff
@@ -0,0 +1,52 @@
...
+    actualAttributeValue === null ||
+    actualAttributeValueToken === undefined ||
+    isPossibleBindingAttributeValue(actualAttributeValue)
+  ) {
+    return [];
+  }
+
+  const ui5Property = getPropertyByAttributeKey(attribute, model);
+  const propType = ui5Property?.type;
+  if (propType?.kind !== "PrimitiveType" || propType.name !== "Boolean") {
```
bd82, line 24: are there typos in the SDK? e.g "Boolean" vs "boolean"?
should we ignore case? or should the SDK be fixed?

packages/xml-views-validation/src/validators/attributes/invalid-boolean-value.ts
```diff
@@ -0,0 +1,52 @@
...
+    find(possibleBooleanValues, (_) => _ === actualAttributeValue) === undefined
+  ) {
+    const possibleValuesWithQuotes = map(
+      possibleBooleanValues,
+      (_) => `"${_}"`
+    );
+    return [
+      {
+        kind: "InvalidBooleanValue",
+        message: `Invalid value: ${
```
bd82, line 39: should the message include the word "boolean"?
Would it make it easier to understand?

packages/xml-views-validation/src/validators/attributes/invalid-boolean-value.ts
```diff
@@ -0,0 +1,52 @@
...
+          actualAttributeValueToken.image
+        }, expecting one of: [${possibleValuesWithQuotes.join(", ")}].`,
```
bd82, line 41: one of [x,y,z,t] is a generic format from enum.
In boolean should we just say `expecting "true" or "false"`?

packages/xml-views-validation/src/validators/attributes/unknown-enum-value.ts
```diff
@@ -1,40 +1,30 @@
...
 import { UnknownEnumValueIssue } from "../../../api";
+import { isPossibleBindingAttributeValue } from "../../utils/is-binding-attribute-value";
 
-// TODO: Some logic here is very similar to: `enumSuggestions` in xml-view-completions,
-//  evaluate extracting the common parts
 export function validateUnknownEnumValue(
   attribute: XMLAttribute,
   model: UI5SemanticModel
 ): UnknownEnumValueIssue[] {
-  const elementTagFqn = xmlToFQN(attribute.parent);
```
bd82, line 16: 👍

packages/xml-views-validation/src/validators/attributes/unknown-enum-value.ts
```diff
@@ -1,40 +1,30 @@
...
-    return [];
-  }
-
   const actualAttributeValue = attribute.value;
   const actualAttributeValueToken = attribute.syntax.value;
   if (
     actualAttributeValue === null ||
-    actualAttributeValueToken === undefined
+    actualAttributeValueToken === undefined ||
+    isPossibleBindingAttributeValue(actualAttributeValue)
```
bd82, line 17: 👍

packages/xml-views-validation/src/utils/is-binding-attribute-value.ts
tal-sapan: I think we shouldn't check the syntax of the binding expression (like if the brackets are balanced) and consider anything with `{` and `}` (in this order) as a possible binding expression that shouldn't be validated. I don't do anything with the result expect check if it matches so it doesn't matter if it's greedy.

packages/xml-views-validation/src/validators/attributes/invalid-boolean-value.ts
tal-sapan: There are but it's handled in the semantic model, the strings used here (`"PrimitiveType"` and `"Boolean"`) are well-defined

tal-sapan: Although I'm thinking about making the primitive type names lowercase because that's generally how they are written in the api.json and the sdk

packages/xml-views-validation/src/validators/attributes/invalid-boolean-value.ts
tal-sapan: Regarding this and the "one of" comment, we should consult with UA on the message

packages/logic-utils/api.d.ts
tal-sapan: Changed

packages/logic-utils/api.d.ts
tal-sapan: Changed

packages/xml-views-validation/src/validators/attributes/invalid-boolean-value.ts
tal-sapan: I will change according to your suggestion and we'll consult UA later

packages/xml-views-validation/src/utils/is-binding-attribute-value.ts
bd82: I'm fine with the logic, just meant that the description in the comment is more like look for a "pair" of curly brackets than "matching" (assuming matching implies balanced). thats why I said this comment may be nitpicking :)

tal-sapan: Ah, ok. I didn't realize you were talking about the comment.

PR #119, opened 2020-05-13
feat: tooltip on hover

sapirpol:


packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+  UI5Namespace,
+  UI5EnumValue,
+} from "@ui5-language-assistant/semantic-model-types";
+import { getNodeDocumentation } from "./documentation";
+
+interface XMLElementHover {
+  kind: "XMLElementHover";
+  astNode: XMLElement;
+}
+interface XMLKeyAttributeHover {
```
bd82, line 42: > XMLKeyAttribute -> XMLAttributeKey ?

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+  kind: "XMLKeyAttributeHover";
+  astNode: XMLAttribute;
+}
+interface XMLValueAttributeHover {
```
bd82, line 46: as comment above

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+  astNode: XMLAttribute;
+}
+export type XMLHoverContext =
+  | XMLElementHover
+  | XMLKeyAttributeHover
+  | XMLValueAttributeHover;
+
+class HoverContextVisitor implements XMLAstVisitor {
+  public hoverContext: XMLHoverContext | undefined;
+  constructor(public chosenOffset: number | undefined) {}
```
bd82, line 57: Use more whitespace, e.g between block of class fields and the constructor.

bd82: or constructor block and method block

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+  visitXMLElement(node: XMLElement): void {
+    const openName = node.syntax.openName;
+    const closeName = node.syntax.closeName;
+    if (this.isInsideRange(openName)) {
```
bd82, line 61: `isInsideRange` name makes sense if it gets 2 parameters, in that case it should not be a method and instead be a utility function.

If we keep it as a method, perhaps call it something like `isHoverOffsetInsideToken`

packages/language-server/src/hover.ts
bd82, line 57: `chosenOffset` could be called just `offset` as we can understand the `hover` part from the context (inside an HoverVisitor).

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+      this.hoverContext = { astNode: node, kind: "XMLElementHover" };
+    }
+
+    if (this.isInsideRange(closeName)) {
+      this.hoverContext = { astNode: node, kind: "XMLElementHover" };
+    }
+  }
+
+  visitXMLAttribute?(node: XMLAttribute): void {
```
bd82, line 70: what is this question mark?

packages/language-server/src/hover.ts
bd82, line 57: I think the `offset` should always be a number, when could it be undefined?

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+
+    if (this.isInsideRange(value)) {
+      this.hoverContext = { astNode: node, kind: "XMLValueAttributeHover" };
+    }
+  }
+
+  isInsideRange(name: XMLToken | undefined): boolean {
+    if (
+      name !== undefined &&
+      this.chosenOffset !== undefined &&
```
bd82, line 86: if `offset` is always a number we won't need the undefined check

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+      this.hoverContext = { astNode: node, kind: "XMLKeyAttributeHover" };
+    }
+
+    if (this.isInsideRange(value)) {
+      this.hoverContext = { astNode: node, kind: "XMLValueAttributeHover" };
+    }
+  }
+
+  isInsideRange(name: XMLToken | undefined): boolean {
+    if (
```
bd82, line 84: you could return the result of the condition directly without wrapping in an `if/else`

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+export function getHoverContext(
+  model: UI5SemanticModel,
+  textDocumentPosition: TextDocumentPositionParams,
+  document: TextDocument
+): Hover | undefined {
+  const documentText = document.getText();
+  const { cst, tokenVector } = parse(documentText);
+  const ast = buildAst(cst as DocumentCstNode, tokenVector);
+  const offset = document.offsetAt(textDocumentPosition.position);
+  const ui5Node = getUI5Node(ast, offset, model);
```
bd82, line 106: `getUI5Node` -- > `findUI5Node` maybe even `findUI5NodeAtOffset` ?

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+  if (ui5Node != undefined) {
+    return transformToLspHover(ui5Node, model);
```
bd82, line 108: `transformToLspHover` -> `transformToLspHoverResponse`?

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  return undefined;
+}
+
+function getUI5Node(
+  ast: XMLDocument,
```
bd82, line 115: Generally with over 2 params it is recommended to use Javascript Config Object pattern.
- https://stackoverflow.com/questions/7466817/javascript-configuration-pattern

bd82: This won't get any UI5Nodes, just HoverSpecific Nodes maybe even call it:

`findUI5HoverNodeAtOffset` (ugly and long I know, but also accurate).

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+  offset: number,
+  model: UI5SemanticModel
+): BaseUI5Node | undefined {
```
bd82, line 118: We can probably be more specific than `BaseUI5Node` and define a "HoverUI5Node` ,e.g:
```typescript
type HoverUI5Node = UI5Class | UI5Attribute | ....;
```

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+    return undefined;
+  }
+  switch (visitor.hoverContext?.kind) {
+    case "XMLElementHover":
+      return getUI5NodeByElement(visitor.hoverContext.astNode, model);
+    case "XMLKeyAttributeHover":
+      return getUI5NodeByKey(visitor.hoverContext.astNode, model);
+    case "XMLValueAttributeHover":
+      return getUI5NodeByValue(visitor.hoverContext.astNode, model);
+    case undefined:
```
bd82, line 131: 1. I don't think we need `case undefined:`
2. While assertNever did not work, we can still do a `default:` case and throw an error there for "exhaustiveness check"

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+  ui5Node: BaseUI5Node,
+  model: UI5SemanticModel
+): Hover | undefined {
+  const hoverItem: Hover = {
+    contents: getNodeDocumentation(ui5Node, model),
+  };
+  return hoverItem;
+}
+
+function getUI5NodeByElement(
```
bd82, line 146: I believe @tal-sapan introduced utilities for both these cases (aggregation / class).
I think we should re-use those.

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+
+  if (astNode.parent.type === "XMLDocument" || ui5Class != undefined) {
+    return getClassByElement(astNode, model);
+  } else {
+    const parentElementClass = getClassByElement(astNode.parent, model);
+    return findAggragationByName(parentElementClass, astNode.name);
+  }
+}
+
+function getUI5NodeByKey(
```
bd82, line 163: Key is a too generic term, what about `getUI5NodeByXMLAttribute`?

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+  astNode: XMLAttribute,
+  model: UI5SemanticModel
+): UI5Prop | UI5Event | UI5Association | undefined {
+  const parentElementClass = getClassByElement(astNode.parent, model);
+  return findUI5ClassMemberByName(parentElementClass, astNode.key);
+}
+
+function getUI5NodeByValue(
```
bd82, line 171: similar to above comment

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+  astNode: XMLAttribute,
+  model: UI5SemanticModel
+): UI5Namespace | UI5EnumValue | undefined {
+  if (astNode.key != null && isXMLNamespaceKey(astNode.key)) {
```
bd82, line 175: `!=` -> `!==`

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    const ui5Enum = find(propType.fields, ["name", value]);
+    return ui5Enum;
+  }
+
+  return undefined;
+}
+
+export function isXMLNamespaceKey(key: string): boolean {
```
bd82, line 244: @tal-sapan is also creating utilities related to this in \@xml-tools, we should re-use them (when merged)

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+function getUI5NodeByKey(
+  astNode: XMLAttribute,
+  model: UI5SemanticModel
+): UI5Prop | UI5Event | UI5Association | undefined {
+  const parentElementClass = getClassByElement(astNode.parent, model);
+  return findUI5ClassMemberByName(parentElementClass, astNode.key);
+}
+
+function getUI5NodeByValue(
+  astNode: XMLAttribute,
```
bd82, line 172: Lets be more specific with the param name, its an `attribute` / `attrib` / `xmlAttrib` / ... 
not a **general** astNode.

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+  model: UI5SemanticModel
+): UI5Namespace | UI5EnumValue | undefined {
+  if (astNode.key != null && isXMLNamespaceKey(astNode.key)) {
+    const ui5Namespace = find(
+      model.namespaces,
+      (namespace) => astNode.value === ui5NodeToFQN(namespace)
+    );
+    return ui5Namespace;
+  } else {
```
bd82, line 181: Open question: 

Should we assume it must be an enum and try to resolve as an enum
or should we first check if its an enum and only then handle the enum case?

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+    const tag = astNode.parent;
+    const elementClass = getClassByElement(tag, model);
+    return getUI5EnumByElement(elementClass, astNode.key, astNode.value);
+  }
+
+  return undefined;
```
bd82, line 187: If we have an `if/else` with `return statement` in both branches, is this `return undefined` reachable?

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+}
+
+function findUI5ClassMemberByName(
```
bd82, line 190: This should probably go into logic-utils

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+  ui5Class: UI5Class | undefined,
+  targetName: string | null
+): UI5Prop | UI5Event | UI5Association | undefined {
```
bd82, line 193: I saw some code by @tal-sapan just today that showed even aggregations can sometimes be in XML Attributes.

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+    const found = find(allClassMembers, (_) => _.name === targetName);
+    return found;
+  }
+
+  return undefined;
+}
+
+function findAggragationByName(
+  ui5Class: UI5Class | undefined,
+  name: string | null
```
bd82, line 210: `targetName`?

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+    const elementClass = getClassByElement(tag, model);
+    return getUI5EnumByElement(elementClass, astNode.key, astNode.value);
+  }
+
+  return undefined;
+}
+
+function findUI5ClassMemberByName(
+  ui5Class: UI5Class | undefined,
+  targetName: string | null
```
bd82, line 192: should we support `targetName : null` case?

packages/language-server/src/hover.ts
bd82: should we support `targetName : null` case?

packages/language-server/src/hover.ts
tal-sapan: See https://github.com/SAP/xml-tools/pull/173

packages/language-server/src/hover.ts
tal-sapan: That's correct, aggregations can be bound inside the class tag.
See here: https://sapui5.hana.ondemand.com/#/topic/91f057786f4d1014b6dd926db0e91070

packages/language-server/src/hover.ts
tal-sapan: See https://github.com/SAP/ui5-language-assistant/pull/120

packages/language-server/src/hover.ts
tal-sapan: I suggest `isOffsetInToken`

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+  | XMLValueAttributeHover;
+
+class HoverContextVisitor implements XMLAstVisitor {
+  public hoverContext: XMLHoverContext | undefined;
+  constructor(public chosenOffset: number | undefined) {}
+  visitXMLElement(node: XMLElement): void {
+    const openName = node.syntax.openName;
+    const closeName = node.syntax.closeName;
+    if (this.isInsideRange(openName)) {
+      this.hoverContext = { astNode: node, kind: "XMLElementHover" };
```
tal-sapan, line 62: Please specify the `kind` property first

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    if (this.isInsideRange(closeName)) {
+      this.hoverContext = { astNode: node, kind: "XMLElementHover" };
```
tal-sapan, line 66: Please specify the `kind` property first.
Also, if the `closeName` is different from the `openName` we should show the `closeName` information (regardless of the tag name mismatch validation). Would this require changing the code here? Maybe adding a new kind for `XMLElementCloseTagHover`?

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+    }
+  }
+
+  visitXMLAttribute?(node: XMLAttribute): void {
+    const key = node.syntax.key;
+    const value = node.syntax.value;
+
+    if (this.isInsideRange(key)) {
```
tal-sapan, line 74: **Future:** we should handle hover on attribute key namespace (it should show the namespace documentation)

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+      this.hoverContext = { astNode: node, kind: "XMLKeyAttributeHover" };
+    }
+
+    if (this.isInsideRange(value)) {
+      this.hoverContext = { astNode: node, kind: "XMLValueAttributeHover" };
+    }
+  }
+
+  isInsideRange(name: XMLToken | undefined): boolean {
```
tal-sapan, line 83: You can accept `SourcePosition` here instead of `XMLToken`, and rename `name` to `position`. Then maybe rename the function to `isOffsetInRange`.

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+      name.endOffset >= this.chosenOffset
+    ) {
+      return true;
+    }
+
+    return false;
+  }
+}
+
+export function getHoverContext(
```
tal-sapan, line 97: Rename to `getHover` or `getHoverResponse`?

packages/language-server/src/hover.ts
tal-sapan: I like `findUI5NodeAtOffset`

packages/language-server/src/hover.ts
tal-sapan: @bd82 since the return type is called `Hover` I think the function name is ok here

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+}
+
+function getUI5Node(
+  ast: XMLDocument,
+  offset: number,
+  model: UI5SemanticModel
+): BaseUI5Node | undefined {
+  const visitor = new HoverContextVisitor(offset);
+  accept(ast, visitor);
+  if (visitor.hoverContext?.kind === undefined) {
```
tal-sapan, line 121: Change the condition to `if (visitor.hoverContext === undefined)` and then you won't have the undefined case in the `switch` and also you won't have to access `visitor.hoverContext` attributes with a `?` (if it doesn't work extract it to a const before the `if`)

packages/language-server/src/hover.ts
tal-sapan, line 131: If `visitor.hoverContext` is not undefined then `assertNever` should work (even if you do't throw an exception)

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+      return undefined;
+  }
+}
+
+function transformToLspHover(
+  ui5Node: BaseUI5Node,
+  model: UI5SemanticModel
+): Hover | undefined {
+  const hoverItem: Hover = {
+    contents: getNodeDocumentation(ui5Node, model),
```
tal-sapan, line 141: **Future:** add the range information (this should be returned from the visitor, I think, and transformed to LSP Range here)

tal-sapan, line 141: I think we should also add the detail (like in `CompletionItem`) as a sort of title, with a linebreak or 3 between them (check what looks best)

packages/language-server/src/hover.ts
tal-sapan: Yes, please re-use `getUI5ClassByXMLElement` and `getUI5AggregationByXMLElement` from `xml-node-to-ui5-node.ts`

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+  const parentElementClass = getClassByElement(astNode.parent, model);
+  return findUI5ClassMemberByName(parentElementClass, astNode.key);
+}
+
+function getUI5NodeByValue(
+  astNode: XMLAttribute,
+  model: UI5SemanticModel
+): UI5Namespace | UI5EnumValue | undefined {
+  if (astNode.key != null && isXMLNamespaceKey(astNode.key)) {
+    const ui5Namespace = find(
```
tal-sapan, line 176: I'm not 100% sure that only UI5 Namespaces can appear in `xmlns` attributes. Some classes exist under non-namespaces. Use `findSymbol` from `semantic-model` instead.

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+      model.namespaces,
+      (namespace) => astNode.value === ui5NodeToFQN(namespace)
+    );
+    return ui5Namespace;
+  } else {
+    const tag = astNode.parent;
```
tal-sapan, line 182: Please rename `tag` to `element` or `xmlElement`

packages/language-server/src/hover.ts
tal-sapan, line 181: You can remove the `else` because you have a `return` statement right above it. It will reduce the nesting levels, especially if we get more cases for attribute value.

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+    const elementClass = getClassByElement(tag, model);
+    return getUI5EnumByElement(elementClass, astNode.key, astNode.value);
+  }
+
+  return undefined;
+}
+
+function findUI5ClassMemberByName(
+  ui5Class: UI5Class | undefined,
```
tal-sapan, line 191: I think `ui5Class` shouldn't be allowed to be `undefined` here. It should be checked before calling this function.

packages/language-server/src/hover.ts
tal-sapan, line 192: Can you call it `name`?

packages/language-server/src/hover.ts
tal-sapan: I agree but make it receive the attribute instead of its key

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+): UI5Prop | UI5Event | UI5Association | undefined {
+  if (ui5Class != undefined) {
+    const allProps: (UI5Prop | UI5Event | UI5Association)[] = flattenProperties(
+      ui5Class
+    );
+    const allEvents = flattenEvents(ui5Class);
+    const allAssociations = flattenAssociations(ui5Class);
+    const allClassMembers = allProps.concat(allEvents).concat(allAssociations);
+    const found = find(allClassMembers, (_) => _.name === targetName);
+    return found;
```
tal-sapan, line 202: This is very similar to what I did, I will extract it to a function

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  return undefined;
+}
+
+function findAggragationByName(
```
tal-sapan, line 208: This function is not necessary if you use `getUI5AggregationByXMLElement` as I mentioned above

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+  if (astNode.key != null && isXMLNamespaceKey(astNode.key)) {
+    const ui5Namespace = find(
+      model.namespaces,
+      (namespace) => astNode.value === ui5NodeToFQN(namespace)
+    );
+    return ui5Namespace;
+  } else {
+    const tag = astNode.parent;
+    const elementClass = getClassByElement(tag, model);
+    return getUI5EnumByElement(elementClass, astNode.key, astNode.value);
```
tal-sapan, line 184: I suggest to extract this to 2 functions: (1) `getEnumProperty` which will return the property if it's an enum (using `getUI5PropertyByXMLAttributeKey`), then if it's not `undefined`, return (2) `getEnumValueByName` on the returned enum and the attribute value

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+
+    const ui5Enum = find(propType.fields, ["name", value]);
+    return ui5Enum;
+  }
+
+  return undefined;
+}
+
+export function isXMLNamespaceKey(key: string): boolean {
+  const namespaceRegex = /^xmlns(:(?<prefix>\w*))?$/;
```
tal-sapan, line 245: Additionally this regex isn't correct (it misses some cases)

packages/language-server/src/hover.ts
tal-sapan, line 244: Even before it's merged you can use `isXMLNamespaceKey` from logic-utils

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,255 @@
...
+  return key.match(namespaceRegex) !== null;
+}
+
+export function getClassByElement(
```
tal-sapan, line 249: This function exists in logic-utils (`getUI5ClassByXMLElement`)

packages/language-server/src/hover.ts
tal-sapan: I prefer to do it in another PR so you can keep this code for now and if I add it before this PR is merged I will add another comment about it

packages/language-server/src/hover.ts
bd82: Good point

packages/language-server/src/hover.ts
bd82: I think this will become something like `getAstNodeAtOffset` when moved to XML-Tools

packages/language-server/src/hover.ts
bd82: Great, it is useful when TSC does not get too confused 😄

packages/language-server/src/hover.ts
bd82: I think I suggested `targetName` to help Sapir understand the logical flow
and avoid `_.name === name` which is a little ambiguous.

tal-sapan: ok

packages/language-server/src/hover.ts
tal-sapan: It is merged

packages/language-server/src/hover.ts
tal-sapan: The PR I mentioned was merged

packages/xml-views-tooltip/src/documentation.ts
```diff
@@ -0,0 +1,154 @@
+import {
```
tal-sapan, line 1: This file should be in logic-utils. Since you copied it as-is, please put a comment at the top that says where you copied it from, add an `istanbul ignore file` comment that says it will move to logic-utils and will be tested there soon.

packages/xml-views-tooltip/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "devDependencies": {},
+  "scripts": {
+    "ci": "npm-run-all clean compile lint coverage",
+    "clean": "rimraf ./lib ./coverage ./nyc_output",
+    "compile": "yarn run clean && tsc -p .",
+    "compile:watch": "tsc -p . --watch",
+    "lint": "eslint . --ext .ts --max-warnings=0 --ignore-path=../../.gitignore",
+    "test": "mocha",
+    "coverage": "nyc mocha",
+    "update-snapshots": "node ./scripts/update-diagnostics-snapshots"
```
tal-sapan, line 41: This is irrelevant here, please remove

packages/xml-views-tooltip/src/tooltip.ts
```diff
@@ -0,0 +1,140 @@
...
+  UI5Aggregation,
+  UI5SemanticModel,
+  UI5EnumValue,
+  UI5Enum,
+  BaseUI5Node,
+} from "@ui5-language-assistant/semantic-model-types";
+import { findSymbol } from "@ui5-language-assistant/semantic-model";
+
+export function findUI5HoverNodeAtOffset(
+  visitor:
```
tal-sapan, line 28: Rename to `astPosition`

packages/xml-views-tooltip/src/tooltip.ts
```diff
@@ -0,0 +1,140 @@
...
+      return getUI5PropertyByXMLAttributeKey(visitor.astNode, model);
+    case "XMLAttributeValue":
+      return findUI5NodeByXMLAttributeValue(visitor.astNode, model);
+  }
+}
+
+function getUI5NodeByElement(
+  astNode: XMLElement,
+  model: UI5SemanticModel,
+  kind: string
```
tal-sapan, line 54: Don't send the `kind` here but a boolean `isOpenName` (or `isClosingName`)

packages/xml-views-tooltip/src/tooltip.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+  switch (visitor.kind) {
+    case "XMLElementOpenName":
+      return getUI5NodeByElement(visitor.astNode, model, visitor.kind);
+    case "XMLElementCloseName":
+      return getUI5NodeByElement(visitor.astNode, model, visitor.kind);
+    case "XMLAttributeKey":
+      return getUI5PropertyByXMLAttributeKey(visitor.astNode, model);
+    case "XMLAttributeValue":
+      return findUI5NodeByXMLAttributeValue(visitor.astNode, model);
```
tal-sapan, line 47: Add `assertNever` in the default case (so we don't forget to update this function if we support more types)

packages/xml-views-tooltip/src/tooltip.ts
```diff
@@ -0,0 +1,140 @@
...
+function getUI5NodeByElement(
+  astNode: XMLElement,
+  model: UI5SemanticModel,
+  kind: string
+): UI5Class | UI5Aggregation | undefined {
+  const fqnClassName =
+    kind === "XMLElementOpenName"
+      ? xmlToFQN(astNode)
+      : elementClosingTagToFQN(astNode);
+  const ui5Class = find(
```
tal-sapan, line 60: `model.classes` is a map from fqn to the class, don't use `find`

packages/xml-views-tooltip/src/tooltip.ts
```diff
@@ -0,0 +1,140 @@
...
+  const parentElementClass = getUI5ClassByXMLElement(astNode.parent, model);
+  const nameByKind =
+    kind === "XMLElementOpenName"
+      ? astNode.syntax.openName?.image
+      : astNode.syntax.closeName?.image;
+  return findAggragationByName(parentElementClass, nameByKind);
+}
+
+function findAggragationByName(
+  ui5Class: UI5Class | undefined,
```
tal-sapan, line 78: This should not be allowed to be `undefined`

packages/xml-views-tooltip/src/tooltip.ts
```diff
@@ -0,0 +1,140 @@
...
+  targetName: string | undefined
```
tal-sapan, line 79: This should not be allowed to be `undefined`

packages/xml-views-tooltip/src/tooltip.ts
```diff
@@ -0,0 +1,140 @@
...
+): UI5Aggregation | undefined {
+  if (ui5Class != undefined) {
```
tal-sapan, line 81: Use `!==`, not `!=`
Also when `ui5Class` is never undefined, you won't need this condition

packages/xml-views-tooltip/src/tooltip.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    return ui5Aggregation;
+  }
+
+  return undefined;
+}
+
+function elementClosingTagToFQN(astElement: XMLElement): string {
+  const baseName = astElement.syntax.closeName?.image
+    ? astElement.syntax.closeName?.image
```
tal-sapan, line 96: Use `??` instead of `? :`

packages/xml-views-tooltip/src/tooltip.ts
```diff
@@ -0,0 +1,140 @@
...
+    : "";
+  const prefixXmlns = astElement.ns ? astElement.ns : DEFAULT_NS;
```
tal-sapan, line 98: The `ns` property is taken from the opening tag. You have to split the closing tag name to namespace and local name yourself. You can see how I did it for attributes (it's the same logic) in `uknown-attribute-key.ts` (function `splitAttributeByNamespace`)

packages/xml-views-tooltip/src/tooltip.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  return getUI5EnumByElement(attribute, model);
+}
+
+function getUI5EnumByElement(
+  attribute: XMLAttribute,
+  model: UI5SemanticModel
+): UI5EnumValue | undefined {
+  const enumProp = getEnumProperty(attribute, model);
+  const enumValue = find(enumProp?.fields, ["name", attribute.value]);
```
tal-sapan, line 129: Check if `enumProp === undefined`

packages/xml-views-tooltip/src/tooltip.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  return enumValue;
```
tal-sapan, line 131: Remove the blank lines

packages/xml-views-tooltip/package.json
```diff
@@ -0,0 +1,46 @@
...
+  },
+  "dependencies": {
+    "@ui5-language-assistant/semantic-model": "^1.2.1",
+    "@ui5-language-assistant/semantic-model-types": "1.2.0",
+    "@ui5-language-assistant/logic-utils": "1.1.3",
+    "@xml-tools/ast": "4.2.0",
+    "@xml-tools/ast-position": "1.1.0",
+    "@xml-tools/common": "0.1.0",
+    "lodash": "4.17.15",
+    "vscode-languageserver": "6.1.1",
```
tal-sapan, line 29: I don't think this is used in this package, please remove it

packages/language-server/src/hover-tooltip.ts
```diff
@@ -0,0 +1,63 @@
...
+export function getHoverResponse(
+  model: UI5SemanticModel,
+  textDocumentPosition: TextDocumentPositionParams,
+  document: TextDocument
+): Hover | undefined {
+  const documentText = document.getText();
+  const { cst, tokenVector } = parse(documentText);
+  const ast = buildAst(cst as DocumentCstNode, tokenVector);
+  const offset = document.offsetAt(textDocumentPosition.position);
+  const visitor = getAstNodeInPosition(ast, offset);
```
tal-sapan, line 46: Rename to `astPosition`

packages/language-server/src/hover-tooltip.ts
```diff
@@ -0,0 +1,63 @@
...
+import {
+  UI5Class,
+  UI5Aggregation,
+  UI5SemanticModel,
+  UI5Namespace,
+  UI5EnumValue,
+  UI5Enum,
+  BaseUI5Node,
+} from "@ui5-language-assistant/semantic-model-types";
+import { getNodeDocumentation } from "./documentation";
```
tal-sapan, line 33: This should be the last import

packages/language-server/src/hover-tooltip.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  return undefined;
+}
+
+function transformToLspHover(
+  ui5Node: BaseUI5Node,
+  model: UI5SemanticModel
+): Hover | undefined {
+  const hoverItem: Hover = {
+    contents: getNodeDocumentation(ui5Node, model),
```
tal-sapan, line 60: The hover should have some sort of title, not just the documentation. Like the `details` for completion item (you can reuse/copy it). Add it in the beginning of the documentation followed by a new line (or more than 1 if it doesn't look good).

packages/language-server/src/hover-tooltip.ts
```diff
@@ -0,0 +1,63 @@
...
+  textDocumentPosition: TextDocumentPositionParams,
+  document: TextDocument
+): Hover | undefined {
+  const documentText = document.getText();
+  const { cst, tokenVector } = parse(documentText);
+  const ast = buildAst(cst as DocumentCstNode, tokenVector);
+  const offset = document.offsetAt(textDocumentPosition.position);
+  const visitor = getAstNodeInPosition(ast, offset);
+  const ui5Node = findUI5HoverNodeAtOffset(visitor, model);
+  if (ui5Node != undefined) {
```
tal-sapan, line 48: `!==`

packages/xml-views-tooltip/src/tooltip.ts
```diff
@@ -0,0 +1,140 @@
...
+  const fqnClassName =
+    kind === "XMLElementOpenName"
+      ? xmlToFQN(astNode)
+      : elementClosingTagToFQN(astNode);
+  const ui5Class = find(
+    model.classes,
+    (ui5class) => fqnClassName === ui5NodeToFQN(ui5class)
+  );
+
+  if (astNode.parent.type === "XMLDocument" || ui5Class != undefined) {
```
tal-sapan, line 65: `!==`

packages/xml-views-tooltip/src/tooltip.ts
```diff
@@ -0,0 +1,140 @@
...
+  if (visitor === undefined) {
+    return undefined;
+  }
+  switch (visitor.kind) {
+    case "XMLElementOpenName":
+      return getUI5NodeByElement(visitor.astNode, model, visitor.kind);
+    case "XMLElementCloseName":
+      return getUI5NodeByElement(visitor.astNode, model, visitor.kind);
+    case "XMLAttributeKey":
+      return getUI5PropertyByXMLAttributeKey(visitor.astNode, model);
```
tal-sapan, line 45: This might not be a property (it could be an event, association, aggregation)

packages/language-server/test/tooltip-spec.ts
```diff
@@ -0,0 +1,188 @@
...
+    const xmlSnippet = `<mvc:View
+                          xmlns:mvc="sap.ui.core.mvc"
+                          xmlns="sap.m"> 
+                          <content>
+                            <List showSeparator⇶s="All"></List>
+                          </content>
+                        </mvc:View>`;
+    const response = getHoverItem(xmlSnippet, ui5SemanticModel);
+    expectExists(response, "Hover item");
+    expect(JSON.stringify(response.contents)).to.include(
```
tal-sapan, line 32: Instead of `JSON.stringify` check `response.contents.value` (in all tests)

packages/language-server/test/tooltip-spec.ts
```diff
@@ -0,0 +1,188 @@
...
+  TextDocumentPositionParams,
+} from "vscode-languageserver";
+
+describe.only("the UI5 language assistant Hover Tooltip Service", () => {
+  let ui5SemanticModel: UI5SemanticModel;
+  before(async () => {
+    ui5SemanticModel = await generateModel({ version: "1.74.0" });
+  });
+
+  it("will get hover content UI5 property", () => {
```
tal-sapan, line 22: Add `context` according to ast position (e.g. `context("hover on attribute key")` which will have this test, `context("hover on attribute value")` which will have the enum and namespace in xmlns tests)

packages/language-server/test/tooltip-spec.ts
```diff
@@ -0,0 +1,188 @@
...
+                          </content>
+                        </mvc:View>`;
+    const response = getHoverItem(xmlSnippet, ui5SemanticModel);
+    expectExists(response, "Hover item");
+    expect(JSON.stringify(response.contents)).to.include(
+      "Separators between the items including the last and the first one."
+    );
+  });
+
+  it("will get hover content UI5 namespace", () => {
```
tal-sapan, line 52: add - "in xmlns attribute"

packages/language-server/test/tooltip-spec.ts
```diff
@@ -0,0 +1,188 @@
...
+    expect(JSON.stringify(response.contents)).to.include(
+      "Child Controls of the view"
+    );
+  });
+
+  it("will get hover content UI5 Aggregation - close tag", () => {
+    const xmlSnippet = `<mvc:View
+                          xmlns:mvc="sap.ui.core.mvc"
+                          xmlns="sap.m"> 
+                          <conten⇶t>
```
tal-sapan, line 116: The cursor is in the open tag, it should be in the close tag

packages/language-server/test/tooltip-spec.ts
```diff
@@ -0,0 +1,188 @@
...
+                          <content></L⇶ist>
+                        </mvc:View>`;
+    const response = getHoverItem(xmlSnippet, ui5SemanticModel);
+    expectExists(response, "Hover item");
+    expect(JSON.stringify(response.contents)).to.include(
+      "The List control provides a container for all types of list items."
+    );
+  });
+
+  it("will get hover content UI5 Aggregation - close tag", () => {
```
tal-sapan, line 140: The description is wrong here

packages/language-server/test/tooltip-spec.ts
```diff
@@ -0,0 +1,188 @@
...
+    document
+  );
+
+  return hoverItem;
+}
+
+function getXmlSnippet(
+  xmlSnippet: string
+): { document: TextDocument; position: Position } {
+  const xmlSnippetWithoutRanges = xmlSnippet
```
tal-sapan, line 176: You don't need to handle ranges in this function

packages/language-server/test/tooltip-spec.ts
```diff
@@ -0,0 +1,188 @@
...
+} from "@ui5-language-assistant/test-utils";
+import {
+  TextDocument,
+  Position,
+  Hover,
+  TextDocumentIdentifier,
+  TextDocumentPositionParams,
+} from "vscode-languageserver";
+
+describe.only("the UI5 language assistant Hover Tooltip Service", () => {
```
tal-sapan, line 16: Add tests for open/close tag:
- Open tag with namespace
- Close tag with namespace (same as open tag)
- Close tag without namespace and no default namespace defined (should return `undefined`)
- Close tag with namespace which is is different from open tag namespace (should use the close tag namespace)
- Close tag without namespace when the open tag has a namespace (should use the default namespace)

tal-sapan: It can be in the tooltip package (doesn't have to be in language-server)

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,90 @@
...
+  textDocumentPosition: TextDocumentPositionParams,
+  document: TextDocument
+): Hover | undefined {
+  const documentText = document.getText();
+  const { cst, tokenVector } = parse(documentText);
+  const ast = buildAst(cst as DocumentCstNode, tokenVector);
+  const offset = document.offsetAt(textDocumentPosition.position);
+  const astPosition = getAstNodeInPosition(ast, offset);
+  if (astPosition !== undefined) {
+    const ui5Node = findUI5HoverNodeAtOffset(astPosition, model);
```
tal-sapan, line 38: **Not for now** but maybe we should rename `findUI5HoverNodeAtOffset` because to `findUI5HoverNodeAtPosition`

packages/language-server/src/hover.ts
```diff
@@ -0,0 +1,90 @@
...
+  };
+  return hoverItem;
+}
+
+function addTitle(title: string, documentation: MarkupContent): MarkupContent {
+  const markupWithTitle = "`" + title + "`\n\n\n" + documentation.value;
+  return { kind: MarkupKind.Markdown, value: markupWithTitle };
+}
+
+function getNodeDetail(node: BaseUI5Node): string {
```
tal-sapan, line 65: Add a comment that this was copied from completion-items and should be moved to logic-utils

packages/xml-views-tooltip/README.md
```diff
@@ -1,11 +1,8 @@
...
-# @ui5-language-assistant/xml-views-completion
+# @ui5-language-assistant/xml-views-tooltip
 
-Logic for UI5 XML views content assist implemented by combining:
-
-- [@xml-tools/content-assist](https://github.com/sap/xml-tools/tree/master/packages/content-assist) for the generic syntactic XML information.
-- [@ui5-language-assistant/semantic-model](../semantic-model) for the semantic UI5 data.
+Logic for UI5 XML views hover tooltip.
 
 Supported Content Assist scenarios:
```
tal-sapan, line 7:
```suggestion
Supported tooltip scenarios:
```

packages/xml-views-tooltip/README.md
```diff
@@ -23,17 +20,17 @@ Supported Content Assist scenarios:
 
 ### In XML `xmlns` attributes keys:
 
-- suggestions in the xmlns prefix part.
+- Hover on xmlns.
```
tal-sapan, line 23: This should be removed (there is no hover on xmlns attribute **keys**)

packages/xml-views-tooltip/src/tooltip.ts
```diff
@@ -21,88 +25,135 @@ import {
...
+}
+
+function splitQNameByNamespace(
+  qName: string
+): { ns: string | undefined; name: string } {
+  if (!includes(qName, ":")) {
+    return { name: qName, ns: undefined };
+  }
+  const match = qName.match(/(?<ns>[^:]*)(:(?<name>.*))?/);
+  // There will always be a match because the attribute key always contains a colon at this point
```
tal-sapan, line 106: qName instead of "the attribute key"

packages/xml-views-tooltip/test/tooltip-spec.ts
```diff
@@ -0,0 +1,103 @@
...
+import {
+  TextDocument,
+  Position,
+  Hover,
+  TextDocumentIdentifier,
+  TextDocumentPositionParams,
+} from "vscode-languageserver";
+import { buildAst } from "@xml-tools/ast";
+
+describe.only("the UI5 language assistant Hover Tooltip Service", () => {
```
tal-sapan, line 22: Remove the `.only`

packages/xml-views-tooltip/test/tooltip-spec.ts
```diff
@@ -0,0 +1,103 @@
...
+    return ui5Node;
+  }
+
+  return undefined;
+}
+
+function getXmlSnippet(
+  xmlSnippet: string
+): { document: TextDocument; position: Position } {
+  const xmlSnippetWithoutRanges = xmlSnippet
```
tal-sapan, line 91: You don't need to handle ranges in this function

packages/language-server/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ui5-language-assistant/language-server",
-  "version": "1.5.1",
+  "version": "1.5.0",
```
tal-sapan, line 3: Please revert this change

packages/xml-views-tooltip/src/tooltip.ts
```diff
@@ -75,12 +72,15 @@ function findUI5NodeByElement(
     return undefined;
   }
   const nameByKind = isOpenName
-    ? astNode.syntax.openName?.image
-    : astNode.syntax.closeName?.image;
+    ? /* istanbul ignore next */
```
tal-sapan, line 75: Add a comment that the openName/closeName cannot be undefined here because otherwise the ast position visitor wouldn't return their types

packages/xml-views-tooltip/src/tooltip.ts
```diff
@@ -108,11 +108,15 @@ function splitQNameByNamespace(
...
-    name: matchGroups.name ?? "",
+    name:
+      matchGroups.name ??
+      /* istanbul ignore next */
+      "",
   };
 }
 
 function elementClosingTagToFQN(xmlElement: XMLElement): string {
+  /* istanbul ignore next */
```
tal-sapan, line 119: Add a comment that the closeName can't be undefined here because otherwise the ast position visitor wouldn't return its type

packages/xml-views-tooltip/src/tooltip.ts
```diff
@@ -132,7 +136,8 @@ function findUI5NodeByXMLAttributeKey(
   const parentElementClass = getUI5ClassByXMLElement(astNode.parent, model);
   return parentElementClass
     ? findUI5ClassMemberByName(parentElementClass, astNode.key)
-    : undefined;
+    : /* istanbul ignore next */
```
tal-sapan, line 139: You can test this case - hover on attribute in unknown tag

packages/xml-views-tooltip/test/tooltip-spec.ts
```diff
@@ -10,61 +10,211 @@ import {
...
+                              <List showSeparators="Al⇶l"></List>
+                            </content>
+                          </mvc:View>`;
+      const ui5Node = getUI5Node(xmlSnippet, ui5SemanticModel);
+      expectExists(ui5Node, "UI5Node");
+      expect(ui5Node.name).to.equal("All");
+      expect(ui5Node.kind).equal("UI5EnumValue");
+    });
+
+    it("will get hover content UI5 enum without information", () => {
```
tal-sapan, line 65: It's not an enum, it's a non-enum property (boolean). You can write "UI5 property value for attribute which is not an enum"

packages/xml-views-tooltip/test/tooltip-spec.ts
```diff
@@ -10,61 +10,211 @@ import {
...
+                              <List showSeparators="All"></List>
+                            </content>
+                          </mvc:View>`;
+      const ui5Node = getUI5Node(xmlSnippet, ui5SemanticModel);
+      expectExists(ui5Node, "UI5Node");
+      expect(ui5Node.name).to.equal("content");
+      expect(ui5Node.kind).equal("UI5Aggregation");
+    });
+
+    it("will get hover content UI5 namespace", () => {
```
tal-sapan, line 119: This should be in a different `context` (it's an attribute value)

packages/xml-views-tooltip/test/tooltip-spec.ts
```diff
@@ -10,61 +10,211 @@ import {
...
+                              <List showSeparators="All"></List>
+                            </content>
+                          </mvc:View>`;
+      const ui5Node = getUI5Node(xmlSnippet, ui5SemanticModel);
+      expectExists(ui5Node, "UI5Node");
+      expect(ui5Node.name).to.equal("m");
+      expect(ui5Node.kind).equal("UI5Namespace");
+    });
+
+    it("will get hover content UI5 class - incorrect class name", () => {
```
tal-sapan, line 133: It's "unknown tag with unknown parent tag"

packages/xml-views-tooltip/test/tooltip-spec.ts
```diff
@@ -10,61 +10,211 @@ import {
...
+                            xmlns:mvc="sap.ui.core.mvc"
+                            xmlns="sap.m"> 
+                          </Butt⇶on>`;
+      const ui5Node = getUI5Node(xmlSnippet, ui5SemanticModel);
+      expectExists(ui5Node, "UI5Node");
+      expect(ui5Node.name).to.equal("Button");
+      expect(ui5Node.kind).equal("UI5Class");
+    });
+
+    it("will get hover content UI5 class when namespace ofopen and close tag are different", () => {
```
tal-sapan, line 186:
```suggestion
    it("will not get hover content when only the close tag has the wrong namespace", () => {
```

packages/xml-views-tooltip/test/tooltip-spec.ts
```diff
@@ -45,6 +45,18 @@ describe("the UI5 language assistant Hover Tooltip Service", () => {
       const ui5Node = getUI5Node(xmlSnippet, ui5SemanticModel);
       expect(ui5Node).to.not.exist;
     });
+
+    it("will get hover content UI5 property with unknown tag", () => {
```
tal-sapan, line 49: "in unknown tag"

packages/xml-views-tooltip/src/documentation.ts
```diff
@@ -1,156 +0,0 @@
-/* istanbul ignore file - soon will move to logic utils with tests */
```
tal-sapan: Did you remove this file accidentally?

PR #120, opened 2020-05-14
feat: validate attribute key is known in class and aggregation tags

tal-sapan:
+ test: fix snapshot creation util to support more than 1 validation issue


tal-sapan:
@bd82 There are a few `TODO`s in the validation code, mainly related to "special" attributes that we might want to move to the model. This will affect code completion as well so I think we should ask the product about it.

packages/xml-views-completion/src/providers/elementName/classes.ts
```diff
@@ -96,55 +93,36 @@ function computeClassSuggestionContext({
...
-      });
-    } else if (startsWithUpperCase.test(parentXMLElementName)) {
-      return handleDefaultAggregationScenario({ parentXMLElement, model });
-    }
-    // unrecognized scenario
-    else {
-      return NOT_FOUND;
-    }
+  // If the parent tag is a class this is an implicit (default) aggregation
+  const parentUI5Class = getUI5ClassByXMLElement(parentXMLElement, model);
```
bd82, line 97: 👍

bd82, line 97: 👍

packages/xml-views-validation/src/utils/special-namespaces.ts
```diff
@@ -0,0 +1,4 @@
+export const TEMPLATING_NS =
```
bd82, line 1: Should we have a map/dictionary instead of const per NS?

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  // Only validate (known) classes and aggregations
+  const xmlElement = attribute.parent;
+  const ui5Class = getUI5ClassByXMLElement(xmlElement, model);
+  let ui5Aggregation: UI5Aggregation | undefined = undefined;
+  if (ui5Class === undefined) {
+    ui5Aggregation = getUI5AggregationByXMLElement(xmlElement, model);
+  }
+
+  // Aggregations don't have additional allowed attributes
```
bd82, line 44: `additional` -> `any` ?

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  if (isAttributeNameAlwaysValid(attribute)) {
+    return [];
+  }
+
+  // Only validate (known) classes and aggregations
+  const xmlElement = attribute.parent;
+  const ui5Class = getUI5ClassByXMLElement(xmlElement, model);
+  let ui5Aggregation: UI5Aggregation | undefined = undefined;
+  if (ui5Class === undefined) {
```
bd82, line 40: Is this `IF` really needed?

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,179 @@
...
+    return [];
+  }
+
+  if (isAttributeNameAlwaysValid(attribute)) {
+    return [];
+  }
+
+  // Only validate (known) classes and aggregations
+  const xmlElement = attribute.parent;
+  const ui5Class = getUI5ClassByXMLElement(xmlElement, model);
```
bd82, line 38: Could ` const ui5Class = getUI5ClassByXMLElement(xmlElement, model);` be moved to line 58?

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,179 @@
...
+        severity: "error",
+      },
+    ];
+  }
+
+  if (ui5Class === undefined) {
+    return [];
+  }
+
+  // Special case: "class" property can be used on all Controls and selected Elements.
```
bd82, line 63: - This is a CSS class?
- Can we extract the white list to a record/array?

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,179 @@
...
+  // TODO Should this be in the model? Should it be suggested for code completion?
+  if (attribute.key === "class") {
+    return [];
+  }
+
+  // Special case: "binding" property is a synonym for "objectBindings" special setting on ManagedObject
+  // TODO Should this be in the model? Should it be suggested for code completion?
+  if (attribute.key === "binding") {
```
bd82, line 71: naively sounds like the model needs to be expended

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,179 @@
...
+    return [];
+  }
+
+  // Special case: "stashed" property can appear under sap.uxap.ObjectPageLazyLoader
+  // Technically it can be under any Element but it should only be used for this one.
+  // TODO Should this be in the model? Should it be suggested for code completion?
+  if (
```
bd82, line 78: again naively seems like a concern of the model.

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,179 @@
...
+    ui5NodeToFQN(ui5Class) === "sap.uxap.ObjectPageLazyLoader"
+  ) {
+    return [];
+  }
+
+  const allProps: (
+    | UI5Prop
+    | UI5Event
+    | UI5Association
+    | UI5Aggregation
```
bd82, line 89: can aggregations appear as XML attributes?

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,179 @@
...
+  )[] = flattenProperties(ui5Class);
+  const allEvents = flattenEvents(ui5Class);
+  const allAssociations = flattenAssociations(ui5Class);
+  // Aggregations can be used as attributes for binding
+  const allAggregations = flattenAggregations(ui5Class);
+  const allClassUI5Attributes = allProps
+    .concat(allEvents)
+    .concat(allAssociations)
+    .concat(allAggregations);
+  // TODO Should we filter out non-public/protected names?
```
bd82, line 99: I naively think usage of private/hidden/top-secret visibility is a **different** validation.

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,179 @@
...
+  flattenProperties,
+  flattenEvents,
+  flattenAssociations,
+  flattenAggregations,
+  ui5NodeToFQN,
+} from "@ui5-language-assistant/logic-utils";
+import { UnknownAttributeKeyIssue } from "../../../api";
+import { TEMPLATING_NS, CUSTOM_DATA_NS } from "../../utils/special-namespaces";
+
+export function validateUnknownAttributeKey(
```
bd82, line 23: Subjective / Style question:

There are many branches which **return**, does it make sense to add more **else if** to clearly distinguish the logical flows instead of relying on needing to recognize the **return** statements?

It won't affect correctness, but would it make the flow easier to grok?

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
bd82: e.g HTML props whitelist?

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,179 @@
...
+      },
+      severity: "error",
+    },
+  ];
+}
+
+function isAttributeNameAlwaysValid(attribute: XMLAttribute): boolean {
+  // This will never happen - it was checked before calling this function
+  /* istanbul ignore if */
+  if (attribute.key === null) {
```
bd82, line 122: may be able to model this at the signature level:

`XMLAttribute & {key:string}`

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: The attributes in `isAttributeNameAlwaysValid` (checked before) are allowed, classes have additional allowed attributes

bd82: can aggregations have `xmlns` attributes?

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: No need to check if it's an aggregation if we already found out it's a class

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,179 @@
...
+    return true;
+  }
+  // xmlns attributes are always valid
+  // TODO "xmlns:" should not return true (change when using new function in xml-utils)
+  if (isXMLNamespaceKey(attribute.key)) {
+    return true;
+  }
+  const { ns, name } = splitAttributeByNamespace(attribute);
+  // sap.ui.dt:designtime is always valid
+  if (ns === "sap.ui.dt" && name === "designtime") {
```
bd82, line 132: This acts as "annotations" in UI5 xml views?

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,179 @@
...
+  }
+  // sap.ui.core:require is always valid
+  // TODO this is not true for FragmentDefinition inside xml views
+  // See: https://openui5nightly.hana.ondemand.com/topic/b11d853a8e784db6b2d210ef57b0f7d7
+  if (ns === "sap.ui.core" && name === "require") {
+    return true;
+  }
+
+  // templates require is always valid
+  if (ns === TEMPLATING_NS && name === "require") {
```
bd82, line 143: I feel as though this whitelist dark magic should be defined outside the logic
in some map/array.

bd82: I mean not just the consts, rather instead of conditional per case, check if the attributeKey is part of the collection.

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: It would work (since aggregations and classes names don't intersect) but the check for class is more straight-forward (it only checks the current tag) so I think it should be done first. (It's also slightly more efficient since we don't go over all the aggregations of the parent class, but I don't think that matters much since there aren't many aggregations for each class.)

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
bd82: That may also simplify tests

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  return false;
+}
+
+function splitAttributeByNamespace(
+  attribute: XMLAttribute
+): { ns: string | undefined; name: string | undefined } {
+  // This will never happen - it was checked before calling this function
+  /* istanbul ignore if */
+  if (attribute.key === null) {
```
bd82, line 160: may be able to model this assumption at the signature level:

XMLAttribute & {key:string}

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,179 @@
...
+    return { name: undefined, ns: undefined };
+  }
+  if (!includes(attribute.key, ":")) {
+    return { name: attribute.key, ns: undefined };
+  }
+  const match = attribute.key.match(/(?<ns>[^:]*)(:(?<name>.*))?/);
+  // There will always be a match because the attribute key always contains a colon at this point
+  /* istanbul ignore next */
```
bd82, line 168: This is a typescript signature problem, perhaps use a cast instead?

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: Only `class` is supported, not all html properties.
HTML tags can be used but they are in a specific namespace (the details are in the task for validating tag names). We don't validate their attributes.
We can have a whitelist for the special properties, but there will still be the special case of custom data namespace (since it doesn't have an attribute name).

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  // All attributes in the custom data namespace are always valid
+  if (ns === CUSTOM_DATA_NS) {
+    return true;
+  }
+
+  return false;
+}
+
+function splitAttributeByNamespace(
```
bd82, line 155: I think very similar logic is done during the AST Building, perhaps 
it should be extracted to XML-tools/commons

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: And yes, it's a css class
https://openui5nightly.hana.ondemand.com/topic/b564935324f449209354c7e2f9903f22

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: Yes. All aggregations can be specified as a binding expression on the class tag, and aggregations of `0..1` cardinality and `atlTypes=[string]` can be specified as a string value.
Maybe we should add them to code completion too.

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: That depends if you prefer `else if`s or smaller nesting levels. I'm used to smaller nesting levels...

packages/xml-views-validation/test/validators/attributes/unknown-attribute-key-spec.ts
```diff
@@ -0,0 +1,810 @@
...
+    ui5SemanticModel = await generateModel({ version: "1.74.0" });
+  });
+
+  context("true positive scenarios", () => {
+    it("will detect an invalid attribute key in root class element", () => {
+      const xmlSnippet = `
+          <mvc:View
+            xmlns:mvc="sap.ui.core.mvc"
+            xmlns="sap.m"
+            🢂busy1🢀="true">
```
bd82, line 28: could we use names that are more clearly wrong, e.g `busy_TYPO` / `busy_OOPS` / ...
- in general for this whole file.

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: I tried exactly this but typescript didn't let me send the attribute even though I checked the `key` isn't null, and I didn't want to cast it

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: Yes

packages/xml-views-validation/test/validators/attributes/unknown-attribute-key-spec.ts
```diff
@@ -0,0 +1,810 @@
...
+        },
+      });
+    });
+
+    it("will detect 'stashed' as invalid in aggregation element", () => {
+      const xmlSnippet = `
+      <mvc:View
+        xmlns:mvc="sap.ui.core.mvc"
+        xmlns="sap.m">
+        <content 🢂stashed🢀="">
```
bd82, line 220: There is a limit of use cases we can reasonably check, maybe using maps/arrays for whitelists instead of conditionals will reduce the cyclomatic complexity and thus reduce the tests needed for 100% coverage.

packages/xml-views-validation/test/validators/attributes/unknown-attribute-key-spec.ts
```diff
@@ -0,0 +1,810 @@
...
+          },
+          assertion: (issues) => {
+            expect(issues).to.be.empty;
+          },
+        });
+      });
+    });
+
+    context("non-reproducible unit tests", () => {
+      it("will not detect an issue when the attribute doesn't have a key", () => {
```
bd82, line 787: may not be worth testing this and instead ignoring this case using `istanbul ignore`

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: It's used to define if a component can be hidden/changed when writing extensible applications (that allow UI adaptation).
It's specified here under "Restrict adaptation for certain controls":
https://sapui5.hana.ondemand.com/#/topic/de9fd55c69af4b46863f5d26b5d796c4

packages/xml-views-validation/test/validators/attributes/unknown-attribute-key-spec.ts
tal-sapan: Even if I extract the special cases to a list I would still like to test specific cases to make sure we didn't miss anything

packages/xml-views-validation/src/utils/special-namespaces.ts
tal-sapan: I added the constants because they will be relevant in other validations too (like for the tag name)

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: I changed the function, hope it's more readable now

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: I asked Yuval, waiting for his answer

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: I added a link to the binding documentation

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: I added an array, but I think we should still test this

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: Changed to use `!` (which was pretty much for these cases)

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: For attributes?

packages/xml-views-validation/test/validators/attributes/unknown-attribute-key-spec.ts
tal-sapan: ok

packages/xml-views-validation/test/validators/attributes/unknown-attribute-key-spec.ts
tal-sapan: I wanted to make sure we don't have unexpected errors

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: There is no logic specifically for attributes currently although namespaces are defined for attribute keys in XML, keeping this logic here for now

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
bd82: It style thing so your pick, I am slightly worried about refactoring and moving things around breaking correctness which is more obvious when there is an explicit if/else

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
bd82: Yeah I just checked this myself, even with newest TypeScript version (3.9.x).
TSC is smart, but not smart enough to understand the previous IF condition as a typeguard
that removes ones of the possible union types.

I think we should handle this with a **cast**, Feels like we are turning a **design time** problem (type inference) into a **runtime problem** (untested branch) so in fact **increasing** the scope of the problem, WDYT?

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,198 @@
...
+  // xmlns attributes are always valid
+  // TODO "xmlns:" should not return true (change when using new function in xml-tools)
+  if (isXMLNamespaceKey(attribute.key)) {
+    return true;
+  }
+
+  const { ns, name } = splitAttributeByNamespace(attribute);
+
+  // The following attributes are valid on all tags
+  // undefined name means all attribute names in the namespace are valid
```
bd82, line 91: > // undefined name means all attribute names in the namespace are valid

Other options (instead of comment for special behavior/semantics)
- name: "*" (assuming * is not a valid name in XML)
- ```
     const ALL = undefined;
     const ALWAYS_VALID_ATTRIBUTES = [{ ns: CUSTOM_DATA_NS, name: ALL }]
  ```

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,198 @@
...
+    // See: https://sapui5.hana.ondemand.com/#/topic/263f6e5a915f430894ee290040e7e220
+    { ns: TEMPLATING_NS, name: "require" },
+    // See: https://sapui5.hana.ondemand.com/#/topic/91f0c3ee6f4d1014b6dd926db0e91070
+    { ns: CUSTOM_DATA_NS, name: undefined },
+  ];
+
+  if (
+    find(
+      ALWAYS_VALID_ATTRIBUTES,
+      (_) => _.ns === ns && (_.name === undefined || _.name === name)
```
bd82, line 107: See comment above to make `_.name === undefined` easier to understand.
Also having _.x === x and _.y === y in the same statement is doubly confusing.

Possible Options:
- Avoid shortcut `_` name here and use a meaningful var name.
- Save the results of the three conditionals in named variables.
- add prefix/suffix to better distinguish `_.x` and `x`
- Or some combination of the above...

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  return false;
+}
+
+function isValidUI5ClassAttribute(
+  ui5Class: UI5Class,
+  attribute: XMLAttribute
+): boolean {
+  // The following attributes are valid on all classes
+  // undefined fqn means the attribute is valid on all classes
```
bd82, line 121: Same as above, try to model "undefined" === "*" in the code instead of in a comment.

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,198 @@
...
+        severity: "error",
+      },
+    ];
+  }
+
+  // No validations on unknown tags
+  return [];
+}
+
+function isAttributeNameAlwaysValid(attribute: XMLAttribute): boolean {
```
bd82, line 76: This function is ~100 lines, could we at a minimum extract the `ALWAYS_VALID_XYZ` constants.
Maybe something else could be extracted as well?

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,198 @@
...
+  // TODO "xmlns:" should not return true (change when using new function in xml-tools)
+  if (isXMLNamespaceKey(attribute.key)) {
+    return true;
+  }
+
+  const { ns, name } = splitAttributeByNamespace(attribute);
+
+  // The following attributes are valid on all tags
+  // undefined name means all attribute names in the namespace are valid
+  const ALWAYS_VALID_ATTRIBUTES = [
```
bd82, line 92: we always search on the `ns` first, would a dictionary be more efficient / appropriate to model this?

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
```diff
@@ -0,0 +1,198 @@
...
+}
+
+function isValidUI5ClassAttribute(
+  ui5Class: UI5Class,
+  attribute: XMLAttribute
+): boolean {
+  // The following attributes are valid on all classes
+  // undefined fqn means the attribute is valid on all classes
+  // TODO Should these be in the model? Should they be suggested for code completion?
+  const ALWAYS_VALID_CLASS_ATTRIBUTES = [
```
bd82, line 123: we always search on the `ns` first, would a dictionary be more efficient / appropriate to model this?

packages/xml-views-validation/test/validators/attributes/unknown-attribute-key-spec.ts
```diff
@@ -0,0 +1,444 @@
...
+        it("will not detect an issue when the attribute is 'class'", () => {
+          assertNoIssues(`
+            <mvc:View
+              xmlns:mvc="sap.ui.core.mvc"
+              class="small">
+            </mvc:View>`);
+        });
+
+        it("will not detect an issue when the attribute namespace is custom data", () => {
+          assertNoIssues(`
```
bd82, line 291: I really like `assertNoIssues` 👍

packages/xml-views-validation/test/validators/attributes/unknown-attribute-key-spec.ts
```diff
@@ -0,0 +1,444 @@
...
+
+describe("the unknown attribute name validation", () => {
+  let ui5SemanticModel: UI5SemanticModel;
+
+  before(async () => {
+    ui5SemanticModel = await generateModel({ version: "1.74.0" });
+  });
+
+  context("true positive scenarios", () => {
+    function assertSingleIssue(xmlSnippet: string, message: string): void {
```
bd82, line 24: we will want to re-use these assertion utils in other specs as well (future upgrade)
Maybe the stack traces will be less useful on errors, but ours tests are getting verbose otherwise...

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: As it should not be offered in code completion I prefer to keep it in the validation, at least for now

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: ok

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: Done

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: Done

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: There are 3 entries in this array, I don't think efficiency it important here. I think It would make the code more complex.

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: Since the error should be the same according to Frank (they give the same error in runtime) I think it's ok to keep it here. If we need to differentiate them in the future we can extract it to a different validation.

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
tal-sapan: Done

packages/xml-views-validation/src/validators/attributes/unknown-attribute-key.ts
bd82: normally I would think these are different errors, but if Frank says it should be the same error its fine.

Afterall UI5 private !== TypeScript private.

packages/xml-views-validation/test/validators/attributes/unknown-attribute-key-spec.ts
bd82: its o.k. now, but if we have 100 special cases we cannot reasonably test all..
- Maybe with generated tests

PR #121, opened 2020-05-14
feat: validation for none unique IDs

bd82:


packages/xml-views-validation/src/validators/attributes/none-unique-id.ts
```diff
@@ -0,0 +1,79 @@
...
+): NoneUniqueIDIssue[] {
+  const idCollector = new IdsCollectorVisitor();
+  accept(xmlDoc, idCollector);
+  const idsToXMLElements = idCollector.idsToXMLElements;
+
+  const duplicatedIdsRecords = omitBy(idsToXMLElements, (_) => _.length > 1);
+  const issues: NoneUniqueIDIssue[] = flatMap(
+    duplicatedIdsRecords,
+    (currDupAttribsPerID, id) => {
+      return map(currDupAttribsPerID, (currDupAttrib, currIdx) => {
```
tal-sapan, line 24: This is a bit complex. Could you add a comment that it maps each duplicate ID attribute to an issue, then flattens the list of issues-list-per-id? Or maybe add a named function `createIssueForDuplicateID(id, allDuplicateIDAttributes, currAttribute, currAttributeIndex)` that returns the issue and call it from here?

packages/xml-views-validation/src/validators/attributes/none-unique-id.ts
```diff
@@ -0,0 +1,79 @@
...
+        const currDuplicatedIdValToken = currDupAttrib.syntax.value;
+        const otherDuplicatedIdsAttribs = clone(currDupAttribsPerID).splice(
+          currIdx
+        );
+
+        return {
+          kind: "NoneUniqueIDIssue",
+          message: `None Unique ID value: "${id}" found.`,
```
tal-sapan, line 32: The term is "non-unique" or "nonunique", not "none-unique". Could you change at least the strings that use the term (the `kind` and `message`)?
https://en.wiktionary.org/wiki/non-unique#English

tal-sapan, line 32: The old error message was: `The control ID is the same as another control ID in the view. Please enter a unique ID.`
Maybe we shouldn't write "control" (it could be an element) and maybe we should drop the "please" but could you change the message to be similar?

packages/language-server/src/xml-view-diagnostics.ts
```diff
@@ -59,9 +56,24 @@ function validationIssuesToLspDiagnostics(
...
+            },
+          })),
+        };
       /* istanbul ignore next - defensive programming */
       default:
-        assertNever(issueKind);
+        // We should use assertNever, However, TSC cannot seem to apply exhaustiveness checks
+        // on inner properties (`currentIssue.kind`) nor does it seem to apply type guards correctly
+        // when we first extract the kind (`const issueKind = currIssue.kind)... before the switch.
+        throw Error("None Exhaustive Match");
```
tal-sapan, line 76: As discussed, use `assertNever` here and cast inside the specific `case`

packages/xml-views-validation/src/validators/attributes/none-unique-id.ts
bd82: I feel the previous message was too long and verbose
and the new message will include "related information" which will add even more lines to the error (verbosity).

- The comment how to fix seems redundant, it should be self evident... e.g see wrong use of const in JetBrains:
  
  image

I'd rather let UX review all these strings at once, while only treating the old implementation as a guideline only.

packages/xml-views-validation/src/validators/attributes/none-unique-id.ts
bd82: refactored to simplify things (new named function to handle a single ID duplications)

packages/xml-views-validation/src/validators/attributes/none-unique-id.ts
bd82: done

packages/language-server/src/xml-view-diagnostics.ts
bd82: done

packages/xml-views-validation/src/validators/attributes/none-unique-id.ts
```diff
@@ -0,0 +1,79 @@
...
+    if (
+      attrib.key === "id" &&
+      attrib.value !== null &&
+      attrib.value !== "" &&
+      attrib.syntax.value !== undefined &&
+      attrib.parent.name !== null &&
+      // Heuristic to limit false positives by only checking tags starting with upper
+      // case names, This would **mostly** limit the checks for things that can actually be
+      // UI5 Elements.
+      /^[A-Z]/.test(attrib.parent.name)
```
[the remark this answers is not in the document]
bd82: done

bd82:
Need to update the snapshot creation to remove or modify the URI
```
   -          "uri": "file:///home/circleci/repo/packages/language-server/test/snapshots/xml-view-diagnostics/non-unique-id/input.xml"
   +          "uri": "file://C:\\workspace\\ui5-language-assistant\\packages\\language-server\\test\\snapshots\\xml-view-diagnostics\\non-unique-id\\input.xml"

File: packages/language-server/test/snapshots/xml-view-diagnostics/non-unique-id/output-lsp-response.json [hunk 5]
```diff
@@ -0,0 +1,44 @@
...
+      "end": { "line": 5, "character": 19 }
+    },
+    "severity": 1,
+    "source": "UI5 Language Assistant",
+    "message": "Non-unique ID value: \"OOPS\" found.",
+    "relatedInformation": [
+      {
+        "message": "identical ID also used here",
+        "location": {
+          "uri": "file://C:\\workspace\\ui5-language-assistant\\packages\\language-server\\test\\snapshots\\xml-view-diagnostics\\non-unique-id\\input.xml",
```
bd82, line 35: TODO: fix the snapshot to avoid including uris of specific machines

packages/xml-views-validation/src/validators/document/non-unique-id.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+}
+
+// We only care about UI5 elements/controls IDs when check non-unique IDs
+// `id` attributes in these: **known** namespaces which are sometimes used
+// in UI5 xml-views are definitively not relevant for this validation
+const whiteListedNamespaces: Record<string, boolean> = {
+  "http://www.w3.org/1999/xhtml": true,
+  "http://www.w3.org/2000/svg": true,
+  "http://schemas.sap.com/sapui5/extension/sap.ui.core.template/1": true,
```
tal-sapan, line 97: Please use the constants from `special-namespaces.ts` (add constants for xhtml and svg), I don't like having copy-paste of these namespaces all over the code

packages/xml-views-validation/src/validators/document/non-unique-id.ts
```diff
@@ -0,0 +1,105 @@
...
+};
+
+function isNoneUI5id(attrib: XMLAttribute): boolean {
+  const parentElement = attrib.parent;
+  const parentPrefix = parentElement.ns ?? DEFAULT_NS;
+  const parentResolvedNamespace = parentElement.namespaces[parentPrefix];
```
tal-sapan, line 103: You can use `resolveXMLNS` from `logic-utils` for this now

packages/xml-views-validation/test/validators/document/non-unique-id-spec.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    testNonUniqueIDScenario = (opts): void =>
+      testValidationsScenario({
+        model: ui5SemanticModel,
+        validators: {
+          document: [validateNonUniqueID],
+        },
+        xmlText: opts.xmlText,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        assertion: opts.assertion as any,
```
tal-sapan, line 30: Why do we need to use `any` here?

packages/xml-views-validation/test/validators/document/non-unique-id-spec.ts
```diff
@@ -0,0 +1,170 @@
...
+  computeExpectedRanges,
+  testValidationsScenario,
+} from "../../test-utils";
+
+describe("the use of non unique id validation", () => {
+  let ui5SemanticModel: UI5SemanticModel;
+
+  let testNonUniqueIDScenario: (opts: {
+    xmlText: string;
+    assertion: (issues: NonUniqueIDIssue[]) => void;
```
tal-sapan, line 16: Maybe calculate the expected ranges and send it to the assertion? It will save some code which is not test-specific (after we have more than 1 test that checks the ranges)

packages/xml-views-validation/test/validators/document/non-unique-id-spec.ts
```diff
@@ -0,0 +1,170 @@
...
+        validators: {
+          document: [validateNonUniqueID],
+        },
+        xmlText: opts.xmlText,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        assertion: opts.assertion as any,
+      });
+  });
+
+  context("true positive scenarios", () => {
```
tal-sapan, line 34: Add a test where there are 3 of the same ID, and a test where there are 2 non-unique IDs in the document

packages/xml-views-validation/test/validators/document/non-unique-id-spec.ts
```diff
@@ -0,0 +1,170 @@
...
+
+      testNonUniqueIDScenario({
+        xmlText: xmlSnippet,
+        assertion: (issues) => {
+          expect(issues).to.be.empty;
+        },
+      });
+    });
+
+    it("will not detect issues for id attributes under low case element tags", () => {
```
tal-sapan, line 132: low case -> lowercase

.eslintrc.js
```diff
@@ -0,0 +1,25 @@
...
+    "prettier",
+  ],
+  rules: {
+    "@typescript-eslint/no-use-before-define": [
+      "error",
+      { functions: false, classes: false },
+    ],
+    // We actually need to ignore multiple things the TypeScript Compiler cannot
+    // seem to figure out by itself.
+    "@typescript-eslint/ban-ts-ignore": "off",
```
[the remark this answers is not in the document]
bd82: force description option in next version of typescript-eslint which is what I want to enable here (in different PR which also updates TypeScript version)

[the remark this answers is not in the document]
tal-sapan: Ok. We should remember to do that.

packages/language-server/src/xml-view-diagnostics.ts
```diff
@@ -44,7 +45,8 @@ function validationIssuesToLspDiagnostics(
       message: currIssue.message,
     };
 
-    switch (currIssue.kind) {
+    const issueKind = currIssue.kind;
+    switch (issueKind) {
```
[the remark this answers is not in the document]
bd82: Tried again, TypeScript is not smart enough in this case.

image

packages/xml-views-validation/src/validators/document/non-unique-id.ts
```diff
@@ -0,0 +1,105 @@
...
+      const currDupIdValToken = currDupAttrib.syntax.value;
+      // Related issues must not include the "main" issue attribute
+      const relatedOtherDupIDAttribs = reject(
+        duplicatedAttributes,
+        (_, arrIdx) => arrIdx === currAttribIdx
+      );
+
+      return {
+        kind: "NonUniqueIDIssue" as "NonUniqueIDIssue",
+        message: `Non-unique ID value: "${id}" found.`,
```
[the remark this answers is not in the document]
bd82: I reduce verbosity and make it clearer by using the pattern:
> Duplicate ID: "XYZ" found.

- I wanted to use the term "ID" and emphasize it (vs identifier)
- Duplicate is a better term in the error message so I change to it as you suggested.

packages/language-server/src/xml-view-diagnostics.ts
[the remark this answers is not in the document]
tal-sapan: Try `assertNever(currIssue)`

packages/xml-views-validation/src/validators/document/non-unique-id.ts
bd82: ok

packages/xml-views-validation/src/validators/document/non-unique-id.ts
bd82: ok

packages/xml-views-validation/test/validators/document/non-unique-id-spec.ts
bd82: image

packages/xml-views-validation/test/validators/document/non-unique-id-spec.ts
bd82: The ranges are used also to assert on `identicalIDsRanges` which is unique to this validation issue. Passing an assertion function and doing calculations is more verbose yet also more versatile/agile for custom flows.

I removed the re-sending of UI5Model / Validation func as those were indeed verbose without cause.

packages/xml-views-validation/test/validators/document/non-unique-id-spec.ts
bd82: ok

packages/xml-views-validation/test/validators/document/non-unique-id-spec.ts
tal-sapan: I think it can probably be solved by defining a generic type for `assertion` (that `extends UI5XMLViewIssue`)

packages/xml-views-validation/test/validators/document/non-unique-id-spec.ts
tal-sapan: I mean to calculate the expected ranges from the xml snippet and send them as a parameter to the `assertion` function (but you will have to call it explicitly instead of passing it to `assertion` of `testValidationsScenario`). Maybe we should even do this generally in `testValidationsScenario`.

packages/language-server/src/xml-view-diagnostics.ts
```diff
@@ -46,7 +46,7 @@ function validationIssuesToLspDiagnostics(
     };
 
     const issueKind = currIssue.kind;
-    switch (issueKind) {
+    switch (currIssue.kind) {
```
tal-sapan, line 49: Can you remove the casting now in line 68?

packages/xml-views-validation/src/validators/document/non-unique-id.ts
```diff
@@ -91,15 +97,13 @@ class IdsCollectorVisitor implements XMLAstVisitor {
 // We only care about UI5 elements/controls IDs when check non-unique IDs
 // `id` attributes in these: **known** namespaces which are sometimes used
 // in UI5 xml-views are definitively not relevant for this validation
-const whiteListedNamespaces: Record<string, boolean> = {
-  "http://www.w3.org/1999/xhtml": true,
-  "http://www.w3.org/2000/svg": true,
-  "http://schemas.sap.com/sapui5/extension/sap.ui.core.template/1": true,
-};
+const whiteListedNamespaces: Record<string, boolean> = {};
```
tal-sapan, line 100: You can initialize it the same way as before (with the constants inside `[]`)
```js
const whiteListedNamespaces: Record<string, boolean> = {
   [SVG_NS]: true,
   ...
};
```

packages/xml-views-validation/src/validators/document/non-unique-id.ts
```diff
@@ -91,15 +97,13 @@ class IdsCollectorVisitor implements XMLAstVisitor {
...
+whiteListedNamespaces[TEMPLATING_NS] = true;
+whiteListedNamespaces[XHTML_NS] = true;
 
 function isNoneUI5id(attrib: XMLAttribute): boolean {
   const parentElement = attrib.parent;
-  const parentPrefix = parentElement.ns ?? DEFAULT_NS;
-  const parentResolvedNamespace = parentElement.namespaces[parentPrefix];
-  return whiteListedNamespaces[parentResolvedNamespace] === true;
+  const parentResolvedNamespace = resolveXMLNS(parentElement) ?? "";
+  return whiteListedNamespaces[parentResolvedNamespace];
```
tal-sapan, line 108: This will return `undefined` instead of `false` if the key doesn't exist

packages/xml-views-validation/test/validators/document/non-unique-id-spec.ts
bd82: Ah, yeah could be useful and reduce some LOC

packages/language-server/src/xml-view-diagnostics.ts
bd82: No, this change above was mistakenly committed(WIP), either the assetNever does not work or casting is needed. So as discussed previously we choose **assertNever** as it provides the bigger value.

packages/xml-views-validation/src/validators/document/non-unique-id.ts
bd82: oh nice.

packages/xml-views-validation/src/validators/document/non-unique-id.ts
bd82: thanks

packages/xml-views-validation/test/validators/document/non-unique-id-spec.ts
bd82: 3 of the same ID is a good use case to test.

2 separate duplicate IDs less so, because then I would be testing the validation framework or even more precisely the ASTVisitor. so I don't think its the responsibility of a specific validation tests to prove they run on all relevant AST Nodes.

Maybe we could expend a smoke test here:
- https://github.com/SAP/ui5-language-assistant/blob/master/packages/xml-views-validation/test/api-spec.ts

- But it already shows 2 separate validations executed on two separate nodes.

bd82:
I think I resolved or replied to all comments.

packages/xml-views-validation/test/validators/document/non-unique-id-spec.ts
bd82: Hmm maybe it would indeed be useful to have multiple IDs test because there is a custom logic here to collect the IDs

packages/language-server/src/xml-view-diagnostics.ts
tal-sapan: It works in `tooltip.ts`. I wonder what's different there.

packages/xml-views-validation/src/utils/special-namespaces.ts
```diff
@@ -4,7 +4,7 @@ export const TEMPLATING_NS =
 /* istanbul ignore next - the constants are used but coverage is not reported for some reason */
```
tal-sapan, line 4: Maybe this comment can also be removed?

packages/xml-views-validation/src/validators/document/non-unique-id.ts
tal-sapan: This would still return `undefined` instead of `false` for non-whitelisted namespaces

packages/xml-views-validation/test/validators/document/non-unique-id-spec.ts
tal-sapan: Also I think there should be a test for duplicate ID in custom control (unknown namespace with tag starting with uppercase)

bd82: > Also I think there should be a test for duplicate ID in custom control (unknown namespace with tag starting with uppercase)

Yes, I'll implement this

PR #123, opened 2020-05-17
fix: use more accurate check for binding expression

tal-sapan:


packages/xml-views-validation/src/utils/is-binding-attribute-value.ts
```diff
@@ -1,7 +1,9 @@
...
-  // The current check is very naive - it only looks for matching curly brackets anywhere in the value.
-  // In the future we might perform a more extensive check, but this is enough for now, so we don't
-  // return false positives in attribute value validations.
-  return /{(.*)}/.test(value);
+  // The current check doesn't diffrenciate between attribute types, although for non-string
+  // attributes the check can be simpler.
+  // Note that the syntax might not be correct (e.g. there is no check for balanced brackets)
+  // but the runtime will still treat it as a binding expression even in that case (it will throw an error).
+  // This will match everything that contains an un-escaped { character.
+  return /^(?:[^\\]|\\.)*{/.test(value);
```
bd82, line 8: could we use negative lookbehind assertion?
- https://2ality.com/2017/05/regexp-lookbehind-assertions.html#negative-lookbehind-assertions

tal-sapan: There is a case we didn't consider - `\\{` should be matched (because the `\` is escaped and not the `{`). This is probably why the match starts at the beginning of the string and checks for matching escaped pairs.

PR #125, opened 2020-05-17
feat(semantic-model): set content as View default aggregation

tal-sapan:


packages/semantic-model/src/fix-api-json.ts
```diff
@@ -0,0 +1,27 @@
...
+export function fixLibrary(libraryName: string, fileContent: unknown): void {
+  forEach(libraryFixes, (fix) => {
+    fix(libraryName, fileContent);
+  });
+}
+
+// Exported for test purpose
+export function addViewDefaultAggregation(
+  libraryName: string,
+  content: unknown
```
bd82, line 15: maybe I forgot something, but why is the content `unknown`? Do not we have a schema/d.ts for the api.json files?

tal-sapan: The fixes are done before the validation that the json matches the schema (in case we want to fix something that isn't in the schema to match the schema)

PR #129, opened 2020-05-18
Validate unknown tag

tal-sapan:
~~**Note:** this is built on top of the unknown attribute PR (#120) because it re-uses some of it. Do not merge until that PR is merged and this PR is rebased.~~

### DIFF
--- a/packages/logic-utils/api.d.ts
+++ b/packages/logic-utils/api.d.ts
@@ -113,3 +113,18 @@ export function getUI5PropertyByXMLAttributeKey(
   attribute: XMLAttribute,
   model: UI5SemanticModel
 ): UI5Prop | undefined;
+
+/**
+ * Return the UI5 Namespace from the specified or default XML Element namespace
+ *
+ * @param xmlElement
+ * @param model
+ */
+export function getUI5NodeFromXMLElementNamespace(
+  xmlElement: XMLElement,
+  model: UI5SemanticModel
+): {
+  namespace: BaseUI5Node | undefined;
+  isDefault: boolean;
+  isXmlnsDefined: boolean;
+};
--- a/packages/logic-utils/src/utils/xml-node-to-ui5-node.ts
+++ b/packages/logic-utils/src/utils/xml-node-to-ui5-node.ts
@@ -3,14 +3,16 @@ import {
   flattenProperties,
   flattenAggregations,
 } from "@ui5-language-assistant/logic-utils";
-import { XMLElement, XMLAttribute } from "@xml-tools/ast";
+import { XMLElement, XMLAttribute, DEFAULT_NS } from "@xml-tools/ast";
 import {
   UI5Class,
   UI5SemanticModel,
   UI5Prop,
   UI5Aggregation,
+  BaseUI5Node,
 } from "@ui5-language-assistant/semantic-model-types";
 import { find } from "lodash";
+import { findSymbol } from "@ui5-language-assistant/semantic-model";
 
 export function getUI5ClassByXMLElement(
   element: XMLElement,
@@ -53,4 +55,30 @@ export function getUI5PropertyByXMLAttributeKey(
   const properties = flattenProperties(elementClass);
   const ui5Property = find(properties, ["name", attribute.key]);
   return ui5Property;
+}
+
+export function getUI5NodeFromXMLElementNamespace(
+  xmlElement: XMLElement,
+  model: UI5SemanticModel
+): {
+  namespace: BaseUI5Node | undefined;
+  isDefault: boolean;
+  isXmlnsDefined: boolean;
+} {
+  const isDefault = xmlElement.ns === undefined;
+  const xmlNamespace = xmlElement.namespaces[xmlElement.ns ?? DEFAULT_NS];
+  if (xmlNamespace === undefined) {
+    return {
+      namespace: undefined,
+      isDefault: isDefault,
+      isXmlnsDefined: false,
+    };
+  }
+
+  const ui5Namespace = findSymbol(model, xmlNamespace);
+  return {
+    namespace: ui5Namespace,
+    isDefault: isDefault,
+    isXmlnsDefined: true,
+  };
 }
--- a/packages/logic-utils/test/utils/xml-node-to-ui5-node-spec.ts
+++ b/packages/logic-utils/test/utils/xml-node-to-ui5-node-spec.ts
@@ -11,6 +11,7 @@ import {
   getUI5AggregationByXMLElement,
   getUI5PropertyByXMLAttributeKey,
   ui5NodeToFQN,
+  getUI5NodeFromXMLElementNamespace,
 } from "../../src/api";
 import { find } from "lodash";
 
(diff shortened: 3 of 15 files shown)


packages/logic-utils/api.d.ts
```diff
@@ -113,3 +113,18 @@ export function getUI5PropertyByXMLAttributeKey(
   attribute: XMLAttribute,
   model: UI5SemanticModel
 ): UI5Prop | undefined;
+
+/**
+ * Return the UI5 Namespace from the specified or default XML Element namespace
```
bd82, line 118: I get the feeling this description needs re-wording

packages/logic-utils/api.d.ts
```diff
@@ -113,3 +113,18 @@ export function getUI5PropertyByXMLAttributeKey(
...
+ *
+ * @param xmlElement
+ * @param model
+ */
+export function getUI5NodeFromXMLElementNamespace(
+  xmlElement: XMLElement,
+  model: UI5SemanticModel
+): {
+  namespace: BaseUI5Node | undefined;
```
bd82, line 127: How can namespace be BaseUI5Node? should it not always be UI5Namespace?
Or is it also possible to return UI5Class (e.g for aggregation tag?)
Basically I am not sure what this utility function does...

packages/logic-utils/src/utils/xml-node-to-ui5-node.ts
```diff
@@ -54,3 +56,29 @@ export function getUI5PropertyByXMLAttributeKey(
...
+  const xmlNamespace = xmlElement.namespaces[xmlElement.ns ?? DEFAULT_NS];
+  if (xmlNamespace === undefined) {
+    return {
+      namespace: undefined,
+      isDefault: isDefault,
+      isXmlnsDefined: false,
+    };
+  }
+
+  const ui5Namespace = findSymbol(model, xmlNamespace);
```
bd82, line 78: why do we need to look for a namespace in the **whole** model instead of only the list of namespaces? does `namespace` in the function mean UI5Namespae or more than that?

packages/logic-utils/test/utils/xml-node-to-ui5-node-spec.ts
```diff
@@ -218,6 +219,128 @@ describe("The @ui5-language-assistant/logic-utils <getUI5PropertyByXMLAttributeK
...
+    } = getUI5NodeFromXMLElementNamespace(rootElement, ui5Model);
+    expect(isDefault).to.be.false;
+    expect(isXmlnsDefined).to.be.true;
+    expectExists(ui5Node, "ui5 namespace");
+    expect(ui5NodeToFQN(ui5Node)).to.equal("sap.ui.core.mvc");
+  });
+
+  it("returns the namespace for tag in a defined namespace when namespace points to an enum", () => {
+    const xmlText = `
+        <mvc:View xmlns:mvc="sap.ui.core.BusyIndicatorSize">
```
bd82, line 247: can a UI5 enum be used as an XML namespace? is this valid?

packages/logic-utils/src/utils/xml-node-to-ui5-node.ts
```diff
@@ -54,3 +56,29 @@ export function getUI5PropertyByXMLAttributeKey(
   const ui5Property = find(properties, ["name", attribute.key]);
   return ui5Property;
 }
+
+export function getUI5NodeFromXMLElementNamespace(
```
bd82, line 60: should this be a **general** logic util or is it too **specific** for a single use case?

packages/xml-views-validation/src/validators/elements/unknown-tag-name.ts
```diff
@@ -0,0 +1,199 @@
+import { UnknownTagNameIssue } from "../../../api";
```
bd82, line 1: import order

packages/xml-views-validation/src/validators/elements/unknown-tag-name.ts
```diff
@@ -0,0 +1,199 @@
...
+  const issueDefaults = {
+    kind: "UnknownTagName" as "UnknownTagName",
+    offsetRange: {
+      start: xmlElement.syntax.openName.startOffset,
+      end: xmlElement.syntax.openName.endOffset,
+    },
+    severity: "error" as "error",
+  };
+
+  // To avoid false positives, we assume unrecognized namespaces are user-defined so we don't validate tags in them
```
bd82, line 60: this comment should be above line 68 or line 69, closer to the condition or the empty return

packages/xml-views-validation/src/validators/elements/unknown-tag-name.ts
```diff
@@ -0,0 +1,199 @@
...
+    return [];
+  }
+
+  return [
+    {
+      ...issueDefaults,
+      message: getMessage(
+        UNKNOWN_CLASS_IN_NS,
+        xmlElement.name,
+        ui5NodeToFQN(ui5Namespace)
```
bd82, line 84: Should we use some "pure" xml utils to extract the Namespace
instead of relying on the FQN of a UI5Node?

Currently this assumes that if `ui5Namespace === undefined` there is no issue.
But that may not be the case in the future (e.g custom controls depending how we model them).

packages/xml-views-validation/src/validators/elements/unknown-tag-name.ts
```diff
@@ -0,0 +1,199 @@
...
+      end: xmlElement.syntax.openName.endOffset,
+    },
+    severity: "error" as "error",
+  };
+
+  // To avoid false positives, we assume unrecognized namespaces are user-defined so we don't validate tags in them
+  // (they could be user-defined class tags).
+  // They could also be valid special namespaces like xhtml.
+  // TODO There should be an error in xml-tools if the namespace is not defined in an xmlns attribute
+  const { namespace: ui5Namespace } = getUI5NodeFromXMLElementNamespace(
```
bd82, line 64: Can we assume everything starting with `sap.[x]` should be valid? or are custom namespaces also using `sap.` prefix?

packages/xml-views-validation/src/validators/elements/unknown-tag-name.ts
```diff
@@ -0,0 +1,199 @@
...
+  name: string;
+  syntax: { openName: XMLToken };
+};
+type NamedXMLElementWithNS = NamedXMLElement & { ns: string };
+
+function validateTagWithNamespace(
+  xmlElement: NamedXMLElementWithNS,
+  model: UI5SemanticModel
+): UnknownTagNameIssue[] {
+  const issueDefaults = {
```
bd82, line 51: you only use this const once, so just in-line it to improve readability (no need to read it before the conditions)

packages/xml-views-validation/src/validators/elements/unknown-tag-name.ts
```diff
@@ -0,0 +1,199 @@
...
+    },
+    severity: "error" as "error",
+  };
+
+  // If the default namespace is defined and it's not a known namespace, don't validate
+  // (it could be a custom class tag or a special namespace like xhtml or template)
+  const {
+    namespace: ui5Namespace,
+    isXmlnsDefined,
+  } = getUI5NodeFromXMLElementNamespace(xmlElement, model);
```
bd82, line 108: `getUI5NodeFromXMLElementNamespace` is confusing for me.
I don't see that we actually need the UI5Node, we can bulid the FQN from the XML AST.
Should we have a simpler utility or set of utilities that check if:
- Extract the XML AstNode namespace and baseName parts (if this does not already exist).
- A matching UI5Namespace **exists** in the model?

bd82: Would this improve readability?

packages/xml-views-validation/src/validators/elements/unknown-tag-name.ts
```diff
@@ -0,0 +1,199 @@
...
+      message: getMessage(
+        UNKNOWN_CLASS_IN_NS,
+        xmlElement.name,
+        ui5NodeToFQN(ui5Namespace)
+      ),
+    },
+  ];
+}
+
+function validateTagWithoutNamespace(
```
bd82, line 90: ~110 lines with significant logical flow, I think many helpers should be extracted to enable
understanding the high level logical flow easily.

packages/xml-views-validation/test/validators/element/unknown-tag-name-spec.ts
```diff
@@ -0,0 +1,525 @@
...
+  let ui5SemanticModel: UI5SemanticModel;
+
+  before(async () => {
+    ui5SemanticModel = await generateModel({ version: "1.74.0" });
+  });
+
+  context("true positive scenarios", () => {
+    let assertSingleIssue: (xmlSnippet: string, message: string) => void;
+    before(() => {
+      assertSingleIssue = partial(
```
bd82, line 27: this "before" does not actually call the productive code or take much time.
So while it does not hurt to have it in the before clause, it also does not help much.

packages/xml-views-validation/test/validators/element/unknown-tag-name-spec.ts
```diff
@@ -0,0 +1,525 @@
...
+    });
+
+    context("tag with namespace", () => {
+      it("will detect an invalid class name in root tag", () => {
+        assertSingleIssue(
+          `<🢂mvc:View_TYPO🢀
+            xmlns:mvc="sap.ui.core.mvc"
+            xmlns="sap.m">
+          </mvc:View_TYPO>`,
+          getMessage(UNKNOWN_CLASS_IN_NS, "View_TYPO", "sap.ui.core.mvc")
```
bd82, line 45: We are not actually testing the error message if we use the same logic to build it as it is built in production.

bd82: And unlike "snapshot tests" no human is visually confirming it.

packages/logic-utils/api.d.ts
tal-sapan: It returns the UI5 symbol associated with the xml element namespace (or the default namespace if the xml element doesn't have a namespace).
There are classes whose parent isn't a `UI5Namespace` but a different symbol. For `Element` subclasses I could only find 1 such example in version 1.71.* (an enum the contains classes under it), but since it exists, I think we should support it.

packages/logic-utils/src/utils/xml-node-to-ui5-node.ts
tal-sapan: There are 2 kinds of "`namespace`s" in the function - the xml namespace and the ui5 "namespace" (which might not be a `UI5Namespace` but a different kind of UI5 symbol as I mentioned above).

packages/logic-utils/test/utils/xml-node-to-ui5-node-spec.ts
tal-sapan: Unfortunately

tal-sapan: It might be a bug but there are several such cases in the model (most classes on a non-namespace "namespace" don't inherit from `Element` but there are some that do)

packages/logic-utils/src/utils/xml-node-to-ui5-node.ts
tal-sapan: I think we will use it for other logic too, like in hover (on xmlns attribute value or element/attribute namespace prefix if we want to implement that)

packages/xml-views-validation/src/validators/elements/unknown-tag-name.ts
tal-sapan: We can use `xmlElement.namespaces[xmlElement.ns]`, do you think it would be better? The FQN should be exactly the same but if we find a different namespace (due to a bug or different logic) I would prefer if the user could see where we actually looked for the class.

packages/xml-views-validation/src/validators/elements/unknown-tag-name.ts
tal-sapan: They can use `sap` prefix (but they shouldn't use any namespaces that are defined in `sap` libraries). Anyway there is already a validation on it - so even if they use an undefined namespace that starts with `sap.ui.core` I wouldn't give an error here.

packages/xml-views-validation/src/validators/elements/unknown-tag-name.ts
tal-sapan: Ok (I used it in several places in the other function so I wanted it to be similar)

packages/xml-views-validation/test/validators/element/unknown-tag-name-spec.ts
tal-sapan: It's in a `before` because the model is built in a `before` so I couldn't call `partial` with it directly inside the `context`

packages/xml-views-validation/test/validators/element/unknown-tag-name-spec.ts
tal-sapan: This is true (we only test the parameters here, assuming they appear in the message string) but the other option is to have the message strings copied in the tests. The snapshot test has all the messages (unless I missed something) so in my opinion it's enough validation for the message strings.

packages/logic-utils/api.d.ts
bd82: I would first inspect if that is an API.json error, if it is invalid api.json
I do not mind not supporting 1/100000 edge case.

Even if it is a valid scenario I would prefer to be more specific about that types:
`namespace: UI5Namespace | UI5Enum`

or even:
`namespaceLike: UI5Namespace | UI5Enum`

packages/logic-utils/src/utils/xml-node-to-ui5-node.ts
bd82: As discussed above we need to understand if that is a valid api.json.
Not sure added complexity is worth it for 0.00001% of cases.

packages/logic-utils/test/utils/xml-node-to-ui5-node-spec.ts
bd82: As discussed above, lets understand how valid this is and how much we want to support it.

packages/logic-utils/src/utils/xml-node-to-ui5-node.ts
bd82: Only if this hierarchy is actually valid in api.json

packages/xml-views-validation/src/validators/elements/unknown-tag-name.ts
bd82: What about: https://github.com/SAP/ui5-language-assistant/blob/master/packages/logic-utils/api.d.ts#L14-L18 ?

packages/xml-views-validation/src/validators/elements/unknown-tag-name.ts
bd82: Its **subjective**, I generally a preference to emphasizing the **logical flow** (what) instead of the **how**.

packages/logic-utils/api.d.ts
tal-sapan: I don't think it's an invalid api.json.
Making the type more specific would make the code more complex because we will have to only search for specific symbol types in the model.
We also have no way to verify which symbol types have sub-symbols in compilation type so we won't know when we have to add types to this `namespaceLike` type.
And while `Element`s are mostly defined on `Namespace`s (except the 3 I found on an enum), there are non-`Element` classes defined on other things like `Class`es.
I think the simple thing is to treat namespaces as being any symbol type. We only use them (as xml namespaces) for resolution of the classes anyway. And if we have other scenarios we can check the `kind`.

tal-sapan: It might make sense though to define a "concrete" UI5 type (namespace/class/enum etc, things that have a sensible FQN) and use that instead of the `BaseUI5Node` which can be things like properties and aggregations.

packages/xml-views-validation/src/validators/elements/unknown-tag-name.ts
tal-sapan: This returns the class FQN, not namespace. If we use that the message will be different (`class <FQN> was not found` instead of `class <name> was not found in <namespace>`).

bd82: ok

PR #130, opened 2020-05-19
fix: remove message for "sap.ui.vk.BaseNodeProxy" on library startup

tal-sapan:
Remove it from its "implements" array when building the model.
Note: I didn't add tests since the strict model building tests already check this (I removed the same fix from the test code).

### DIFF
--- a/packages/semantic-model/src/fix-api-json.ts
+++ b/packages/semantic-model/src/fix-api-json.ts
@@ -1,7 +1,10 @@
-import { forEach, find, get } from "lodash";
+import { forEach, find, get, remove } from "lodash";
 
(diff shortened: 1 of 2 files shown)


packages/semantic-model/src/fix-api-json.ts
```diff
@@ -25,3 +28,16 @@ export function addViewDefaultAggregation(
     }
   }
 }
+
+export function fixSelfImplementingClass(
```
bd82, line 32: This is becoming more and more like the compiler in dts-generator in UI5-TypeScript.
We will need to re-evaluate in the future how to merge these projects.

PR #137, opened 2020-05-20
build(language-server): add missing dependency

tal-sapan:


bd82:
I wonder if there is an eslint rule for mono repo deps

PR #138, opened 2020-05-21
test: add unit tests - node documentation and special properties

sapirpol:


tal-sapan:
I added tests for the deprecated in another PR, but it doesn't go through the `getNodeDocumentation` function. You can keep both tests for deprecation or remove one of them. I think it would be best if you have a test that sends both the version and text for the deprecation, no need for other deprecation message tests.

tal-sapan:
It's ok if you need to return some of the `istanbul ignore` comments for this (I moved this function anyway)

packages/language-server/test/documentation-spec.ts
```diff
@@ -0,0 +1,83 @@
...
+          since: undefined,
+          text: undefined,
+        },
+      });
+      const result = getNodeDocumentation(ui5Enum, ui5SemanticModel);
+      expect(result.value).to.include("Experimental.");
+      expect(result.value).to.not.include("since");
+    });
+
+    it("will get documentation with experimentalInfo - without since", () => {
```
bd82, line 69: I think this test appears twice
`it("will get documentation with experimentalInfo - without since", () => {`

packages/language-server/test/documentation-spec.ts
```diff
@@ -65,19 +57,5 @@ describe("The @ui5-language-assistant/language-server <getNodeDocumentation> fun
       expect(result.value).to.include("Experimental.");
       expect(result.value).to.not.include("since");
     });
```
tal-sapan, line 59: I think the test you removed was supposed to be without `since` and with `text` (or with both of them)

packages/language-server/test/documentation-spec.ts
```diff
@@ -50,12 +52,25 @@ describe("The @ui5-language-assistant/language-server <getNodeDocumentation> fun
         experimentalInfo: {
           isExperimental: true,
           since: undefined,
-          text: undefined,
+          text: "dummyy-text",
```
tal-sapan, line 55: Now there is no test where both properties aren't sent...

PR #141, opened 2020-05-24
fix: use shorter description without jsdoc tags in deprecation warning

tal-sapan:
Move some documentation functions to logic-utils and add tests

### DIFF
--- a/packages/logic-utils/api.d.ts
+++ b/packages/logic-utils/api.d.ts
@@ -136,7 +136,7 @@ export function getUI5NodeFromXMLElementNamespace(
  * @param opts.title - The opening sentance for the deprecation message. It should not contain a dot. "Deprecated" by default.
  */
 export function getDeprecationMessage(opts: {
-  title?: string | undefined;
+  title?: string;
   since: string | undefined;
   text: string | undefined;
 }): string;
--- a/packages/logic-utils/api.d.ts
+++ b/packages/logic-utils/api.d.ts
@@ -9,6 +9,7 @@ import {
   UI5Association,
   UI5SemanticModel,
   UI5Type,
+  UI5DeprecatedInfo,
 } from "@ui5-language-assistant/semantic-model-types";
 
 /**
@@ -128,3 +129,48 @@ export function getUI5NodeFromXMLElementNamespace(
   isDefault: boolean;
   isXmlnsDefined: boolean;
 };
+
+/**
+ * Get the deprecated message. The returned string contains jsdoc tags.
+ *
+ * @param opts.title - The opening sentance for the deprecation message. It should not contain a dot. "Deprecated" by default.
+ */
+export function getDeprecationMessage(opts: {
+  title?: string | undefined;
+  since: string | undefined;
+  text: string | undefined;
+}): string;
+
+/**
+ * Get a snippet (first line) of the deprecated documentation without jsdoc tags.
+ *
+ * @param title - The opening sentance for the deprecation message. It should not contain a dot. "Deprecated" by default.
+ * @param deprecatedInfo
+ * @param model
+ */
+export function getDeprecationPlainTextSnippet(
+  title: string | undefined,
+  deprecatedInfo: UI5DeprecatedInfo,
+  model: UI5SemanticModel
+): string;
+
+/**
+ * Convert jsdoc description to markdown format string
+ * @param jsdocDescription
+ * @param model
+ */
+export function convertJSDocToMarkdown(
+  jsdocDescription: string,
+  model: UI5SemanticModel
+): string;
+
+/**
+ * Get a link according to the link text.
+ * Supported links:
+ * - http/https links - returned as-is
+ * - Other strings are considered to be UI5 FQNs. A link to the relevant SDK page is returned.
+ *
+ * @param model
+ * @param link
+ */
+export function getLink(model: UI5SemanticModel, link: string): string;
--- a/packages/logic-utils/api.d.ts
+++ b/packages/logic-utils/api.d.ts
@@ -133,7 +133,7 @@ export function getUI5NodeFromXMLElementNamespace(
 /**
  * Get the deprecated message. The returned string contains jsdoc tags.
  *
- * @param opts.title - The opening sentance for the deprecation message. It should not contain a dot. "Deprecated" by default.
+ * @param opts.title - The opening sentence for the deprecation message. It should not contain a dot. "Deprecated" by default.
  */
 export function getDeprecationMessage(opts: {
   title?: string;
@@ -144,15 +144,13 @@ export function getDeprecationMessage(opts: {
 /**
  * Get a snippet (first line) of the deprecated documentation without jsdoc tags.
  *
- * @param title - The opening sentance for the deprecation message. It should not contain a dot. "Deprecated" by default.
- * @param deprecatedInfo
- * @param model
+ * @param opts.title - The opening sentence for the deprecation message. It should not contain a dot. "Deprecated" by default.
  */
-export function getDeprecationPlainTextSnippet(
-  title: string | undefined,
-  deprecatedInfo: UI5DeprecatedInfo,
-  model: UI5SemanticModel
-): string;
+export function getDeprecationPlainTextSnippet(opts: {
+  title?: string;
+  deprecatedInfo: UI5DeprecatedInfo;
+  model: UI5SemanticModel;
+}): string;
 
 /**
  * Convert jsdoc description to markdown format string
--- a/packages/logic-utils/src/utils/documentation.ts
+++ b/packages/logic-utils/src/utils/documentation.ts
@@ -123,7 +123,7 @@ function convertJSDoc(
     },
     { matcher: /<ul>/g, replacement: { markdown: "", plaintext: "" } },
     { matcher: /<\/ul>/g, replacement: { markdown: "\n\n", plaintext: "\n" } },
-    // TODO should we handled ordered lists (ol)?
+    // TODO should we handle ordered lists (ol)?
 
     // Code value
     {
@@ -158,7 +158,7 @@ function convertJSDoc(
     {
       matcher: /{@link (([^\s}]+)\s)?([^}]+)}/g,
       replacement: {
-        markdown: (all, _, type, text) => {
+        markdown: (all, _, type, text): string => {
           return `[${text}](${getLink(model, type ?? text)})`;
         },
         plaintext: "$3",
--- a/packages/logic-utils/src/utils/documentation.ts
+++ b/packages/logic-utils/src/utils/documentation.ts
@@ -1,0 +1,191 @@
+import {
+  UI5DeprecatedInfo,
+  UI5SemanticModel,
+} from "@ui5-language-assistant/semantic-model-types";
+import { ReplaceFunction, forEach, unescape } from "lodash";
+
+const MAX_SNIPPET_LENGTH = 500;
+
+export function getDeprecationPlainTextSnippet(
+  title: string | undefined,
+  deprecatedInfo: UI5DeprecatedInfo,
+  model: UI5SemanticModel
+): string {
+  let text = deprecatedInfo.text;
+  if (text !== undefined) {
+    text = convertJSDocToPlainText(text, model);
+    // Only take the first sentence.
+    // A sentence is defined by a dot followed by a whitespace character.
+    // TODO Should we allow the result to have more than 1 line? (If so, add "s" flag to the regex)
+    const matches = text.match(/^((?:.*?)\.)\s/);
+    if (matches !== null) {
+      text = matches[1];
+    }
+    // Limit the length of the snippet
+    if (text.length > MAX_SNIPPET_LENGTH) {
+      text = text.substring(0, MAX_SNIPPET_LENGTH) + "...";
+    }
+  }
+
+  const doc = getDeprecationMessage({
+    title,
+    since: deprecatedInfo.since,
+    text,
+  });
+  return doc;
+}
+
+export function getDeprecationMessage({
+  title,
+  since,
+  text,
+}: {
+  title?: string;
+  since: string | undefined;
+  text: string | undefined;
+}): string {
+  let contents = title ?? "Deprecated";
+  contents += since ? ` since version ${since}` : "";
+  contents += ".";
+  contents += text ? ` ${text}` : "";
+  return contents;
+}
+
+function convertJSDocToPlainText(
+  jsdocDescription: string,
+  model: UI5SemanticModel
+): string {
+  return convertJSDoc(jsdocDescription, "plaintext", model);
+}
+
+export function convertJSDocToMarkdown(
+  jsdocDescription: string,
+  model: UI5SemanticModel
+): string {
+  return convertJSDoc(jsdocDescription, "markdown", model);
+}
+
+type ConvertTarget = "markdown" | "plaintext";
+function convertJSDoc(
+  jsdoc: string,
+  target: ConvertTarget,
+  model: UI5SemanticModel
+): string {
+  // The replacements are applied in order
+  const tagMatcherToReplacement: {
+    matcher: RegExp;
+    replacement: Record<ConvertTarget, string | ReplaceFunction>;
+  }[] = [
+    // Italics
+    {
+      matcher: /<i>(.+?)<\/i>/g,
+      replacement: { markdown: "*$1*", plaintext: "$1" },
+    },
+
+    // Bold
+    {
+      matcher: /<b>(.+?)<\/b>/g,
+      replacement: { markdown: "**$1**", plaintext: "$1" },
+    },
+    {
+      matcher: /<strong>(.+?)<\/strong>/g,
+      replacement: { markdown: "**$1**", plaintext: "$1" },
+    },
+
+    // Emphasis
+    {
+      matcher: /<em>(.+?)<\/em>/g,
+      replacement: { markdown: "***$1***", plaintext: "$1" },
+    },
+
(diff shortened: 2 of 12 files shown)


packages/logic-utils/api.d.ts
```diff
@@ -128,3 +129,48 @@ export function getUI5NodeFromXMLElementNamespace(
   isDefault: boolean;
   isXmlnsDefined: boolean;
 };
+
+/**
+ * Get the deprecated message. The returned string contains jsdoc tags.
+ *
+ * @param opts.title - The opening sentance for the deprecation message. It should not contain a dot. "Deprecated" by default.
```
bd82, line 136: `sentance` -> sent**e**nce

packages/logic-utils/api.d.ts
```diff
@@ -128,3 +129,48 @@ export function getUI5NodeFromXMLElementNamespace(
...
+export function getDeprecationMessage(opts: {
+  title?: string;
+  since: string | undefined;
+  text: string | undefined;
+}): string;
+
+/**
+ * Get a snippet (first line) of the deprecated documentation without jsdoc tags.
+ *
+ * @param title - The opening sentance for the deprecation message. It should not contain a dot. "Deprecated" by default.
```
bd82, line 147: `sentance` -> sent**e**nce

packages/logic-utils/api.d.ts
```diff
@@ -128,3 +129,48 @@ export function getUI5NodeFromXMLElementNamespace(
...
+ * @param deprecatedInfo
+ * @param model
+ */
+export function getDeprecationPlainTextSnippet(
```
bd82, line 151: why is `getDeprecationMessage` using **config object pattern** with optional property.
while getDeprecationPlainTextSnippet is using **ordered args** with property which may have an undefined value?

I think we should be consistent and also that with 3 or more args config object pattern should be the default for **PUBLIC** apis.

bd82: Otherwise likely more breaking changes would occure.

packages/logic-utils/api.d.ts
```diff
@@ -128,3 +129,48 @@ export function getUI5NodeFromXMLElementNamespace(
...
+  deprecatedInfo: UI5DeprecatedInfo,
+  model: UI5SemanticModel
+): string;
+
+/**
+ * Convert jsdoc description to markdown format string
+ * @param jsdocDescription
+ * @param model
+ */
+export function convertJSDocToMarkdown(
```
bd82, line 162: completely subjective:
- config object pattern vs ordered args.
- normally over kill for 2 args, but maybe not when it is a public api to avoid future breaking changes?

Your choice.

packages/logic-utils/src/utils/documentation.ts
```diff
@@ -0,0 +1,191 @@
+import {
+  UI5DeprecatedInfo,
+  UI5SemanticModel,
+} from "@ui5-language-assistant/semantic-model-types";
+import { ReplaceFunction, forEach, unescape } from "lodash";
+
+const MAX_SNIPPET_LENGTH = 500;
```
bd82, line 7: is a 400+ characters sentence readable in the problems view? should this limit be shorter?

packages/logic-utils/src/utils/documentation.ts
```diff
@@ -0,0 +1,191 @@
...
+    // Only take the first sentence.
+    // A sentence is defined by a dot followed by a whitespace character.
+    // TODO Should we allow the result to have more than 1 line? (If so, add "s" flag to the regex)
+    const matches = text.match(/^((?:.*?)\.)\s/);
+    if (matches !== null) {
+      text = matches[1];
+    }
+    // Limit the length of the snippet
+    if (text.length > MAX_SNIPPET_LENGTH) {
+      text = text.substring(0, MAX_SNIPPET_LENGTH) + "...";
```
bd82, line 26: The "..." dots are not that usable when there is no way to open the link or expand them for the full text, but I do not have an alternative to that :(

packages/logic-utils/src/utils/documentation.ts
```diff
@@ -0,0 +1,191 @@
...
+  return doc;
+}
+
+export function getDeprecationMessage({
+  title,
+  since,
+  text,
+}: {
+  title?: string;
+  since: string | undefined;
```
bd82, line 44: if we mark the type as optional `since?` would passing undefined to it raise a compilation error?

packages/logic-utils/api.d.ts
tal-sapan: Ok. It seemed more important in `getDeprecationMessage` because that function receives 3 string arguments and here the arguments have different types so you couldn't accidentally send  the wrong argument as long as you have type checks.

packages/logic-utils/src/utils/documentation.ts
tal-sapan: I couldn't find an example this long, do you have something long I can check it on?

packages/logic-utils/src/utils/documentation.ts
tal-sapan: Since we have hover now, you will either see it in the description (if you hover on the problem marker) or if you see the message in the problem view, click on it and hover over the message. (That's why I really wanted to have the hover before we did this change.)

packages/logic-utils/src/utils/documentation.ts
tal-sapan: No

packages/logic-utils/src/utils/documentation.ts
```diff
@@ -0,0 +1,191 @@
...
+        plaintext: "$3",
+      },
+    },
+  ];
+
+  let contents = jsdoc;
+  forEach(tagMatcherToReplacement, (_) => {
+    const replacement = _.replacement[target];
+    // Workaround for typescript signature definition
+    if (typeof replacement === "string") {
```
bd82, line 173: 174 and 176 lines seem to do that same thing.
If there is a problem with string.prototype.definitions maybe we should solve it 
at design time with a cast instead of adding a runtime branch?

packages/logic-utils/src/utils/documentation.ts
```diff
@@ -0,0 +1,191 @@
...
+    } else {
+      contents = contents.replace(_.matcher, replacement);
+    }
+  });
+
+  return contents;
+}
+
+export function getLink(model: UI5SemanticModel, link: string): string {
+  if (link.startsWith("http:") || link.startsWith("https:")) {
```
bd82, line 184: Should we convert all http links to http**s** links? is this safe?

packages/logic-utils/src/utils/documentation.ts
```diff
@@ -0,0 +1,191 @@
...
+}
+
+type ConvertTarget = "markdown" | "plaintext";
+function convertJSDoc(
+  jsdoc: string,
+  target: ConvertTarget,
+  model: UI5SemanticModel
+): string {
+  // The replacements are applied in order
+  const tagMatcherToReplacement: {
```
bd82, line 75: Maybe subjective, but is there a value to defining the model outside the function?

- I tend to prefer separation of model and logic.
- In this case the model is semi complex, would all these regexps be initialized everytime the `convertJSDoc` function is called?

test-packages/test-utils/src/utils/semantic-model-builder.ts
```diff
@@ -100,7 +100,9 @@ export function buildUI5Association(
   };
 }
 
-export function buildUI5Class(opts: PartialWithName<UI5Class>): UI5Class {
+export function buildUI5Class<T extends PartialWithName<UI5Class>>(
+  opts: T
+): UI5Class & T {
```
bd82, line 105: This is great for managing the optional vs known to exist props, should it be adjusted for all the builder functions?

bd82: Hmm, but what if we pass something that is not a UI5Class prop?
Should T be constrained to properties that exist in UI5Class?

packages/logic-utils/src/utils/documentation.ts
tal-sapan: The problem is that `replace` has 2 overloads: `(RegExp, string)` and `(RegExp, ReplaceFunction)`, instead of having the second argument as `string | ReplaceFunction`. Casting a `string | ReplaceFunction` to either one of them seems really wrong...

packages/logic-utils/test/utils/documentation-spec.ts
```diff
@@ -0,0 +1,252 @@
+import {
```
bd82, line 1: import order issue, but I think we should just sort out the eslint rule when we have time, so don't bother fixing it now we are wasting time on these things 😄

packages/logic-utils/src/utils/documentation.ts
tal-sapan: I don't want to convert links if they're absolute... definitely not when we don't know where they link to

packages/logic-utils/test/utils/documentation-spec.ts
```diff
@@ -0,0 +1,252 @@
...
+          500,
+          "1234567890"
+        ) +
+        "..."
+    );
+  });
+});
+
+describe("The @ui5-language-assistant/logic-utils <convertJSDocToMarkdown> function", () => {
+  const model = buildUI5Model({});
```
bd82, line 199: its best to put `  const model = buildUI5Model({});` in a `before` clause

Even though its fast, it makes debugging any code that is directly inside the `describe`
difficult as even with a (another) test specified as `.only` this initialization code will be invoked on all tests execution and may force defining more complex breakpoints (e.g: stop only after previous breakpoint was already hit).

I wrote like 10,000+ lines of tests without proper `before` clauses in another project and now I am paying the price 😢

packages/logic-utils/src/utils/documentation.ts
tal-sapan: I'll move it outside for efficiency but it's part of the function logic, at least as long as we don't allow to send anything else (like a custom type instead of `plaintext` and `markdown`) to the function (and if we do then we should define a proper interface with a method for each replacement, but it seems like overkill for the 2 use cases we have now).

test-packages/test-utils/src/utils/semantic-model-builder.ts
tal-sapan: I think so but we can do it when we need it (which we will for the other deprecated validations)

tal-sapan: > Hmm, but what if we pass something that is not a UI5Class prop?
> Should T be constrained to properties that exist in UI5Class?

The type definition will still be correct because it will contain all the sent properties.
I wouldn't do additional effort in this definition for test functions.

packages/logic-utils/api.d.ts
tal-sapan: Done

packages/logic-utils/api.d.ts
tal-sapan: Done

packages/logic-utils/api.d.ts
tal-sapan: Done

packages/logic-utils/src/utils/documentation.ts
tal-sapan: Moved

packages/logic-utils/test/utils/documentation-spec.ts
tal-sapan: Done

test-packages/test-utils/api.d.ts
```diff
@@ -24,7 +24,9 @@ import { FetchResponse } from "@ui5-language-assistant/language-server";
 //	easily build (partial) data structures for tests with mandatory "name" field
 export type PartialWithName<T> = { name: string } & Partial<T>;
 
-export function buildUI5Class(opts: PartialWithName<UI5Class>): UI5Class;
+export function buildUI5Class<T extends PartialWithName<UI5Class>>(
+  opts: T
+): UI5Class & T;
```
bd82, line 29: same questions on T generic type as in the implementation comments

packages/logic-utils/src/utils/documentation.ts
bd82: I don't have a long example, but max line length is often 80 or 120.
You can **hardcode** return a long sentence and see if it is rendered reasonably or not.

packages/logic-utils/api.d.ts
bd82: >  types so you couldn't accidentally send the wrong argument as long as you have type checks.

It is more about the ability to change in the future (e.g make some prop optional, have multiple optional props) without breaking changes

packages/logic-utils/src/utils/documentation.ts
bd82: Oh cool good point, this resolves the problem.

packages/logic-utils/src/utils/documentation.ts
bd82: o.k. but it is still just a types signature problem (design time) so we should not resolve it by changing the runtime.

I suggest using a cast / `// @ts-ignore` comment
- In TypeScript 3.9 we can use `// @ts-expect-error` comment
So when that signature is fixed we will get a compliation error to remove the comment.

- https://www.typescriptlang.org/docs/handbook/release-notes/overview.html#-ts-expect-error-comments

bd82: Can also make a PR to fix it here:
- https://github.com/microsoft/TypeScript/blob/master/src/lib/es5.d.ts#L422-L434
- For OSS contribution goals :)

packages/logic-utils/api.d.ts
tal-sapan: Well, it's not exactly public API that is supposed to be used outside this mono-repo, this package is more for shared utilities between packages in the mono-repo.

packages/logic-utils/src/utils/documentation.ts
bd82: According to ECMAScript specs it should be string | replaceFunc
- http://www.ecma-international.org/ecma-262/#sec-string.prototype.replace

test-packages/test-utils/src/utils/semantic-model-builder.ts
bd82: > I wouldn't invest additional effort in this definition for test functions.

That is valid, but should the test functions at least be **consistent**?
When someone copies and creates a new such utility would they copy from the Class pattern or one of the other patterns?

packages/logic-utils/src/utils/documentation.ts
bd82: Should we have a more consistent modeling of optional props here?
Or is this constrained by what we actually get from the model?

Maybe it does not matter enough?

packages/logic-utils/src/utils/documentation.ts
tal-sapan: The contribution looks like a good idea and there is a good reason for this in the best practice guide (which is coincidentally exactly our use case):
https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#use-union-types
I hope they will accept it.

bd82: Cool, I will update to TypeScript 3.9 and add a `ts-expect-error` (different PR) so once
TypeScript@next is out with your fix, we will get a reminder to fix it in our code.

bd82:
The 2 blocking concerns are:
- consistent test utils signatures (Even if not prefect).
- Choose a max description length by seeing how different lengths render in VSCode (avoid hostile takeover of problems view area...)

bd82:
I will approve, You can merge when you have those issues resolved (in one way or another).

packages/logic-utils/src/utils/documentation.ts
tal-sapan: > o.k. but it is still just a types signature problem (design time) so we should not resolve it by changing the runtime.
> 
> I suggest using a cast / `// @ts-ignore` comment
> 
> * In TypeScript 3.9 we can use `// @ts-expect-error` comment
>   So when that signature is fixed we will get a compilation error to remove the comment.
> * https://www.typescriptlang.org/docs/handbook/release-notes/overview.html#-ts-expect-error-comments

The problem with that is that if the type mismatch would not be because of the current reason (overload defined instead of a union type), and the types actually would not match, we won't get an error.
I'll change type type definition instead, I think it should work.

tal-sapan:
> * Choose a max description length by seeing how different lengths render in VSCode (avoid hostile takeover of problems view area...)

In my computer and in theia it looks like it only ever takes 1 line... could this be a configuration?
If this is the behavior, should we limit it at all? Maybe removing the tags would be enough? (You can see the full description on hover but I don't think it's really bad that it's long there)
image
image

bd82:
In VSCode it can easily take over the problems view space:

image

Note that by default the messages are expended and each one has to be individually collapsed to clear up the space.

Perhaps the description should be at most two lines and at most some short length? maybe even at most two lines each of 100 characters?

tal-sapan:
> In VSCode it can easily take over the problems view space:

The second screenshot I put is from vscode... maybe the problem is line breaks and not the length of the line?

tal-sapan:
Since the problem seems to be the line breaks and not the length of the text I think we should take the first line (until a `\n`) of the deprecated description disregarding sentences or line length.

bd82:
> Since the problem seems to be the line breaks and not the length of the text I think we should take the first line (until a \n) of the deprecated description disregarding sentences or line length.

👍

PR #143, opened 2020-05-24
feat: add settings to include deprecated and experimental APIs

tal-sapan:
In code completion. They are false by default.


tal-sapan:
This PR doesn't include tests.

packages/language-server/src/completion-items.ts
```diff
@@ -27,41 +27,75 @@ import {
...
-    model: model,
-    offset: document.offsetAt(textDocumentPosition.position),
+    model: opts.model,
+    offset: opts.document.offsetAt(opts.textDocumentPosition.position),
     cst: cst as DocumentCstNode,
     ast: ast,
     tokenVector: tokenVector,
   });
 
+  const filteredSuggestions = await filterBySettings(
```
bd82, line 49: naively I would prefer to pass the settings to this function as a param
instead of reading the settings everytime.

This looks like part of the state the language server needs to maintain such as UI5 Version / manifest.json property for stable ID validation ect.

I don't think the functional approach of no-state is good here when we are dealing with potentially slow File System operations.

Nor do I think its the responsibility of the completionProvider to read the settings

packages/language-server/src/completion-items.ts
```diff
@@ -27,41 +27,75 @@ import {
...
-    model,
-    textDocumentPosition.position
+    filteredSuggestions,
+    opts.model,
+    opts.textDocumentPosition
   );
   return completionItems;
 }
 
+async function filterBySettings(
```
bd82, line 62: I believe the actual filtering should happen in the xml-view-completion package.
It is not the responsibility of the language-server package, it is **semantics** of UI5 not
relevant for the LSP response transformations.

packages/language-server/src/server.ts
```diff
@@ -54,7 +78,8 @@ connection.onCompletion(
       const documentUri = textDocumentPosition.textDocument.uri;
       const document = documents.get(documentUri);
       if (document) {
-        return getCompletionItems(model, textDocumentPosition, document);
+        updateDocumentSettings(document.uri);
```
bd82, line 81: Why do we need to `updateDocumentSettings` on every completion request?

packages/language-server/src/server.ts
```diff
@@ -54,7 +78,8 @@ connection.onCompletion(
...
+        return getCompletionItems({ model, textDocumentPosition, document });
```
bd82, line 82: I think the last known settings should be passed as an **argument** to this function

packages/language-server/src/server.ts
```diff
@@ -99,6 +124,37 @@ documents.onDidChangeContent(async (changeEvent) => {
...
+      setGlobalSettings(change.settings.UI5LanguageAssistant);
+    }
+  }
+  // No further actions are required currently during configuration change. In the future we might want to
+  // re-validate the files.
+});
+
+// Only keep settings for open documents
+documents.onDidClose((e) => {
+  clearDocumentSettings(e.document.uri);
```
bd82, line 155: we have a **per document** specific setting here?

packages/language-server/test/completion-items-classes-spec.ts
```diff
@@ -549,4 +549,92 @@ describe("the UI5 language assistant Code Completion Services - classes", () =>
       ],
     });
   });
+
+  it("will not return deprecated suggestions according to settings", async () => {
```
bd82, line 553: I think these tests belong in xml-view-completion
Note we are testing the completions identity/data not the LSP related transformed data.

packages/language-server/test/completion-items-literals-spec.ts
```diff
@@ -56,4 +56,24 @@ describe("the UI5 language assistant Code Completion Services", () => {
 
     expect(suggestionKinds).to.deep.equal([CompletionItemKind.Constant]);
   });
+
+  it("will return true and false when settings don't allow deprecated and experimental APIs", async () => {
```
bd82, line 60: what is it inside it

packages/language-server/test/completion-items-utils.ts
```diff
@@ -23,22 +28,29 @@ export function getTagName(textEdit: TextEdit | undefined): string | undefined {
...
+  ui5SemanticModel: UI5SemanticModel,
+  settings?: Settings
+): Promise<CompletionItem[]> {
   const { document, position } = getXmlSnippetDocument(xmlSnippet);
   const uri: TextDocumentIdentifier = { uri: "uri" };
   const textDocPositionParams: TextDocumentPositionParams = {
     textDocument: uri,
     position: position,
   };
+  clearSettings();
```
bd82, line 42: having to `clearSettings();` in our test utils may be a warning flag on how we have implemented the state management in the production code.

packages/settings/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@ui5-language-assistant/settings",
+  "version": "0.0.1",
```
bd82, line 3: should initial version 0.1.0 or 0.0.1?

packages/vscode-ui5-language-assistant/test/suite/extension-spec.ts
```diff
@@ -78,7 +78,7 @@ describe("the Language Server Client Integration Tests", () => {
       "unified",
       "upload",
       "util",
-      "ux3",
+      "uxap",
```
bd82, line 81: why did this change? it does not look like a filter has been applied?

packages/vscode-ui5-language-assistant/package.json
```diff
@@ -24,7 +24,19 @@
             "verbose"
           ],
           "default": "off",
-          "description": "Traces the communication between VS Code and the UI5 language server."
+          "description": "Trace the communication between VS Code and the UI5 language server."
+        },
+        "UI5LanguageAssistant.codeAssist.deprecated": {
+          "scope": "resource",
```
bd82, line 30: how does `"scope": "resource"` work if a user wants both global settings and per document settings?

would it be:
> "*" : true
> "foo/bar.view.xml" : false

?

packages/settings/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "lib",
+    "baseUrl": ".",
+    "esModuleInterop": true
```
bd82, line 7: - why do we need esModuleInterop?
- If we need it should it be on the base root tsconfig.json?

packages/settings/src/settings.ts
```diff
@@ -0,0 +1,56 @@
+import deepFreezeStrict from "deep-freeze-strict";
+import { Settings } from "../api";
+
+// These properties are defined (with their default values) in the package.json of the client
+const defaultSettings: Settings = {
+  codeAssist: { deprecated: false, experimental: false },
+};
+deepFreezeStrict(defaultSettings);
+
+// The global settings, used when the `workspace/configuration` request is not supported by the client.
```
bd82, line 10: Does this mean we only support a per document configuration?
Can the user define that for ALL documents they want experimental?

packages/settings/README.md
```diff
@@ -0,0 +1,33 @@
+[![npm (scoped)](https://img.shields.io/npm/v/@ui5-language-assistant/settings.svg)](https://www.npmjs.com/package/@ui5-language-assistant/settings)
```
bd82, line 1: Do we really need a whole settings **package**? (300+LOC) for a few wrappers around a `Map` and some default values?


What is the desired benefit?

packages/language-server/src/server.ts
tal-sapan: There are 2 flows for settings, depending on the client capabilities:
1. The client doesn't support workspace/document-level settings (`workspace/configuration` request). In this case we use global settings (which arrive in `onDidChangeConfiguration`) and don't try to fetch the workspace/document settings.
2. The client supports `workspace/configuration` request. In this case we ask for the document's settings when we need them if we don't have them, and clear all the document settings when the configuration changes (`onDidChangeConfiguration`).
Currently we need the settings only for the code completion, so this is where we do the call . It isn't done on every request, only if we don't have the document settings (this logic is inside `updateDocumentSettings`).

This logic is based on the [language server extension guide in vscode documentation](https://code.visualstudio.com/api/language-extensions/language-server-extension-guide).

tal-sapan: Would it be clearer if I renamed `updateDocumentSettings` to `ensureDocumentSettings`?

packages/language-server/src/server.ts
tal-sapan: Yes. It's taken directly from the implementation guide. The workspace configuration request receives the document URI.
I don't think there is a way to set settings per document in vs code or theia, it's on either user or workspace level, but this is the API.

packages/language-server/test/completion-items-literals-spec.ts
tal-sapan: It's under the `describe`, not `it`

packages/language-server/test/completion-items-utils.ts
tal-sapan: We can clear the settings for the specific document. The state of the settings is global because it's part of a different request, this is how it's managed.

packages/settings/package.json
tal-sapan: I think when we release it will increase the version to 1.0.0 or 0.1.0 (not sure which) anyway

packages/vscode-ui5-language-assistant/test/suite/extension-spec.ts
tal-sapan: The test checks `includes.members` and not `equalsInAnyOrder`. `ux3` was removed from the result. `uxap` was always there, just not tested.
I think there is general room for improvements in the integration tests - instead of testing several completion use cases we can only check one, but we should check other scenarios too like hover and validation (at least as a smoke test), and maybe even a test specific for how settings affect the operations (e.g. check no deprecated APIs are returned by default, then turn the setting on and off and check again).

packages/vscode-ui5-language-assistant/package.json
tal-sapan: It means that it's the lowest possible level to change the setting (I don't think there is UI in vscode or theia for resource-level settings though, so it's probably the same as workspace-level settings here). The settings are managed in the client, and the server asks for the document-level settings (we cache the result in the server).  When the settings change we clear the cache so we know to ask again.

packages/settings/tsconfig.json
tal-sapan: For the `deep-freeze`

packages/settings/src/settings.ts
tal-sapan: It can be defined on user or workspace level
image
On theia it looks like it can be defined per folder or project
image
Whatever the user didn't set is not written to any file and the client falls back to the more general setting.

packages/settings/README.md
tal-sapan: To be able to access it from different packages if necessary, and have the settings structure defined in a single place (except the client package.json, but I think we should add a test to ensure they're in sync).
The language-server obviously needs to access the settings because this is where they are updated. If we want validations or code completion to access it they shouldn't depend on the language-server.

packages/language-server/src/completion-items.ts
tal-sapan: I don't disagree with this, I just wasn't sure if we want to do it in that level because it might be in several places and we'll need to pass the document URI. Though if we do it in `getXMLViewCompletions` it would only be in one place which is reasonable. Should I move it to there?

packages/language-server/src/completion-items.ts
tal-sapan: The settings are read from a cache, not from the client. (It's in a `Thenable` to make sure we don't set the settings for the same document twice at the same time because the first time is still being resolved, not because it has to be re-resolved every time). So it's not a file system operation (although the first time might wait once until the request is completed).

packages/language-server/src/server.ts
tal-sapan: It's one of the options but I'm not sure it's better. I think we'll have to propagate it to too many places (e.g. if we will use settings to control how validations behave) if we don't use a state.

packages/vscode-ui5-language-assistant/package.json
tal-sapan: I added some screenshots in one of the questions below

packages/language-server/test/completion-items-utils.ts
tal-sapan: Removed

packages/settings/README.md
tal-sapan: I added the test

packages/language-server/test/completion-items-literals-spec.ts
tal-sapan: Done

packages/language-server/src/completion-items.ts
tal-sapan: Changed

packages/language-server/src/completion-items.ts
tal-sapan: Changed

packages/language-server/src/server.ts
tal-sapan: Renamed and added a comment

packages/language-server/src/server.ts
tal-sapan: Changed

packages/language-server/test/completion-items-classes-spec.ts
tal-sapan: Moved (though I kept a couple of smoke tests to make sure we actually pass the settings)

packages/settings/tsconfig.json
tal-sapan: I got some compilation errors in other places when I tried to move it to the base tsconfig.json, I'll try to look into it soon

packages/settings/package.json
bd82: we will monitor it

packages/settings/README.md
bd82: this could be resolved by separating the settings types and settings implementation. But its o.k. to start with a single settings package for now.

packages/language-server/src/completion-items.ts
```diff
@@ -27,41 +27,47 @@ import {
...
-  model: UI5SemanticModel,
-  textDocumentPosition: TextDocumentPositionParams,
-  document: TextDocument
-): CompletionItem[] {
-  const documentText = document.getText();
+export function getCompletionItems(opts: {
+  model: UI5SemanticModel;
+  textDocumentPosition: TextDocumentPositionParams;
+  document: TextDocument;
+  documentSettings: Settings;
```
bd82, line 37: is the `Settings` type name too generic?

tal-sapan: It's all the settings of the extension. Do you prefer `UI5LanguageAssistantSettings`? I think it's too long (and it's already imported from `@ui5-language-assistant/settings`)

packages/language-server/test/completion-items-spec.ts
```diff
@@ -832,6 +832,109 @@ describe("the UI5 language assistant Code Completion Services", () => {
...
 
+  context("deprecated", () => {
+    const NO_DEPRECATED_SUGGESTIONS = {
+      codeAssist: { deprecated: false, experimental: true },
+    };
+    const ALLOW_DEPRECATED_SUGGESTIONS = {
+      codeAssist: { deprecated: true, experimental: true },
+    };
+
+    function testDeprecated(xmlSnippet: string, suggestionLabel: string): void {
```
bd82, line 843: if there is only use usage than extracting the test logic to a function can make it harder to follow the flow of a test without reducing LOC.

In our case I think that `testDeprecated` and `testExperimental` can be **combined** to a single utility as they seem mostly duplicated except being passed different "settings" objects.

packages/settings/api.d.ts
```diff
@@ -0,0 +1,31 @@
+export type Settings = CodeAssistSettings & TraceSettings;
```
bd82, line 1: should we give this a less generic name?

packages/settings/src/settings.ts
```diff
@@ -0,0 +1,57 @@
...
+export function clearSettings(): void {
+  documentSettings.clear();
+}
+
+export function clearDocumentSettings(resource: string): void {
+  documentSettings.delete(resource);
+}
+
+export function setGlobalSettings(settings: Settings): void {
+  globalSettings = settings;
```
bd82, line 56: if we want to be completely paranoid we could deepClone the new settings
so nothing external could still have a reference to them and may modify them un-unexpectedly.

packages/settings/test/settings-spec.ts
```diff
@@ -0,0 +1,176 @@
...
+    });
+
+    it("will return the global settings when there is are no settings for the document", async () => {
+      const globalSettings = {
+        codeAssist: { deprecated: true, experimental: false },
+        trace: { server: "off" as const },
+      };
+      setGlobalSettings(globalSettings);
+      const docSettings = await getSettingsForDocument("doc1");
+      expect(docSettings).to.equal(globalSettings);
```
bd82, line 37: `to.deep.equal`?

because we don't care about **reference** equality here, we care about **value** equality.
Having reference equality is an implementation detail that may no longer apply if we add extra "defenses" to our settings module (deepClone).

bd82: The comment above applies to all the tests in this file

packages/settings/test/settings-spec.ts
```diff
@@ -0,0 +1,176 @@
...
+  getDefaultSettings,
+  setGlobalSettings,
+  setSettingsForDocument,
+  hasSettingsForDocument,
+  clearDocumentSettings,
+  clearSettings,
+} from "../src/settings";
+import { expect } from "chai";
+
+context("settings utilities", () => {
```
bd82, line 13: we invested effort in integrating deepFreeze.
Should we also create a test that the (default) settings are immutable?
Maybe also other kinds of settings (per doc/global)

packages/vscode-ui5-language-assistant/package.json
```diff
@@ -58,6 +72,8 @@
     "proxyquire": "2.1.3",
     "vsce": "1.74.0",
     "vscode-languageserver": "6.1.1",
-    "vscode-test": "1.4.0"
+    "vscode-test": "1.4.0",
+    "fs-extra": "9.0.1",
```
bd82, line 76: is not fs-extra also at the root?

packages/vscode-ui5-language-assistant/test/unit/settings-config-spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    // Get the settings from the package.json
+    const packageJsonPath = require.resolve(
+      "vscode-ui5-language-assistant/package.json"
+    );
+    const packageJsonContent = readJsonSync(packageJsonPath);
+    packageJsonSettings =
+      packageJsonContent.contributes.configuration.properties;
+  });
+
+  it("default setting values in package.json have the correct type", () => {
```
bd82, line 18: not important enough to upgrade, but I would dynamically create tests (`it` calls)
one for each property so that a possible failure would have a better error message, e.g:

> property `foo` has default value mismatch between pkg.json and runtime defaults

packages/xml-views-completion/src/api.ts
```diff
@@ -36,7 +37,36 @@ export function getXMLViewCompletions(
       includes(allowedVisibility, _.ui5Node.visibility)
   );
 
-  return publicAndProtectedSuggestions;
+  const filteredSuggestions = filterBySettings(
```
bd82, line 40: `filterBySettings` is implemented using `remove` which **mutates** its argument.
This means that after this statement `filteredSuggestions === publicAndProtectedSuggestions`

So if we add breakpoints during debug it would not be trivial be follow the changes/transformations.

Could we use `filter` or `reject` which create a new array?

packages/xml-views-completion/src/api.ts
```diff
@@ -36,7 +37,36 @@ export function getXMLViewCompletions(
...
+  );
+
+  return filteredSuggestions;
+}
+
+function filterBySettings(
+  suggestions: UI5XMLViewCompletion[],
+  settings: CodeAssistSettings
+): UI5XMLViewCompletion[] {
+  const filteredSuggestions = suggestions;
```
bd82, line 52: If `remove` is better in this function (vs filter/reject), then you can also shallow clone the suggestions before mutating them.

>  const filteredSuggestions = clone(suggestions);

packages/xml-views-completion/test/api-spec.ts
```diff
@@ -104,23 +106,335 @@ describe("The `getXMLViewCompletions()` api", () => {
...
+
+  context("filter deprecated items according to settings", () => {
+    const NO_DEPRECATED_SUGGESTIONS = {
+      codeAssist: { deprecated: false, experimental: true },
+    };
+    const ALLOW_DEPRECATED_SUGGESTIONS = {
+      codeAssist: { deprecated: true, experimental: true },
+    };
+
+    function testDeprecated(
```
bd82, line 118: could should we combine `testDeprecated` and `testExperimental` by passing settings argument to reduce LOC ?

packages/language-server/src/completion-items.ts
bd82: its subjective, the import would be huge but the usage more clear.
Maybe `UI5Settings`?

bd82: or UI5EditorSettings?

packages/vscode-ui5-language-assistant/package.json
tal-sapan: It's a dev dependency on the root

bd82: here too its a devDep

packages/language-server/test/completion-items-spec.ts
tal-sapan: Done

packages/settings/test/settings-spec.ts
tal-sapan: Done

packages/settings/test/settings-spec.ts
tal-sapan: Done for default and global.
It's harder for document because we get a promise. It can be done in the get function but that seems a bit too much (it will probably be called many more times than the set).

packages/settings/src/settings.ts
tal-sapan: Done

packages/vscode-ui5-language-assistant/test/unit/settings-config-spec.ts
tal-sapan: To create an `it` for every setting I'll have to read the `package.json` outside the `before`

packages/xml-views-completion/src/api.ts
tal-sapan: Changed to `reject`

packages/xml-views-completion/test/api-spec.ts
tal-sapan: Done

packages/language-server/src/completion-items.ts
tal-sapan: `UI5LanguageAssitantSettings` looks way too long and I feel like the others aren't accurate... it's the settings for the `UI5LanguageAssitantSettings` vscode extension and language server. Not the editor and not UI5 (`UI5Settings` sounds like the configuration of UI5 inside a project, e.g. the version, which might be part of our settings in the future).
Inside this repository I think it's ok that it's generic because it really is all the settings we have. And it seems unlikely that they will be used from outside this repo.

packages/vscode-ui5-language-assistant/test/unit/settings-config-spec.ts
tal-sapan: I did change the error messages

packages/vscode-ui5-language-assistant/package.json
tal-sapan: Right, I removed it

PR #144, opened 2020-05-24
fix(xml-views-validation): avoid false positive errors for unknown tags

bd82:
If an unknown tag was under a none namespace UI5 entity fqn, e.g the FQN of an enum/interface it
would be treated as a valid known namespace and thus an error would occur.

### DIFF
--- a/packages/xml-views-validation/test/validators/element/unknown-tag-name-spec.ts
+++ b/packages/xml-views-validation/test/validators/element/unknown-tag-name-spec.ts
@@ -310,6 +310,18 @@ describe("the unknown tag name validation", () => {
             </mvc:View>`
           );
         });
+
+        it("will not detect an issue for a class under an none existent namespace which is also a valid UI5 entity FQN", () => {
+          assertNoIssues(
+            `<mvc:View
+              xmlns:mvc="sap.ui.core.mvc"
+              <!-- The error should be on the definition of an invalid xmlns which is not a UI5Namespace rather than on each usage -->
+              xmlns:AvatarColor="sap.m.AvatarColor">
+              <AvatarColor:Green>
+              </AvatarColor:Green>
+            </mvc:View>`
+          );
+        });
       });
     });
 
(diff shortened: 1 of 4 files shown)


bd82:
failure on nodejs 10 is unrelated to this change, Seems master is failing too with some timeout

packages/xml-views-validation/test/validators/element/unknown-tag-name-spec.ts
```diff
@@ -310,6 +310,18 @@ describe("the unknown tag name validation", () => {
...
           );
         });
+
+        it("will not detect an issue for a class under an none existent namespace which is also a valid UI5 entity FQN", () => {
+          assertNoIssues(
+            `<mvc:View
+              xmlns:mvc="sap.ui.core.mvc"
+              <!-- The error should be on the definition of an invalid xmlns which is not a UI5Namespace rather than on each usage -->
+              xmlns:AvatarColor="sap.m.AvatarColor">
+              <AvatarColor:Green>
```
tal-sapan, line 320: Why shouldn't this give an error? `Green` isn't a class (or aggregation), it can't appear in a tag name. An error here is not a false positive. It would fail in runtime.

bd82: If we had `xmlns:bamba="sap.m.Bamba" would we have given an error?`

bd82: Also there is only one error here imho, the invalid namespace.
Every usage of the namespace in a tag should not be an error.

tal-sapan: > If we had `xmlns:bamba="sap.m.Bamba" would we have given an error?`

We would have given a warning on the namespace because we don't know it and it's in an sap library namespace (sap.m). We would not have given an error (or done validations, or code assist) on tags that use this namespace because we consider it to be a custom namespace.

This isn't the case when you put an existing symbol name in the xmlns attribute. That **is** known (so we can't say it's an "unknown namespace"). It cannot be considered a custom namespace (so we cannot treat tags that use it as possible custom controls that should be excluded from validations). And it might have classes inside it which we should validate and suggest for code completion.

bd82:
No capacity to continue with this, closing at this time. It will be tracked in #262

PR #149, opened 2020-05-27
feat: validation for cardinality aggregation

sapirpol:


packages/language-server/src/xml-view-diagnostics.ts
```diff
@@ -56,6 +56,10 @@ function validationIssuesToLspDiagnostics(
         return {
           ...commonDiagnosticPros,
         };
+      case "CardinalityOfAggregation":
```
bd82, line 59: you can simply add this to the list of cases above which "fallthrough" for the base behavior.

packages/xml-views-validation/api.d.ts
```diff
@@ -38,6 +39,10 @@ export interface UseOfDeprecatedClassIssue extends BaseUI5XMLViewIssue {
   kind: "UseOfDeprecatedClass";
 }
 
+export interface CardinalityOfAggregationIssue extends BaseUI5XMLViewIssue {
```
bd82, line 42: `CardinalityOfAggregationIssue` -> `InvalidAggregationCardinalityIssue` ?

packages/xml-views-validation/src/validators/elements/cardinality-of-aggregation.ts
```diff
@@ -0,0 +1,87 @@
+import { filter, forEach, sortBy, map } from "lodash";
+import { XMLElement } from "@xml-tools/ast";
+import { UI5SemanticModel } from "@ui5-language-assistant/semantic-model-types";
+import { getUI5AggregationByXMLElement } from "@ui5-language-assistant/logic-utils";
+import { CardinalityOfAggregationIssue } from "../../../api";
+
+export function validateCardinality(
```
bd82, line 7: `validateAggregationCardinality` ? 

BTW if you will introduce another validation related to cardinality the name would need to be more specific.

packages/xml-views-validation/src/validators/elements/cardinality-of-aggregation.ts
```diff
@@ -0,0 +1,87 @@
...
+  xmlElement: XMLElement,
+  model: UI5SemanticModel
+): CardinalityOfAggregationIssue[] {
+  const aggregation = getUI5AggregationByXMLElement(xmlElement, model);
+  if (aggregation === undefined || xmlElement.syntax.openName === undefined) {
+    return [];
+  }
+
+  const allCousinsElements = getAllICousinsElements(xmlElement);
+  if (allCousinsElements.length > 1 && aggregation.cardinality === "0..1") {
```
bd82, line 17: I would flip this logic to check if there are **NO** issues and explicitly return `[]`
as it would:
- be consistent with how other empty conditions are handled (line 12-14)
- Would not nest the main logic of this function.
- Make (imho) the flow clearer:
  1. Check X --> No problem.
  2. Check Y --> No problem.
  3. Must Be Problem -> Return None Empty Array of issues

packages/xml-views-validation/src/validators/elements/cardinality-of-aggregation.ts
```diff
@@ -0,0 +1,87 @@
...
+    const firstCousin = getFirstCousin(allCousinsElements);
+    // The error will not be shown on first cousin element
+    const currentAggregationSubElements = filter(
```
bd82, line 20: `currentAggregationSubElements` --> `**redundant**AggregationSubElements` ?

packages/xml-views-validation/src/validators/elements/cardinality-of-aggregation.ts
```diff
@@ -0,0 +1,87 @@
...
+      xmlElement.subElements,
+      (_) => !isFirstCousin(_, firstCousin)
+    );
+
+    const cardinalityOfAggregationIssues = map(
+      currentAggregationSubElements,
+      (_): CardinalityOfAggregationIssue => ({
+        kind: "CardinalityOfAggregation",
+        message: `The aggregation '${xmlElement.name}' has cardinality of 0..1 and can only contain one element.`,
```
bd82, line 29: `can only` -> `may only`?

packages/xml-views-validation/src/validators/elements/cardinality-of-aggregation.ts
```diff
@@ -0,0 +1,87 @@
...
+        severity: "error",
+        offsetRange: {
+          start: (_.syntax.openName ?? _.position).startOffset,
```
bd82, line 32: What is the purpose of this  condition `(_.syntax.openName ?? _.position)` here?

packages/xml-views-validation/src/validators/elements/cardinality-of-aggregation.ts
```diff
@@ -0,0 +1,87 @@
...
+    return cardinalityOfAggregationIssues;
+  }
+
+  return [];
+}
+
+function getFirstCousin(allCousinsElements: XMLElement[]): XMLElement {
+  const sortedCousinsElements = sortBy(
+    allCousinsElements,
+    (_) => _.syntax.openName ?? _.position
```
bd82, line 47: handling of tags without open names will add many branches to this code which means many tests, perhaps we should only perform the logic on tags that actually have open names?
What happens from a user's perspective if we don't show an error if there is a redundant empty tag?

packages/xml-views-validation/src/validators/elements/cardinality-of-aggregation.ts
```diff
@@ -0,0 +1,87 @@
...
+  firstCousin: XMLElement
+): boolean {
+  return (
+    (cousinElement.syntax.openName ?? cousinElement.position).startOffset ===
+    (firstCousin.syntax.openName ?? firstCousin.position).startOffset
+  );
+}
+
+/* For example - for the following xml snippet the function will return an array of `Toolbar` elemnt and `tnt:ToolHeader` element
+  `<headerToolbar>
```
bd82, line 64: put this snippet in a code block "```" like in markdown

packages/xml-views-validation/src/validators/elements/cardinality-of-aggregation.ts
```diff
@@ -0,0 +1,87 @@
...
+  cousinElement: XMLElement,
+  firstCousin: XMLElement
+): boolean {
+  return (
+    (cousinElement.syntax.openName ?? cousinElement.position).startOffset ===
+    (firstCousin.syntax.openName ?? firstCousin.position).startOffset
+  );
+}
+
+/* For example - for the following xml snippet the function will return an array of `Toolbar` elemnt and `tnt:ToolHeader` element
```
bd82, line 63: sort out this multi-line comment:
- Document what does cousin means, not just by example
- each line should have a `*` character
- The text does not start on the first `/*` line
- See example: http://www.docjar.net/html/api/java/util/Collections.java.html

packages/xml-views-validation/src/validators/elements/cardinality-of-aggregation.ts
```diff
@@ -0,0 +1,87 @@
...
+  </headerToolbar>
+  <headerToolbar>
+      <tnt:ToolHeader></tnt:ToolHeader>
+  </headerToolbar>`
+*/
+function getAllICousinsElements(xmlElement: XMLElement): XMLElement[] {
+  let allCousinsElements: XMLElement[] = [];
+  if (xmlElement.parent.type === "XMLElement") {
+    const parent = xmlElement.parent;
+    const identicalAggregations = filter(
```
bd82, line 75: its not identical its `sameName`

packages/xml-views-validation/src/validators/elements/cardinality-of-aggregation.ts
```diff
@@ -0,0 +1,87 @@
...
+
+/* For example - for the following xml snippet the function will return an array of `Toolbar` elemnt and `tnt:ToolHeader` element
+  `<headerToolbar>
+      <Toolbar></Toolbar>
+  </headerToolbar>
+  <headerToolbar>
+      <tnt:ToolHeader></tnt:ToolHeader>
+  </headerToolbar>`
+*/
+function getAllICousinsElements(xmlElement: XMLElement): XMLElement[] {
```
bd82, line 71: looking at the code I am no longer sure it is `cousins`

- we start with XMLElement X
- we go to X.parent
- We collect X.subElements.[...]
- So these do not seem like cousins of X.

Maybe we need another term to describe this

packages/xml-views-validation/src/validators/elements/cardinality-of-aggregation.ts
```diff
@@ -0,0 +1,87 @@
...
+  let allCousinsElements: XMLElement[] = [];
+  if (xmlElement.parent.type === "XMLElement") {
+    const parent = xmlElement.parent;
+    const identicalAggregations = filter(
+      parent.subElements,
+      (subElement) =>
+        subElement.name === xmlElement.name && subElement.subElements.length > 0
```
bd82, line 78: for the second condition use `!isEmpty(subElement.subElements)`

isEmpty is part of lodash

packages/xml-views-validation/src/validators/elements/cardinality-of-aggregation.ts
```diff
@@ -0,0 +1,87 @@
...
+    );
+
+    forEach(identicalAggregations, (_) => {
```
bd82, line 81: I think this is a lodash `flatMap`

packages/xml-views-validation/src/validators/elements/cardinality-of-aggregation.ts
bd82: Maybe like `getAllAggregattionElementsAtSameDepth`

and then give more meaningful names to the variables/params, e.g (random thoughts)
xmlElement -> `startAggregation`
subElements -> `subElementsOfContainingClass`

I am sure better names can be found, the point is that if we do
something such as:
- `const parent = xmlElement.parent;`
- `filter( parent.subElements,(subElement) =>`

We do not actually add more information on what these mean in the **current function context**. Normally we don't need to add said information, but when we do none trivial logic it may help.

packages/xml-views-validation/test/validators/element/cardinality-of-aggregation-spec.ts
```diff
@@ -0,0 +1,109 @@
...
+        assertSingleIssueBase,
+        ui5SemanticModel,
+        {
+          element: [validateExplicitAggregationCardinality],
+        },
+        "InvalidAggregationCardinality",
+        "error"
+      );
+    });
+    it("will detect aggregation with cardinality of 0..1 with more than one element", () => {
```
bd82, line 34: add newlines between different test sections for readability

bd82: e.g before the `it`

packages/xml-views-validation/test/validators/element/cardinality-of-aggregation-spec.ts
```diff
@@ -0,0 +1,109 @@
...
+                <m:Toolbar></m:Toolbar>
+                <🢂m:Toolbar🢀></m:Toolbar>
+              </headerToolbar>
+            </m:Panel>
+          </mvc:View>`,
+        getMessage(INVALID_AGGREGATION_CARDINALITY, "headerToolbar")
+      );
+    });
+
+    it("will detect duplicate aggregation with cardinality of 0..1 with at least one element", () => {
```
bd82, line 50: `at least` -> `more than` ?

packages/xml-views-validation/test/validators/element/cardinality-of-aggregation-spec.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  context("negative edge cases", () => {
+    let assertNoIssues: (xmlSnippet: string) => void;
+    before(() => {
+      assertNoIssues = partial(assertNoIssuesBase, ui5SemanticModel, {
+        element: [validateExplicitAggregationCardinality],
+      });
+    });
+
+    it("will not detect an issue when there is an aggregation with only one element", () => {
```
bd82, line 95: look at the validation implementation, there is at least one more situation where we won't have an error which is `0..N` cardinality.

The conditions that return `[]` hint on negative edge cases we need to test:

packages/xml-views-validation/src/validators/elements/cardinality-of-aggregation.ts
```diff
@@ -0,0 +1,113 @@
...
+function getSubElementPosition(
+  xmlElement: XMLElement
+): XMLToken | SourcePosition {
+  return xmlElement.syntax.openName ?? xmlElement.position;
+}
+
+/**
+ *  @param aggregationElem The aggregation from which we will get all its subElements at the same depth.
+ *
+ * For example - for the following xml snippet the function will return an array of `Toolbar` elemnt and `tnt:ToolHeader` element
```
bd82, line 83: also specify what would be the input for the `following snippet`

> Given XMLSnippet `XYZ` and aggregationElem param `ABC`. `Foo` will be returned 

You can use simpler tag names to make the example easier, e.g:
Uncle1, uncle2, cousin1 consin2

PR #150, opened 2020-05-27
fix: support namespace in aggregation name

tal-sapan:
Aggregation tags must have the same namespace as their parent control tag.
This is now supported in code completion, validations and hover.

Additionally, don't filter out the current aggregation during code completion.

### DIFF
--- a/packages/logic-utils/api.d.ts
+++ b/packages/logic-utils/api.d.ts
@@ -172,3 +172,11 @@ export function convertJSDocToMarkdown(
  * @param link
  */
 export function getLink(model: UI5SemanticModel, link: string): string;
+
+/**
+ * Split possibly qualified tag or attribute name to namespace and name
+ * @param qName
+ */
+export function splitQNameByNamespace(
+  qName: string
+): { ns: string | undefined; name: string };
--- a/packages/logic-utils/api.d.ts
+++ b/packages/logic-utils/api.d.ts
@@ -181,3 +181,45 @@ export function getLink(model: UI5SemanticModel, link: string): string;
 export function splitQNameByNamespace(
   qName: string
 ): { prefix: string | undefined; localName: string };
+
+/**
+ * Return the xml namespace defined for the xml element prefix (ns), or undefined if not found
+ * @param xmlElement
+ */
+export function resolveXMLNS(xmlElement: XMLElement): string | undefined;
+
+/**
+ * Return the xml namespace defined for this prefix, or undefined if not found.
+ * The defined namespaces are taken from the xml element.
+ * @param prefix
+ * @param xmlElement
+ */
+export function resolveXMLNSFromPrefix(
+  prefix: string | undefined,
+  xmlElement: XMLElement
+): string | undefined;
+
+/**
+ * Check if the xml element namespace prefixes (ns) reference the same namespace
+ * @param xmlElement1
+ * @param xmlElement2
+ */
+export function isSameXMLNS(
+  xmlElement1: XMLElement,
+  xmlElement2: XMLElement
+): boolean;
+
+/**
+ * Check if the xml namespace prefixes reference the same namespace.
+ * The defined namespaces are taken from the respective xml elements.
+ * @param prefix1
+ * @param xmlElement1
+ * @param prefix2
+ * @param xmlElement2
+ */
+export function isSameXMLNSFromPrefix(
+  prefix1: string | undefined,
+  xmlElement1: XMLElement,
+  prefix2: string | undefined,
+  xmlElement2: XMLElement
+): boolean;
--- a/packages/logic-utils/api.d.ts
+++ b/packages/logic-utils/api.d.ts
@@ -174,9 +174,10 @@ export function convertJSDocToMarkdown(
 export function getLink(model: UI5SemanticModel, link: string): string;
 
 /**
- * Split possibly qualified tag or attribute name to namespace and name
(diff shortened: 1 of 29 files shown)


packages/logic-utils/api.d.ts
```diff
@@ -172,3 +172,11 @@ export function convertJSDocToMarkdown(
  * @param link
  */
 export function getLink(model: UI5SemanticModel, link: string): string;
+
+/**
+ * Split possibly qualified tag or attribute name to namespace and name
+ * @param qName
+ */
+export function splitQNameByNamespace(
```
bd82, line 180: - This possibly belongs in XML-Tools
- `QName` is ambiguous, is this "x.y.z" (regular fqn) or XMLNs fqn "foo:bar"
  There is no way to tell from the docs.
- I know its the xmlns name... so in that case it should use xmlns terms, e.g:
  - ns --> **prefix**

packages/logic-utils/src/utils/split-qname.ts
```diff
@@ -0,0 +1,20 @@
...
+  }
+  const match = qName.match(/(?<ns>[^:]*)(:(?<name>.*))?/);
+  // There will always be a match because qName always contains a colon at this point
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const matchGroups = match!.groups!;
+  return {
+    ns: matchGroups.ns,
+    name:
+      matchGroups.name ??
+      /* istanbul ignore next */
```
bd82, line 17: why is this branch ignored?

bd82:
> Aggregation tags must have the same namespace as their parent control tag.

Do the namespace prefixes need to be **syntactically** equivalent or could they also be resolved to the same namespace? e.g:
- Two xmlns for `sap.m`
- default + explicit ns for `sap.m`

packages/logic-utils/api.d.ts
tal-sapan: > * ~This possibly belongs in XML-Tools~

Actually I agree, but I don't think I'll have time to move it this sprint.

> * `QName` is ambiguous, is this "x.y.z" (regular fqn) or XMLNs fqn "foo:bar"
>   There is no way to tell from the docs.

I'll improve the comment to make it clearer that it's related to xml qualified names.

> * I know its the xmlns name... so in that case it should use xmlns terms, e.g:
>   
>   * ns --> **prefix**

Ok

packages/logic-utils/src/utils/split-qname.ts
tal-sapan: Actually it's unnecessary, I'll remove the `??`

tal-sapan:
> Do the namespace prefixes need to be **syntactically** equivalent or could they also be resolved to the same namespace?

I tested it now and it seems that they can be equivalent instead of the same prefix (non-default and default or both non-default - it works).
The namespace itself must be exactly the same though.
I will fix the logic.

As a side note, I could not get `content` or any other aggregations under `mvc:View` to work with or without a namespace. It always treats them as a class instead of aggregation.

packages/logic-utils/api.d.ts
tal-sapan: > * I know its the xmlns name... so in that case it should use xmlns terms, e.g:
>   
>   * ns --> **prefix**

On the xml element it's called `ns`, even though it's the prefix and not the resolved namespace.

tal-sapan: Changed

packages/logic-utils/src/utils/split-qname.ts
tal-sapan: Changed

tal-sapan:
> Do the namespace prefixes need to be **syntactically** equivalent or could they also be resolved to the same namespace?

Fixed

packages/logic-utils/api.d.ts
bd82: > On the xml element it's called ns, even though it's the prefix and not the resolved namespace.

Yes that is a mistake I've made, but fixing that would be a serious breaking change

bd82:
> As a side note, I could not get content or any other aggregations under mvc:View to work with or without a namespace. It always treats them as a class instead of aggregation.

"Work" in what context?

packages/logic-utils/api.d.ts
```diff
@@ -174,9 +174,52 @@ export function convertJSDocToMarkdown(
...
 export function splitQNameByNamespace(
   qName: string
-): { ns: string | undefined; name: string };
+): { prefix: string | undefined; localName: string };
+
+/**
+ * Return the xml namespace defined for the xml element prefix (ns), or undefined if not found
+ * @param xmlElement
+ */
+export function resolveXMLNS(xmlElement: XMLElement): string | undefined;
```
bd82, line 189: 👍

bd82: These are all very useful utils, we will need to move them to XML-Tools at a later time though

tal-sapan: Agreed

tal-sapan:
> > As a side note, I could not get content or any other aggregations under mvc:View to work with or without a namespace. It always treats them as a class instead of aggregation.
> 
> "Work" in what context?

It doesn't work, as in, I get a runtime error that "content" was not found, and it seems that it searches for it by URL like it's a class.

bd82:
If we have a runtime error we need to open a bug with reproduceable test case

PR #155, opened 2020-06-02
fix: don't allow to use classes from sub-namespace in tag

tal-sapan:


packages/xml-views-tooltip/test/tooltip-spec.ts
```diff
@@ -276,6 +284,15 @@ describe("the UI5 language assistant Hover Tooltip Service", () => {
       expect(ui5Node.kind).equal("UI5Class");
     });
 
+    it("will not get hover content UI5 class with a dot", () => {
```
bd82, line 287: identical test description string appears above.
Mocha uses those strings for uniqueness and it can cause issues

tal-sapan: It's under a different context

packages/logic-utils/test/utils/xml-node-to-ui5-node-spec.ts
```diff
@@ -43,6 +44,16 @@ describe("The @ui5-language-assistant/logic-utils <getUI5ClassByXMLElement> func
     expect(ui5NodeToFQN(ui5Class)).to.equal("sap.ui.core.mvc.View");
   });
 
+  it("returns the class for class in a self-closing tag", () => {
```
bd82, line 47: > class for class 

?

packages/logic-utils/src/utils/xml-to-fqn.ts
```diff
@@ -14,3 +18,22 @@ export function xmlToFQN(astElement: XMLElement): string {
     return baseName;
   }
 }
+
+export function xmlClosingTagToFQN(astElement: XMLElement): string {
+  // TODO: is this the optimal way to handle nameless elements?
```
bd82, line 23: we can add more properties/info in the ASTNodes if needed.

PR #156, opened 2020-06-02
fix: don't offer classes from sub-namespaces in code completion

tal-sapan:
Use utility function to split the element name and namespace prefix.

### DIFF
--- a/packages/xml-views-completion/src/providers/elementName/classes.ts
+++ b/packages/xml-views-completion/src/providers/elementName/classes.ts
@@ -54,20 +54,24 @@ export function classesSuggestions(
     return [];
   }
   const classesMatchingPrefix = filter(classesMatchingType, (_) => {
-    let classNameFilter = ui5NodeToFQN(_);
-    // Check if the namespace prefix matches.
-    // Only classes from the exact same namespace should be returned.
     if (prefixParts.ns !== "") {
-      if (_.parent === undefined || ui5NodeToFQN(_.parent) !== prefixParts.ns) {
-        return false;
-      }
-      // If there is a namespace prefix we only use the class base name to filter the result
-      // (so we don't return results for "core:mvc.View" for example)
-      classNameFilter = _.name;
+      // If the prefix has a namespace (looks like <ns>:<name>),
+      // the namespace part should be matched exactly (only classes from this namespace
+      // should be returned) and the name part should be included in the class base name.
+      // For example, "mvc:View" prefix (where xmlns:mvc="sap.ui.core.mvc")
+      // should match "sap.ui.core.mvc.View" and "sap.ui.core.mvc.HTMLView" classes
+      // and should not match any class from other namespaces (e.g. "sap.ui.core.mvc.subns.View").
+      return (
+        _.parent !== undefined &&
+        ui5NodeToFQN(_.parent) === prefixParts.ns &&
+        includes(_.name, prefixParts.base)
+      );
+    } else {
+      // If there is no namespace in the prefix, the prefix should be included in the
+      // class fully qualified name.
+      // For example, "mvc.View" prefix should match "sap.ui.core.mvc.View" class.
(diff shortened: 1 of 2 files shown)


packages/xml-views-completion/test/providers/elementName/classes-spec.ts
```diff
@@ -360,6 +389,40 @@ describe("The ui5-language-assistant xml-views-completion", () => {
...
+                  xmlText: xmlSnippet,
+                  providers: {
+                    elementName: [classesSuggestions],
+                  },
+                  assertion: (suggestions) => {
+                    assertSuggestionProperties(suggestions, "content");
+                    const suggestionNames = map(suggestions, (_) =>
+                      ui5NodeToFQN(_.ui5Node)
+                    );
+                    expect(suggestionNames).to.deep.equalInAnyOrder([
```
bd82, line 415: I recommend also adding an **explicit** assertion on something that should not be present.
> expect(...).to.not.include.members(["subns-name1"]).

It does not add anything logically, but strengthens the "point" we are trying to prove n this test to a **human** reader.

packages/xml-views-completion/src/providers/elementName/classes.ts
```diff
@@ -52,11 +54,20 @@ export function classesSuggestions(
...
+    // Only classes from the exact same namespace should be returned.
+    if (prefixParts.ns !== "") {
+      if (_.parent === undefined || ui5NodeToFQN(_.parent) !== prefixParts.ns) {
+        return false;
+      }
+      // If there is a namespace prefix we only use the class base name to filter the result
+      // (so we don't return results for "core:mvc.View" for example)
+      classNameFilter = _.name;
+    }
+    // Check if the name matches.
```
bd82, line 68: do the comments in 68-69 add anything?

packages/xml-views-completion/src/providers/elementName/classes.ts
```diff
@@ -52,11 +54,20 @@ export function classesSuggestions(
     return [];
   }
   const classesMatchingPrefix = filter(classesMatchingType, (_) => {
-    const classFqn = ui5NodeToFQN(_);
-    const matchingNamespace = startsWith(classFqn, prefixParts.ns);
-    const classFqnWithoutPrefix = classFqn.substring(prefixParts.ns.length);
-    const matchingBasename = includes(classFqnWithoutPrefix, prefixParts.base);
-    return matchingNamespace && matchingBasename;
+    let classNameFilter = ui5NodeToFQN(_);
```
bd82, line 57: this filter may be slightly more readable if we:
- Avoid mutating classNameFilter
- Have more distinct flows for the FQN filter vs baseName filter.
  - e.g two separate includes statements
- Use **two** variables with clearer names instead of `classNameFilter` e,g:
  - FqnFilter
  - baseNameOnlyFilter.

Basically could the "story"  (partially) via the variable names and the logical flow
instead (less so) then via the comments. in other words more **declarative** code.

Its not a huge thing, its a small function so it can be understood using the comments,
Just a subjective suggestion you may want to evaluate.

packages/xml-views-completion/test/providers/elementName/classes-spec.ts
tal-sapan: Done

packages/xml-views-completion/src/providers/elementName/classes.ts
tal-sapan: Changed

packages/xml-views-completion/src/providers/elementName/classes.ts
tal-sapan: Done

bd82: Looked at the new code.
The combination of in-depth comments with **Examples** and distinct logical branches makes it very easy to understand. 👍

PR #164, opened 2020-06-04
feat: type aggregation validation

sapirpol:


packages/logic-utils/api.d.ts
```diff
@@ -75,6 +75,15 @@ export function findClassesMatchingType({
   model: UI5SemanticModel;
 }): UI5Class[];
 
+/**
+ * Check if a UI5 Class is type of the give `type`.
+ */
+
```
bd82, line 81: redundant ws

packages/xml-views-validation/src/utils/messages.ts
```diff
@@ -11,6 +11,7 @@ export const UNKNOWN_TAG_NAME_IN_NS = `The "{0}" name is neither a class name in
 export const UNKNOWN_TAG_NAME_NO_NS = `The "{0}" name is neither a class name nor an aggregation of its parent tag, please specify a namespace`;
 
 export const INVALID_AGGREGATION_CARDINALITY = `The aggregation "{0}" has cardinality of 0..1 and may only contain one element`;
+export const INVALID_AGGREGATION_TYPE = `The class "{0}" is not matching aggregation type "{1}"`;
```
bd82, line 14: As discussed online.

> The class <Button> is under the aggregation <HeaderBar> and must match the <IBAR> type.

packages/xml-views-validation/src/validators/elements/type-of-aggregation.ts
```diff
@@ -0,0 +1,105 @@
...
+  // Root element can never be inside aggregation
+  if (xmlElement.parent.type === "XMLDocument") {
+    return [];
+  }
+
+  const allowedTypeInAggregation = getAllowedTypeForSubNodesByXMLElement(
+    xmlElement.parent,
+    model
+  );
+  if (ui5class !== undefined && allowedTypeInAggregation !== undefined) {
```
bd82, line 30: flip condition and return empty array if we do not have enough data to compute
and at the last statement return the result of getInvalidAggregationTypeIssue

packages/xml-views-validation/src/validators/elements/type-of-aggregation.ts
```diff
@@ -0,0 +1,105 @@
...
+    return [invalidAggregationTypeIssue];
+  }
+
+  return [];
+}
+
+function getValidAggregationType(
+  aggregationType: UI5Type | undefined
+): UI5Class | UI5Interface | undefined {
+  /* istanbul ignore if */
```
bd82, line 70: if we ignore a branch, lets comment why

bd82: run a nested loop over the entire model and see if such a case exists?

packages/xml-views-validation/test/validators/element/type-of-aggregation-spec.ts
```diff
@@ -0,0 +1,97 @@
...
+    });
+
+    it("will detect mismatch of class to explicit aggregation type", () => {
+      assertSingleIssue(
+        `<mvc:View xmlns:uxap="sap.uxap" xmlns:m="sap.m"
+            xmlns:mvc="sap.ui.core.mvc"
+            xmlns="sap.ui.commons">
+            <m:Panel>
+              <m:headerToolbar>
+                <🢂Toolbar🢀></Toolbar>
```
bd82, line 42: ToolBar does not match the ype of headerToolBa?

bd82: maybe choose something else like a `Button` so its more obvious it should not match?

packages/xml-views-validation/test/validators/element/type-of-aggregation-spec.ts
```diff
@@ -0,0 +1,97 @@
...
+              </m:headerToolbar>
+            </m:Panel>
+          </mvc:View>`,
+        getMessage(INVALID_AGGREGATION_TYPE, "Toolbar", "Toolbar")
+      );
+    });
+
+    it("will detect mismatch of class to explicit aggregation of 'UI5Interface' type", () => {
```
bd82, line 50: Should this description include the term "default aggregation"?
Also I do not think this is "explicit" case

packages/xml-views-validation/test/validators/element/type-of-aggregation-spec.ts
```diff
@@ -0,0 +1,97 @@
...
+              xmlns="sap.ui.commons">
+              <m:Panel>
+                <🢂ToolbarSeparator🢀></ToolbarSeparator>
+              </m:Panel>
+            </mvc:View>`,
+        getMessage(INVALID_AGGREGATION_TYPE, "ToolbarSeparator", "Control")
+      );
+    });
+
+    it("will detect mismatch of class to default aggregation type", () => {
```
bd82, line 63: the description seems inaccurate, this does not look like a default aggregation scenario

packages/xml-views-validation/test/validators/element/type-of-aggregation-spec.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  context("negative edge cases", () => {
+    let assertNoIssues: (xmlSnippet: string) => void;
+    before(() => {
+      assertNoIssues = partial(assertNoIssuesBase, ui5SemanticModel, {
+        element: [validateAggregationType],
+      });
+    });
+
+    it("will not detect an issue when the class is not under aggregation", () => {
```
bd82, line 87: a class is **always** under an aggregation, in the case below it is under the **Default** aggregation of `mvc:View`.

packages/xml-views-validation/test/validators/element/type-of-aggregation-spec.ts
```diff
@@ -0,0 +1,97 @@
+import { partial } from "lodash";
```
bd82, line 1: the pure xml snippets may no be enough to understand why a test should fail or succeed. Basically we do not know from the XML what are the expected and actual types of the aggregation/class.

Perhaps each test should contain an XML Comment in its snippet
to add context to the scenarios

packages/xml-views-validation/src/validators/elements/type-of-aggregation.ts
```diff
@@ -0,0 +1,120 @@
...
+  if (aggregationType === undefined) {
+    return undefined;
+  }
+
+  switch (aggregationType.kind) {
+    case "UI5Class":
+      return aggregationType as UI5Class;
+    case "UI5Interface":
+      return aggregationType as UI5Interface;
+    /* istanbul ignore next - we can only recieve aggregation type of 'UI5Class' or 'UI5Interface' */
```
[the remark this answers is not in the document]
tal-sapan: No. The type might be unresolved (e.g. if there is a typo) or there might be other errors in the api.json (and cases which I'm not even sure are errors, like primitive types, though I think they would be specified in `altTypes` instead). It should be handled in the code.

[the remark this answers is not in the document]
bd82: - Representing unresolved in the model seems fine.
- But assuming an aggregation may not be of enum/primitive types, why should it even be represented in the model?

The model should not represent the API.json verbatim, rather it should represent the **UI5 Semantics**.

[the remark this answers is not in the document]
bd82: I will investigate this outside the scope of this PR

packages/xml-views-validation/test/validators/element/type-of-aggregation-spec.ts
bd82: This is not resolved.
The error message itself is misleading:
- `getMessage(INVALID_AGGREGATION_TYPE, "Toolbar", "Toolbar")`

Can we please pick a more logical scenario to fail? e.g some Button inside headerToolbar

packages/xml-views-validation/test/validators/element/type-of-aggregation-spec.ts
bd82: again this is not resolved, this is not an explicit aggregation case in this snippet
but instead an implicit (default) aggregation of Panel

packages/xml-views-validation/test/validators/element/type-of-aggregation-spec.ts
bd82: The description seems inaccurate to me. I see no defualt aggregation in this snippet

packages/xml-views-validation/test/validators/element/type-of-aggregation-spec.ts
bd82: Again this is unresolved.

The description should be about "no issue when the types match"

packages/xml-views-validation/test/validators/element/type-of-aggregation-spec.ts
```diff
@@ -0,0 +1,97 @@
...
+                  <🢂Toolbar🢀></Toolbar>
+                </m:footer>
+              </m:Page>
+            </mvc:View>`,
+        getMessage(INVALID_AGGREGATION_TYPE, "Toolbar", "IBar")
+      );
+    });
+  });
+
+  context("negative edge cases", () => {
```
bd82, line 94: maybe add some more cases?

- Aggregation under custom class
- Aggregation without type (did we did it in a unit case?)

bd82: Oops looked at wrong commit

PR #170, opened 2020-06-08
feat: deprecated aggregation tag validation

sapirpol:


packages/xml-views-validation/src/utils/deprecated-message-builder.ts
```diff
@@ -27,9 +28,13 @@ export function buildDeprecatedIssueMessage({
       kind = "class";
       name = ui5NodeToFQN(symbol);
       break;
+    case "UI5Aggregation":
+      kind = "aggregation";
+      name = ui5NodeToFQN(symbol);
```
bd82, line 33: > name = ui5NodeToFQN(symbol);

Could be be done **after** the switch case.

packages/xml-views-validation/src/utils/deprecated-message-builder.ts
```diff
@@ -27,9 +28,13 @@ export function buildDeprecatedIssueMessage({
       kind = "class";
       name = ui5NodeToFQN(symbol);
       break;
+    case "UI5Aggregation":
```
bd82, line 31: instead of adding more and more cases over time
we can consider trimming the `UI5` prefix and applying "toLowerCase" on the rest.

packages/xml-views-validation/src/validators/elements/use-of-depracated-aggregation.ts
```diff
@@ -0,0 +1,40 @@
+import { XMLElement } from "@xml-tools/ast";
+import { UI5SemanticModel } from "@ui5-language-assistant/semantic-model-types";
+import { getUI5AggregationByXMLElement } from "@ui5-language-assistant/logic-utils";
+import { UseOfDeprecatedAggregationIssue } from "../../../api";
+import {
+  buildDeprecatedIssueMessage,
+  DeprecatedUI5Symbol,
+} from "../../utils/deprecated-message-builder";
+
+export function validateUseOfDeprecatedAggregation(
```
bd82, line 10: This may be very very similar to deprecatedClass.
Perhaps they can/should be joined?

packages/xml-views-validation/src/utils/deprecated-message-builder.ts
```diff
@@ -27,9 +28,13 @@ export function buildDeprecatedIssueMessage({
...
+      kind = "aggregation";
+      name = ui5NodeToFQN(symbol);
+      break;
     /* istanbul ignore next - defensive programming */
     default:
-      assertNever(symbol.kind);
+      assertNever(symbol);
```
bd82, line 37: why `symbol.kind` --> `symbol` ?

packages/xml-views-validation/src/utils/deprecated-message-builder.ts
tal-sapan: I think there should be a separation between the technical strings and strings that reach the user

packages/xml-views-validation/src/utils/deprecated-message-builder.ts
tal-sapan, line 33: I don't think we should take the FQN of the aggregation. We only use the name everywhere else.

packages/xml-views-validation/src/utils/deprecated-message-builder.ts
bd82: o.k

PR #182, opened 2020-06-11
refactor: use model.version instead of model

sapirpol:


packages/language-server/src/completion-items.ts
```diff
@@ -50,15 +50,15 @@ export function getCompletionItems(opts: {
...
   const completionItems = transformToLspSuggestions(
     suggestions,
-    opts.model,
+    opts.model.version,
     opts.textDocumentPosition
   );
   return completionItems;
 }
 
 function transformToLspSuggestions(
```
bd82, line 59: I wonder if `transformToLspSuggestions` should accept the model?
I know for certain that the documentation stuff that should not be coupled to the UI5Model.

But could these general LSP transformations require the model in the future?


Perhaps its fine to start with a minimal approach as you've implemented.
If and when we need more "context" in those LSP transforms only then we should consider passing the model.

Also LSP transform should be on Logic (Hover/Content Assist/validation) results data structures.
Not on the UI5 Model, Logic on the model already exists inside those logic packages...

tal-sapan: They documentation stuff **is** coupled to the model, actually. It's used to resolve the links (currently it only uses the version but we will need it for more things like links to method/other members when we support them, and to improve the links to types - only add a link if the type exists). It can be optional if we want to ignore links in certain cases or maybe the documentation can accept a function for link resolving instead of passing the model if there is a real reason not to pass it.

bd82: The one scenario where the model would be needed according to the response above seems to be:
- for validating a link to another entity **exists**?

Is this correct?

bd82:
The reason for this change is to reduce implicit coupling between the docs handling and the model. The main question of course is if this coupling is needed or not? (discussed in the comment above).

packages/language-server/src/completion-items.ts
tal-sapan: No, it's to resolve the link URL.
The use cases I already know of:
1. For jsdoc that looks like `{@link name}` where `name` is the fully qualified name of a symbol, the link should be resolved to the symbol URL (e.g.: https://sapui5.hana.ondemand.com/#/api/sap.ui.core.mvc.View). (This is the only use case we handle right now and we don't make sure we know what `name` is.)
1. For jsdoc that looks like `{@link name}` where `name` is the name of a method on the current class, the link should be resolved to the method URL (e.g.: https://sapui5.hana.ondemand.com/#/api/sap.ui.core.mvc.View%23methods/byId). (Note: this will require adding a parameter that will hole the current symbol.)
1. For jsdoc that looks like `{@link name}` where `name` is the name of an event, property, aggregation or special setting on the current class, the link should be resolved to the corresponding section URL in that class (e.g.: https://sapui5.hana.ondemand.com/#/api/sap.ui.core.mvc.View%23events). (Note: this will also require the current symbol.)
1. For jsdoc that looks like `{@link name}` where we don't know what `name` is, there shouldn't be a link at all.

Additionally, we should put the version in the link URL like we already do.

So - either we send the model and current symbol to the documentation function, or we send a "link resolver" function (which we will have to send from both the code completion and hover). Sending the `model.version` will not make an improvement in either direction.

tal-sapan: Also there are at least 2 use cases that don't require the type information from the model:
1. Links that start with `http` or `https` should be returned as-is (we already handle this)
2. Links to the documentation (`topic:123...`) that should be resolved to the docs (e.g. https://sapui5.hana.ondemand.com/1.71.20/#/topic/91f28be26f4d1014b6dd926db0e91070). This requires the model version.

bd82: O.k. now I understand the future need for the model data due to resolution logic.

bd82:
The model would likely be required in the future for advanced resolving of JSDocs links.

PR #183, opened 2020-06-14
feat: deprecated attribute validation

sapirpol:


packages/language-server/src/xml-view-diagnostics.ts
```diff
@@ -59,6 +59,7 @@ function validationIssuesToLspDiagnostics(
         };
       case "UseOfDeprecatedClass":
       case "UseOfDeprecatedAggregation":
+      case "UseOfDeprecatedAttribute":
```
bd82, line 62: the name `UseOfDeprecatedAttribute` is less specific than other names we used here.
and also not consistent.

See other names:
```
      case "UseOfDeprecatedClass":
      case "UseOfDeprecatedAggregation":
```

We can consider more specific issueKinds: ["UseOfDeprecatedField", "UseOfDeprecatedEvent", ...]. even though the implementation may be with a **single** validation function.

This could allow the end user to configure more granular severity levels for different issues (in the future, if/when we implement such a feature).

packages/language-server/test/snapshots/xml-view-diagnostics/use-of-deprecated-aggregation/input.xml
```diff
@@ -0,0 +1,8 @@
+<mvc:View xmlns:mvc="sap.ui.core.mvc" xmlns:charts="sap.ca.ui.charts">
```
bd82, line 1: why did we add a test case for deprecated aggregation **tag** in a PR implementing deprecated **attributes** validation?

packages/xml-views-validation/src/validators/attributes/use-of-depracated-attribute.ts
```diff
@@ -0,0 +1,78 @@
...
+  buildDeprecatedIssueMessage,
+  DeprecatedUI5Symbol,
+} from "../../utils/deprecated-message-builder";
+
+export function validateUseOfDeprecatedAttribute(
+  attribute: XMLAttribute,
+  model: UI5SemanticModel
+): UseOfDeprecatedAttributeIssue[] {
+  if (attribute.syntax.key === undefined || attribute.key === null) {
+    // Can't give an error without a position or value
```
bd82, line 26: the comment does not match the code logic.
> or **value**

packages/xml-views-validation/src/validators/attributes/use-of-depracated-attribute.ts
```diff
@@ -0,0 +1,78 @@
...
+    return [];
+  }
+
+  const ui5PropEventAssociation = findUI5PropEventAssociationByXMLAttributeKey(
```
bd82, line 30: > findUI5PropEventAssociationByXMLAttributeKey

Is there not a strange edge case where an aggregation may also appear as a attribute?

packages/xml-views-validation/src/validators/attributes/use-of-depracated-attribute.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  if (
+    ui5PropEventAssociation === undefined ||
+    ui5PropEventAssociation.deprecatedInfo === undefined
+  ) {
+    return [];
+  }
+
+  const deprecatedIssue: UseOfDeprecatedAttributeIssue = {
+    kind: "UseOfDeprecatedAttribute",
```
bd82, line 43: if we will support multiple more specific kinds of issue here.
maybe we should extract the switchcase from `buildDeprecatedIssueMessage`
something like:

```typescript
function getDeprecatedDetails(symbol: UI5DeprecatedSymbol): 
{
  issueKind:"UseOfDeprecatedEvent" | "UseOfDeperecatedProperty" /* ... */,
  symbolName: string,
  symbolKindName: "Property" | "Event" | "Association" /*...*/
}
```

packages/xml-views-validation/src/validators/attributes/use-of-depracated-attribute.ts
```diff
@@ -0,0 +1,78 @@
...
+  const parentClass = getUI5ClassByXMLElement(attribute.parent, model);
+  if (parentClass === undefined) {
+    return undefined;
+  }
+
+  const allProps: (UI5Prop | UI5Event | UI5Association)[] = flattenProperties(
+    parentClass
+  );
+
+  const allEvents = flattenEvents(parentClass);
```
bd82, line 71: I think aggregations + altTypes can be modeled as XML attributes as well

packages/xml-views-validation/test/validators/attributes/use-of-depracated-attribute-spec.ts
```diff
@@ -0,0 +1,164 @@
...
+      );
+
+      assertSingleIssue(
+        `<mvc:View xmlns:m="sap.m" 
+          xmlns:mvc="sap.ui.core.mvc">
+          <m:App 🢂orientationChange🢀="">
+          </m:App>
+        </mvc:View>`,
+        buildDeprecatedIssueMessage({
+          symbol: orientationChangeEvent as DeprecatedUI5Symbol,
```
bd82, line 75: if you use `assertExists` or `expectExists` w(e.g in line 67)
would it remove the need for the cast?

This comments refers to all uses of this pattern in this file.

packages/language-server/test/snapshots/xml-view-diagnostics/use-of-deprecated-attribute/input.xml
```diff
@@ -0,0 +1,9 @@
+<mvc:View xmlns:m="sap.m" 
+  xmlns:mvc="sap.ui.core.mvc">
+  <m:Page 🢂navButtonText🢀="dummy-text">
```
bd82, line 3: can we locate and add here a scenario of an aggregation with `alttypes`a.k.a aggregation as attribute?

bd82: See: https://openui5nightly.hana.ondemand.com/#/api/sap.ui.table.Table/aggregations

image

packages/logic-utils/src/utils/xml-node-to-ui5-node.ts
```diff
@@ -88,6 +92,37 @@ export function getUI5PropertyByXMLAttributeKey(
   return ui5Property;
 }
 
+export function getUI5NodeByXMLAttributeKey(
```
bd82, line 95: we pass the whole attribute here and we really do need the whole attribute (e.g its parent) to implement this.

So perhaps the name should be `getUI5NodeByXMLAttribute`? **without** the `key`

packages/logic-utils/src/utils/xml-node-to-ui5-node.ts
```diff
@@ -88,6 +92,37 @@ export function getUI5PropertyByXMLAttributeKey(
...
+  attribute: XMLAttribute,
+  model: UI5SemanticModel
+): UI5Prop | UI5Event | UI5Association | UI5Aggregation | undefined {
+  const parentElementClass = getUI5ClassByXMLElement(attribute.parent, model);
+  return parentElementClass
```
bd82, line 100: can we have two separate return statements here?
one in case parentElementClass is undefined and one if it is not.

We may be doing too many things in a single expression which actually confused me the first time  I read it.

packages/logic-utils/test/utils/xml-node-to-ui5-node-spec.ts
```diff
@@ -319,6 +320,93 @@ describe("The @ui5-language-assistant/logic-utils <getUI5AggregationByXMLElement
...
+
+  it("returns the event for known attribute", () => {
+    const xmlText = `
+        <mvc:View xmlns:mvc="sap.ui.core.mvc" afterInit="dummy-text"
+        </mvc:View>`;
+    const attribute = getRootElementAttribute(xmlText, "afterInit");
+
+    const event = getUI5NodeByXMLAttributeKey(attribute, ui5Model);
+    expectExists(event, "ui5 event");
+    expect(ui5NodeToFQN(event)).to.equal("sap.ui.core.mvc.View.afterInit");
```
bd82, line 369: maybe add another assertion that proves this is in fact an event?
> expect(event.kind).to.eql("UI5Event")

packages/language-server/test/snapshots/xml-view-diagnostics/use-of-deprecated-attribute/input.xml
bd82: maybe it would be a problem to find such a scenario because it needs to be both deprecated and have an alt-type

packages/xml-views-validation/src/validators/attributes/use-of-depracated-attribute.ts
```diff
@@ -0,0 +1,81 @@
...
+  buildDeprecatedIssueMessage,
+  DeprecatedUI5Symbol,
+} from "../../utils/deprecated-message-builder";
+
+type UseOfDeprecatedAttributeIssue =
+  | UseOfDeprecatedPropertyIssue
+  | UseOfDeprecatedAggregationIssue
+  | UseOfDeprecatedEventIssue
+  | UseOfDeprecatedAssociationIssue;
+type DeprecatedAttributeIssueKind =
```
bd82, line 27: lets try mapped types

packages/xml-views-validation/src/validators/attributes/use-of-depracated-attribute.ts
```diff
@@ -0,0 +1,81 @@
...
+  };
+
+  return [deprecatedIssue];
+}
+
+function getDeprecatedAttributeIssueKind(
+  ui5Node: UI5Prop | UI5Event | UI5Association | UI5Aggregation
+): DeprecatedAttributeIssueKind {
+  const uiNodeKind = ui5Node.kind;
+  switch (uiNodeKind) {
```
bd82, line 68: ui -> ui5

packages/xml-views-validation/test/validators/attributes/use-of-depracated-attribute-spec.ts
```diff
@@ -0,0 +1,192 @@
...
+      const genericTileClass = ui5SemanticModel.classes["sap.m.GenericTile"];
+      const iconAggregation = find(
+        genericTileClass.aggregations,
+        (_) => _.name === "icon"
+      );
+
+      assertSingleIssue(
+        `<mvc:View xmlns:m="sap.m" 
+          xmlns:mvc="sap.ui.core.mvc">
+          <m:GenericTile 🢂icon🢀="">
```
bd82, line 118: try to use a real example with aggregation with altTypes

packages/language-server/test/snapshots/xml-view-diagnostics/use-of-deprecated-attribute/input.xml
tal-sapan: All aggregations can be used as attributes for binding. You don't need to find one with `altType`.

bd82: ah cool

PR #185, opened 2020-06-15
test: add snapshots for validations

sapirpol:


packages/language-server/test/snapshots/xml-view-diagnostics/invalid-type-of-aggregation/input.xml
```diff
@@ -0,0 +1,16 @@
+<mvc:View xmlns:uxap="sap.uxap" xmlns:m="sap.m"
+    xmlns:mvc="sap.ui.core.mvc" 
+    xmlns="sap.ui.commons">
+    <m:Page>
+        <m:footer>
+            <🢂🢂Toolb🢀ar🢀>
```
bd82, line 6: why are there double ranges here? I think you should pick
an example that is **only** wrong type and not **also** deprecated.
(see the error messages below)

packages/language-server/test/snapshots/xml-view-diagnostics/use-of-deprecated-aggregation/input.xml
```diff
@@ -0,0 +1,8 @@
+<mvc:View xmlns:mvc="sap.ui.core.mvc" xmlns:charts="sap.ca.ui.charts">
+    <mvc:content>
+        <🢂charts:BubbleChart🢀>
```
bd82, line 3: is there any agggregation which is deprecated yet the class it is inside of
is not deprecated?

A quick programmatic search on the module should be able to locate one.

PR #192, opened 2020-06-17
feat: set schema configuration for manifest.json

sapirpol:


package.json
```diff
@@ -12,6 +12,7 @@
     ]
   },
   "scripts": {
+    "build:quick": "lerna run compile && lerna run bundle && lerna run package",
```
bd82, line 15: 👍

packages/vscode-ui5-language-assistant/src/extension.ts
```diff
@@ -51,6 +56,34 @@ export async function activate(context: ExtensionContext): Promise<void> {
     clientOptions
   );
 
+  const jsonSchemaConfig:
```
bd82, line 59: The type `const jsonSchemaConfig:
    | WorkspaceConfiguration
    | undefine` is strange.`

The API of .get()` is 
> get<T>(section: string, defaultValue: T): T

And the (expected) configuration is an Array of { fileMatch, url}

I think we should call the get with the relevant generic type:
- https://github.com/microsoft/vscode/blob/0c25f5a737a0c4b0945783588c13cfee11d0ecb1/src/vs/workbench/api/common/jsonValidationExtensionPoint.ts#L12-L15

Maybe this type is exported somewhere by \@vscode/types?

packages/vscode-ui5-language-assistant/src/extension.ts
```diff
@@ -51,6 +56,34 @@ export async function activate(context: ExtensionContext): Promise<void> {
...
+    | WorkspaceConfiguration
+    | undefined = workspace.getConfiguration().get("json.schemas");
+  if (
+    process.env["THEIA_PARENT_PID"] !== undefined &&
+    isArray(jsonSchemaConfig)
+  ) {
+    const providedManifestSchemaConfigs = find(jsonSchemaConfig, (_) => {
+      return some(
```
bd82, line 67: `some` works on an array, however there may only be a **single** fileMatch property (no array).
- https://github.com/microsoft/vscode/blob/0c25f5a737a0c4b0945783588c13cfee11d0ecb1/src/vs/workbench/api/common/jsonValidationExtensionPoint.ts#L12-L15

packages/vscode-ui5-language-assistant/src/extension.ts
bd82, line 59: I think `jsonSchemaConfig` should be renamed to `jsonSchemaConfig**s**`

packages/vscode-ui5-language-assistant/src/extension.ts
```diff
@@ -51,6 +56,34 @@ export async function activate(context: ExtensionContext): Promise<void> {
...
   );
 
+  const jsonSchemaConfig:
+    | WorkspaceConfiguration
+    | undefined = workspace.getConfiguration().get("json.schemas");
+  if (
+    process.env["THEIA_PARENT_PID"] !== undefined &&
+    isArray(jsonSchemaConfig)
+  ) {
+    const providedManifestSchemaConfigs = find(jsonSchemaConfig, (_) => {
```
bd82, line 66: `providedManifestSchemaConfigs` can be renamed to `hasManifestJsonConfig` (boolean) if you
also inspect the result of the find `!== undefined` in this same statement.

We don't care about the result, only its existence or lack thereof.

packages/vscode-ui5-language-assistant/src/extension.ts
```diff
@@ -51,6 +56,34 @@ export async function activate(context: ExtensionContext): Promise<void> {
...
+      const manifestSchemaConfig = {
+        fileMatch: ["manifest.json"],
+        url:
+          "https://cdn.jsdelivr.net/gh/SAP/ui5-language-assistant/packages/vscode-ui5-language-assistant/resources/manifest-schema/rel-1.19/schema/schema.json",
+      };
+
+      jsonSchemaConfig.push(manifestSchemaConfig);
+      workspace
+        .getConfiguration()
+        .update("json.schemas", jsonSchemaConfig, true);
```
bd82, line 83: why do we update the global settings? could if affect something else we did not mean to do? would it be safer to update the workspace settings?

packages/vscode-ui5-language-assistant/src/extension.ts
```diff
@@ -51,6 +56,34 @@ export async function activate(context: ExtensionContext): Promise<void> {
...
+        _.fileMatch,
+        (fileMatchEntry) => fileMatchEntry === "manifest.json"
+      );
+    });
+
+    if (providedManifestSchemaConfigs === undefined) {
+      const manifestSchemaConfig = {
+        fileMatch: ["manifest.json"],
+        url:
+          "https://cdn.jsdelivr.net/gh/SAP/ui5-language-assistant/packages/vscode-ui5-language-assistant/resources/manifest-schema/rel-1.19/schema/schema.json",
```
amiramw, line 77: Is it available in China?

bd82: good question, but its the same URL that is currently supposed to be used in China.
This is not a new feature, rather a workaround.

packages/vscode-ui5-language-assistant/src/extension.ts
bd82, line 59: we should comment why we are doing this hack (thiea bug not support https://code.visualstudio.com/api/references/contribution-points#contributes.jsonValidation)

bd82:
As discussed lets add multiple ways to ensure this workaround would not be abused:
- Add a `comment` / `description` property in the settings indicating this was auto-magically added, so we could more easily remove it in the future.
- codeowners (e.g for related tests).
- test that ensures the URL added is part of a whitelist with description that we must avoid using many different URLS with this workaround.

packages/vscode-ui5-language-assistant/src/registerManifestSchema.ts
```diff
@@ -0,0 +1,44 @@
...
+import { workspace } from "vscode";
+import {
+  getManifestSchemaConfig,
+  IJSONValidationExtensionPoint,
+} from "./manifestSchemaConfig";
+
+/**
+ *  The configuration will be written in workspace.settings
+ */
+export function registerManifestSchema(): void {
```
bd82, line 12: as discussed over voice, a clearer logical flow is needed here.

bd82:
@amiramw 

This workaround may modify (on theia) the setting.json files. (seems to be `.vscode/settings.json`).
Do you believe this could be a problem if the setting.json file is submitted to source control?

Things I can think of:
- When we remove the workaround the modification in the setting.json may **remain**.
  - Unless we add code to remove it as well.
- This workaround will now add this feature (manifest.json support) in VSCode as well.
  - We will add it in the future via the package.json contributions.
  - However settings.json should have higher priority than package.json contributions.
  - This means that we **must** ensure to delete any such schema entries when we enable the proper solution via package.json contributions.
  - This may also mean that we should keep the `deletion` related code for a long-time as state may remain in a user's github repo.

WDYT?

packages/vscode-ui5-language-assistant/src/registerManifestSchema.ts
```diff
@@ -0,0 +1,104 @@
...
+      jsonSchemaConfig,
+      (jsonSchemaConfigEntry: IJSONValidationExtensionPoint) => {
+        if (isArray(jsonSchemaConfigEntry.fileMatch)) {
+          return some(
+            jsonSchemaConfigEntry.fileMatch,
+            (fileMatchEntry) => fileMatchEntry === MANIFEST_FILE_MATCH
+          );
+        }
+
+        return jsonSchemaConfigEntry.fileMatch === MANIFEST_FILE_MATCH;
```
bd82, line 46: You mentioned that the fileMatch property does not work when it is not an array.
So this code of dealing with both cases may be redundant.

However we saw some VSCode API in which it **could** be an single string.
Perhaps it does not matter as this is a temp workaround.
I will try to find more info.

packages/vscode-ui5-language-assistant/src/registerManifestSchema.ts
```diff
@@ -0,0 +1,104 @@
...
+    } else {
+      // No change is needed
+    }
+  }
+}
+
+function isMissingConfig(
+  jsonSchemaConfig: IJSONValidationExtensionPoint[]
+): boolean {
+  const missingConfigEntry =
```
bd82, line 35: `missingConfigEntry` --> `isMissingConfigEntry`

packages/vscode-ui5-language-assistant/src/registerManifestSchema.ts
```diff
@@ -0,0 +1,104 @@
...
+  jsonSchemaConfig: IJSONValidationExtensionPoint[]
+): number {
+  const previousConfigEntryIndex = findIndex(
+    jsonSchemaConfig,
+    (jsonSchemaConfigEntry: IJSONValidationExtensionPoint) => {
+      if (
+        isArray(jsonSchemaConfigEntry.fileMatch) &&
+        jsonSchemaConfigEntry.fileMatch.length === 1
+      ) {
+        return some(
```
bd82, line 71: maybe add a comment that we are looking for `exact` match.
So be (99%) sure we are only changing settings.json if we have in fact added them.

bd82:
I have consulted online with @idoprz He suggested writing our schemaConfig to the **global** settings
so there would be no possibility of changing files that would be submitted to source control.

packages/vscode-ui5-language-assistant/src/registerManifestSchema.ts
```diff
@@ -0,0 +1,104 @@
...
+  const updatedSchemaConfig = clone(jsonSchemaConfig);
+  updatedSchemaConfig[previusConfigEntryIndex] = getManifestSchemaConfigEntry();
+
+  configureSchemaInWorkspaceSettings(updatedSchemaConfig);
+}
+
+function configureSchemaInWorkspaceSettings(
+  updatedSchemaConfig: IJSONValidationExtensionPoint[]
+): void {
+  // We pass 'undefined' for multi-root
```
bd82, line 99: lets pass `true` to write to the global settings instead of the user's workspace.
> - If `true` updates [Global settings](#ConfigurationTarget.Global).

This will simplify state management as the configurations would be written in some root level `.theia` folder outside the user's workspace, so there would be no source control conflicts.

PR #203, opened 2020-06-22
build: break cyclic package deps

bd82:
semantic-model -> test-utils -> semantic-model

### DIFF
--- a/packages/logic-utils/package.json
+++ b/packages/logic-utils/package.json
@@ -13,6 +13,7 @@
   "license": "Apache-2.0",
   "typings": "./api.d.ts",
   "dependencies": {
+    "@ui5-language-assistant/semantic-model-types": "^1.3.0",
     "@xml-tools/ast": "4.2.1",
     "lodash": "4.17.15"
   },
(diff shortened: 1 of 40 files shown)


packages/logic-utils/package.json
```diff
@@ -13,6 +13,7 @@
   "license": "Apache-2.0",
   "typings": "./api.d.ts",
   "dependencies": {
+    "@ui5-language-assistant/semantic-model-types": "^1.3.0",
```
bd82, line 16: dep used in api.d.ts need to be part of the dependencies.

test-packages/test-utils/api.d.ts
```diff
@@ -96,16 +95,18 @@ export type TestModelVersion = "1.60.14" | "1.74.0" | "1.75.0" | "1.71.14";
...
+ * @param opts.strict - Generate the model in strict mode. True by default.
+ * @param opts.modelGenerator - DI for the actual api.json -> UI5Model builder, pass the `generate`
+ *                              function from the @ui5-language-assistant/semantic-model package.
+ *                              This DI is needed to avoid cyclic dependency graph.
+ *                              ... -> semantic-model -> test-utils -> semantic-model -> ...
  */
 export function generateModel(opts: {
   version: TestModelVersion;
   downloadLibs?: boolean;
-  fixLibs?: boolean;
```
bd82, line 107: fixlibs was not used by any consumers nor implemented by the "provider".

PR #224, opened 2020-07-06
feat: manifest.json state management

sapirpol:


packages/language-server/src/server.ts
```diff
@@ -25,15 +27,33 @@ import {
...
 } from "@ui5-language-assistant/settings";
+import { ServerInitializationOptions } from "../api";
+import { getXMLViewDiagnostics } from "./xml-view-diagnostics";
+import { getHoverResponse } from "./hover";
+
+let workspaceFolderUri: string | null;
+let workspaceFolderPath: string | null;
+type absolutePath = string;
+type manifestsData = Record<absolutePath, { isFlexEnabled: boolean }>;
+const manifestData: manifestsData = Object.create(null);
```
bd82, line 38: `manifestsData` name is used for both type and const.
can we differentiate those? e.g:
- type **M**anifestData (upper case M)

packages/language-server/src/server.ts
```diff
@@ -25,15 +27,33 @@ import {
   hasSettingsForDocument,
   getSettingsForDocument,
 } from "@ui5-language-assistant/settings";
+import { ServerInitializationOptions } from "../api";
+import { getXMLViewDiagnostics } from "./xml-view-diagnostics";
+import { getHoverResponse } from "./hover";
+
+let workspaceFolderUri: string | null;
+let workspaceFolderPath: string | null;
+type absolutePath = string;
```
bd82, line 36: should we use upper case for type names?

packages/language-server/src/server.ts
```diff
@@ -25,15 +27,33 @@ import {
   hasSettingsForDocument,
   getSettingsForDocument,
 } from "@ui5-language-assistant/settings";
+import { ServerInitializationOptions } from "../api";
+import { getXMLViewDiagnostics } from "./xml-view-diagnostics";
+import { getHoverResponse } from "./hover";
+
+let workspaceFolderUri: string | null;
+let workspaceFolderPath: string | null;
```
bd82, line 35: do we need both these workspace related variables as `let`?

packages/language-server/src/server.ts
```diff
@@ -174,3 +225,60 @@ documents.onDidClose((e) => {
...
+  );
+
+  if (requiredManifestPath === undefined) {
+    return false;
+  }
+
+  return manifestData[requiredManifestPath].isFlexEnabled;
+}
+
+async function getManifestDocumentsPromise(): Promise<string[]> {
```
bd82, line 249: can `workspaceFolderPath` be passed here as an argument?
in that case we won't need it as moudle level state

packages/language-server/src/server.ts
```diff
@@ -174,3 +225,60 @@ documents.onDidClose((e) => {
...
+  return globby(`${workspaceFolderPath}/**/manifest.json`, {
+    cwd: `${workspaceFolderPath}`,
+  });
+}
+
+async function setManifestDocuments(): Promise<void[]> {
```
bd82, line 255: set --> initialize

packages/language-server/src/server.ts
bd82, line 249: The name `getManifestDocumentsPromise` is too generic.
the `get` / `set` prefixes should be used mainly for simple getter/setters.
But here there is more advanced logic. in our case we are **scanning/reading/finding** all the manifest.json files. 

So the name of the function should reflect that.

also the suffix `promise` is redundant, its async and returns Promise of string[].

packages/language-server/src/server.ts
```diff
@@ -174,3 +225,60 @@ documents.onDidClose((e) => {
...
 connection.listen();
+
+function isManifestDoc(uri: string): boolean {
+  return uri.endsWith("manifest.json");
+}
+
+function getFlagForXMLFile(uri: string): boolean {
+  const manifestFilesForCurrentFolder = Object.keys(
+    manifestData
+  ).filter((manifestUri) => uri.includes(dirname(manifestUri)));
```
bd82, line 236: don't use built-in `.filter` use lodash's filter

bd82:
Please extract the manifest state handling to **another** file which will provide
utilities to the main server.ts file.

Also please go over the function and variable names again, for example `readFlag` is not accurate, 
which flag are we reading?

packages/language-server/src/server.ts
```diff
@@ -115,17 +132,36 @@ connection.onHover(
...
-  if (getSemanticModelPromise === undefined) {
+  if (
+    getSemanticModelPromise === undefined ||
+    getInitialManifestStatePromise === undefined ||
+    isManifestDoc(changeEvent.document.uri)
+  ) {
     return;
   }
   const ui5Model = await getSemanticModelPromise;
+  await getInitialManifestStatePromise;
```
bd82, line 155: `await manifestStateInitialized;`?

packages/language-server/src/server.ts
```diff
@@ -25,15 +22,35 @@ import {
...
 const documents = new TextDocuments(TextDocument);
 let getSemanticModelPromise: Promise<UI5SemanticModel> | undefined = undefined;
+let getInitialManifestStatePromise: Promise<void[]> | undefined = undefined;
 let initializationOptions: ServerInitializationOptions | undefined;
 let hasConfigurationCapability = false;
 
 connection.onInitialize((params: InitializeParams) => {
   const capabilities = params.capabilities;
+  const workspaceFolderUri = params.rootUri;
+  if (workspaceFolderUri) {
```
bd82, line 47: subjective:

Avoid auto casting, be more specific, e.g:
- workspaceFolderUri !== undefined ?

packages/language-server/src/server.ts
```diff
@@ -25,15 +22,35 @@ import {
...
+    const workspaceFolderPath = URI.parse(workspaceFolderUri).fsPath;
```
bd82, line 48: subjective:

Do we want to emphasize that this is an absolute path because of future
multi root scenarios that would require absolute paths?

- `abs` suffix/prefix/?

packages/vscode-ui5-language-assistant/src/extension.ts
```diff
@@ -36,7 +36,10 @@ export async function activate(context: ExtensionContext): Promise<void> {
   const clientOptions: LanguageClientOptions = {
     documentSelector: [{ scheme: "file", pattern: "**/*.{view,fragment}.xml" }],
     synchronize: {
-      fileEvents: workspace.createFileSystemWatcher("**/*.{view,fragment}.xml"),
+      fileEvents: [
+        workspace.createFileSystemWatcher("**/*.{view,fragment}.xml"),
```
bd82, line 40: do we still need this `.xml` file event? because we don't seem to handle `manfiest.json` in `onContentChange`?

packages/language-server/src/server.ts
```diff
@@ -115,17 +132,36 @@ connection.onHover(
...
+    const uri = change.uri;
+    if (!isManifestDoc(uri)) {
+      return;
+    }
+
+    await updateManifestData(uri, change.type);
+  });
+});
+
 documents.onDidChangeContent(async (changeEvent) => {
```
bd82, line 146: would `onDidChangeContent` ever be triggered for manifest.json?

packages/language-server/src/server.ts
```diff
@@ -115,17 +132,36 @@ connection.onHover(
...
   }
   const ui5Model = await getSemanticModelPromise;
+  await getInitialManifestStatePromise;
   // TODO: should we check we are dealing with a *.[view|fragment].xml?
   //       The client does this, but perhaps we should be extra defensive in case of
   //       additional clients.
   const documentUri = changeEvent.document.uri;
   const document = documents.get(documentUri);
   if (document !== undefined) {
+    // We should pass the flag to diagnostics
```
bd82, line 162: is this comment redundant?

bd82, line 162: add this comment as `TODO` directly above the call to the diagnostics

packages/language-server/src/server.ts
```diff
@@ -115,17 +132,36 @@ connection.onHover(
...
+    const documentPath = URI.parse(documentUri).fsPath;
+    getFlexEnabledFlagForXMLFile(documentPath);
```
bd82, line 164: assign this flag to a temp variable (do eslint/ignore)

packages/language-server/src/manifest-handling.ts
```diff
@@ -0,0 +1,100 @@
+import { maxBy, map, filter } from "lodash";
+import { readFile } from "fs-extra";
+import { URI } from "vscode-uri";
+import { dirname } from "path";
```
bd82, line 4: could you put import "path" at the top, as it is nodejs builtin

packages/language-server/src/manifest-handling.ts
```diff
@@ -0,0 +1,100 @@
...
+import globby from "globby";
+import { FileChangeType } from "vscode-languageserver";
+
+type AbsolutePath = string;
+type ManifestData = Record<AbsolutePath, { isFlexEnabled: boolean }>;
+const manifestData: ManifestData = Object.create(null);
+let manifestDocuments: string[];
```
bd82, line 11: evaluate if we need this variable at the module "root" level?

packages/language-server/src/server.ts
```diff
@@ -115,17 +132,36 @@ connection.onHover(
...
+    await updateManifestData(uri, change.type);
+  });
+});
+
 documents.onDidChangeContent(async (changeEvent) => {
-  if (getSemanticModelPromise === undefined) {
+  if (
+    getSemanticModelPromise === undefined ||
+    getInitialManifestStatePromise === undefined ||
+    isManifestDoc(changeEvent.document.uri)
```
bd82, line 150: isManifestDoc --> isXMLView

packages/language-server/src/manifest-handling.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export function isManifestDoc(uri: string): boolean {
+  return uri.endsWith("manifest.json");
+}
+
+export async function initializeManifestDocuments(
+  workspaceFolderPath: string
+): Promise<void[]> {
+  manifestDocuments = await findAllManifestDocumentsInWorkspace(
```
bd82, line 20: `manifestDocs` as local var

packages/language-server/src/manifest-handling.ts
```diff
@@ -0,0 +1,100 @@
...
+    workspaceFolderPath
+  );
+  const readManifestPromises = map(manifestDocuments, async (manifestDoc) => {
+    const isFlexEnabled = await readFlexEnabledFlagFromManifestFile(
+      manifestDoc
+    );
+    // Parsing of manifest.json failed because the file is invalid
```
bd82, line 27: the comment describes what happened (failure to parse).
But not why we ignore this scenario, because we want to keep the last valid state of the manifest (e.g manifest.json being live edited).

packages/language-server/src/manifest-handling.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+  });
+
+  return Promise.all(readManifestPromises);
+}
+
+export function getFlexEnabledFlagForXMLFile(xmlPath: string): boolean {
+  const manifestFilesForCurrentFolder = filter(
+    Object.keys(manifestData),
+    (manifestPath) => xmlPath.includes(dirname(manifestPath))
```
bd82, line 39: maybe comment that include is sufficent because we use absolute paths.

bd82: maybe use startWith to reduce need for comment?

packages/language-server/src/manifest-handling.ts
```diff
@@ -0,0 +1,100 @@
...
+  );
+  const requiredManifestPath = maxBy(
```
bd82, line 41: `requiredManifestPath` --> "closestManifestPath` ?

packages/language-server/src/manifest-handling.ts
```diff
@@ -0,0 +1,100 @@
...
+  manifestDocuments = await findAllManifestDocumentsInWorkspace(
+    workspaceFolderPath
+  );
+  const readManifestPromises = map(manifestDocuments, async (manifestDoc) => {
+    const isFlexEnabled = await readFlexEnabledFlagFromManifestFile(
+      manifestDoc
+    );
+    // Parsing of manifest.json failed because the file is invalid
+    if (isFlexEnabled !== "INVALID") {
+      manifestData[manifestDoc] = { isFlexEnabled };
```
bd82, line 29: do we keep the structure like manifest.json `flexEnabled` ?

bd82: reasoning is: do we keep our manifest related data as a strict subset of the real manifest data?

packages/language-server/src/manifest-handling.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  return manifestData[requiredManifestPath].isFlexEnabled;
+}
+
+export async function updateManifestData(
+  manifestUri: string,
+  changeType: FileChangeType
+): Promise<void> {
+  const isFlexEnabled = await readFlexEnabledFlagFromManifestFile(manifestUri);
```
bd82, line 57: only `readFlexEnabledFlagFromManifestFile` when the file actually exists (not deleted)

packages/language-server/src/manifest-handling.ts
```diff
@@ -0,0 +1,100 @@
...
+  const manifestPath = URI.parse(manifestUri).fsPath;
+  switch (changeType) {
+    case 1: //created
+    case 2: //changed
+      // Parsing of manifest.json failed because the file is invalid
```
bd82, line 62: Change comment to describe the `why`.
We want to keep last successfully read state, because on the manifet.json file may be actively edited.

packages/language-server/src/manifest-handling.ts
```diff
@@ -0,0 +1,100 @@
...
+      delete manifestData[manifestPath];
+      return;
+  }
+}
+
+async function findAllManifestDocumentsInWorkspace(
+  workspaceFolderPath: string
+): Promise<string[]> {
+  return globby(`${workspaceFolderPath}/**/manifest.json`, {
+    cwd: `${workspaceFolderPath}`,
```
bd82, line 77: inspect if the CWD is needed as we are using an absolute path
and also the main arg path includes the CWD provided.

packages/language-server/src/manifest-handling.ts
```diff
@@ -0,0 +1,100 @@
...
+): Promise<boolean | "INVALID"> {
+  const manifestContent = await readFile(
+    URI.parse(manifestUri).fsPath,
+    "utf-8"
+  );
+  let manifestJsonObject;
+  try {
+    manifestJsonObject = JSON.parse(manifestContent);
+  } catch (err) {
+    console.log(err);
```
bd82, line 92: avoid console.log, we need a proper logger, but this my cause spam
especially  as manifet.json files being actively edited will often fail here.

packages/language-server/src/manifest-handling.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  return manifestData[closestManifestPath].flexEnabled;
+}
+
+export async function updateManifestData(
+  manifestUri: string,
+  changeType: FileChangeType
+): Promise<void> {
+  const manifestPath = URI.parse(manifestUri).fsPath;
+  /*eslint no-case-declarations: "error"*/
```
bd82, line 60: remove eslint comment

packages/language-server/src/server.ts
```diff
@@ -25,15 +22,33 @@ import {
...
-let getSemanticModelPromise: Promise<UI5SemanticModel> | undefined = undefined;
+let semanticModelLoaded: Promise<UI5SemanticModel> | undefined = undefined;
+let manifestStateInitialized: Promise<void[]> | undefined = undefined;
 let initializationOptions: ServerInitializationOptions | undefined;
 let hasConfigurationCapability = false;
 
 connection.onInitialize((params: InitializeParams) => {
   const capabilities = params.capabilities;
+  const workspaceFolderUri = params.rootUri;
+  if (workspaceFolderUri !== null) {
```
bd82, line 47: ensure that starting with empty workspace and then opening a new one/folder
will re-trigger this code.

packages/language-server/src/server.ts
```diff
@@ -115,17 +128,38 @@ connection.onHover(
...
+    manifestStateInitialized === undefined ||
+    !isXMLView(changeEvent.document.uri)
+  ) {
     return;
   }
-  const ui5Model = await getSemanticModelPromise;
+
+  const ui5Model = await semanticModelLoaded;
+  await manifestStateInitialized;
   // TODO: should we check we are dealing with a *.[view|fragment].xml?
```
bd82, line 153: you can remove the TODO: you already handled it 👍

packages/language-server/src/server.ts
```diff
@@ -174,3 +208,7 @@ documents.onDidClose((e) => {
 documents.listen(connection);
 
 connection.listen();
+
+function isXMLView(uri: string): boolean {
+  return /(view|fragment)\.xml/i.test(uri);
```
bd82, line 213: remove case insensitive flag and use end of input anchor
to ensure .xml is end of file name.

bd82: the goal is to align this regexp with the documentSelector pattern in the client.

PR #231, opened 2020-07-13
feat: stable ID validation

sapirpol:


packages/language-server/scripts/update-diagnostics-snapshots.js
```diff
@@ -3,6 +3,7 @@ const klawSync = require("klaw-sync");
 const { forEach, filter, map } = require("lodash");
 const { resolve, dirname } = require("path");
 const { writeFileSync } = require("fs");
+const { existsSync } = require("fs-extra");
```
bd82, line 6: `existsSync` is part of "fs" no need to import it from fs-extra.
- `fs-extra` is an automatic wrapper to `fs`

packages/xml-views-validation/api.d.ts
```diff
@@ -4,6 +4,7 @@ import { UI5SemanticModel } from "@ui5-language-assistant/semantic-model-types";
 export function validateXMLView(opts: {
   model: UI5SemanticModel;
   xmlView: XMLDocument;
+  flexEnabled: boolean;
```
bd82, line 7: make flexEnabled optional (default value at runtime) so we will avoid a **breaking change** in the API

packages/language-server/src/xml-view-diagnostics.ts
```diff
@@ -21,13 +21,15 @@ import {
 export function getXMLViewDiagnostics(opts: {
   document: TextDocument;
   ui5Model: UI5SemanticModel;
+  flexEnabled: boolean;
```
bd82, line 24: Make `flexEnabled` optional to avoid breaking changes.

packages/language-server/test/snapshots/xml-view-diagnostics/non-stable-id/input.xml
```diff
@@ -0,0 +1,15 @@
...
+    xmlns="sap.ui.commons"
+    xmlns:sap.ui.dt=🢂"sap.ui.dt"🢀>
+        <m:CheckBox id="dummy-id"></m:CheckBox>
+        <🢂m:Button🢀>
+            <m:customData>
+                <core:CustomData></core:CustomData>
+            </m:customData>
+        </m:Button>
+        <m:Panel sap.ui.dt:designtime="not-adaptable">
+            <🢂m:Button🢀>
```
bd82, line 12: perhaps avoid using `m:Button` twice in this snippet to make the expected errors easier to read and link back to the xml snippet?

packages/language-server/test/snapshots/xml-view-diagnostics/snapshots-diagnostics-spec.ts
```diff
@@ -1,10 +1,12 @@
+import { resolve, basename, relative, dirname } from "path";
 import klawSync from "klaw-sync";
 import { map, forEach, filter } from "lodash";
-import { basename, relative, dirname } from "path";
+import { existsSync } from "fs-extra";
```
bd82, line 4: existsSync is part of `fs` native node.js module

packages/language-server/test/snapshots/xml-view-diagnostics/snapshots-diagnostics-spec.ts
```diff
@@ -16,20 +18,27 @@ describe(`The language server diagnostics capability`, () => {
   });
   const testDirs = map(inputFiles, (_) => dirname(_.path));
 
-  forEach(testDirs, (dirPath) => {
+  forEach(testDirs, async (dirPath) => {
```
bd82, line 21: do we need this function do be `async` ?

packages/xml-views-validation/src/api.ts
```diff
@@ -1,12 +1,20 @@
...
 import { XMLDocument } from "@xml-tools/ast";
 import { UI5XMLViewIssue } from "../api";
 import { validateXMLView as validateXMLViewImpl } from "./validate-xml-views";
 import { allValidators } from "./validators";
+import { validateNonStableId } from "./validators/elements/non-stable-id";
 
 export function validateXMLView(opts: {
   model: UI5SemanticModel;
   xmlView: XMLDocument;
+  flexEnabled: boolean;
```
bd82, line 12: `flexEnabled` -> optional

packages/xml-views-validation/src/utils/custom-class.ts
```diff
@@ -0,0 +1,36 @@
+import { SVG_NS, TEMPLATING_NS, XHTML_NS } from "./special-namespaces";
```
bd82, line 1: probably copy->pasta, but please move the import of `"./special-namespaces";` to be 
the **last** import in this file.

packages/xml-views-validation/src/utils/custom-class.ts
```diff
@@ -0,0 +1,36 @@
...
+import { XMLElement } from "@xml-tools/ast";
+import { resolveXMLNS } from "@ui5-language-assistant/logic-utils";
+
+// Heuristic to limit false positives by only checking tags starting with upper
+// case names, This would **mostly** limit the checks for things that can actually be
+// UI5 Elements / Controls.
+export function isCustomClass(xmlElement: XMLElement): boolean {
+  // This will never happen - it was checked before calling this function
```
bd82, line 9: > // This will never happen - it was checked before calling this function

We now use this function in multiple places, its a mini API.
I don't think we can still make this assumption about the null value.
But perhaps we can model it in the signature instead.

```typescript
export function isCustomClass(xmlElement: XMLElement & { name: string}): boolean 
```

packages/xml-views-validation/src/utils/custom-class.ts
```diff
@@ -0,0 +1,36 @@
...
+  /* istanbul ignore if */
+  if (xmlElement.name === null) {
+    return false;
+  }
+
+  return /^[A-Z]/.test(xmlElement.name) && !isNoneUI5id(xmlElement);
```
bd82, line 15: I am aware this is copy/paste from my previous code.
But perhaps take the chance to improve the name.

The double negative of `!isNoneUI5id` can be confusing.
But I am having trouble thinking of another name :(

packages/xml-views-validation/src/utils/custom-class.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// We only care about UI5 elements/controls IDs when check non-unique IDs
```
bd82, line 18: > // We only care about UI5 elements/controls IDs when check non-unique IDs

This comment is outdated, it refers to a specific validation but this is not a more generic utility.

packages/xml-views-validation/src/utils/messages.ts
```diff
@@ -13,6 +13,7 @@ export const UNKNOWN_TAG_NAME_NO_NS = `The "{0}" name is neither a class name no
 export const INVALID_AGGREGATION_CARDINALITY = `The aggregation "{0}" has cardinality of 0..1 and may only contain one element`;
 export const INVALID_AGGREGATION_TYPE = `The class "{0}" is under the aggregation "{1}" and must match the type "{2}"`;
 export const NON_UNIQUE_ID = `Duplicate ID: "{0}" found.`;
+export const NON_STABLE_ID = `The class "{0}" is missing stable ID`;
```
bd82, line 16: Perhaps it would be more informative as a variation of:

`The Class Button must declare a none empty ID property when flexEnabled is "true"?`

packages/xml-views-validation/test/validators/index-spec.ts
```diff
@@ -13,7 +13,10 @@ describe("The `allValidators` constant", () => {
   const filePaths = map(klawItems, (_) => _.path);
   validatorPaths = filter(
     filePaths,
-    (_) => _.endsWith(".js") && !_.endsWith("index.js")
+    (_) =>
+      _.endsWith(".js") &&
+      !_.endsWith("index.js") &&
+      !_.endsWith("non-stable-id.js")
```
bd82, line 19: perhaps a short comment here why we exclude `non-stable-id.js`?

packages/xml-views-validation/src/utils/custom-class.ts
```diff
@@ -0,0 +1,36 @@
+import { SVG_NS, TEMPLATING_NS, XHTML_NS } from "./special-namespaces";
+import { XMLElement } from "@xml-tools/ast";
+import { resolveXMLNS } from "@ui5-language-assistant/logic-utils";
+
+// Heuristic to limit false positives by only checking tags starting with upper
+// case names, This would **mostly** limit the checks for things that can actually be
+// UI5 Elements / Controls.
+export function isCustomClass(xmlElement: XMLElement): boolean {
```
bd82, line 8: This is a **heuristic**, so perhaps call it:
`isPossibleCustomClass` ?

packages/xml-views-validation/test/api-spec.ts
```diff
@@ -34,12 +34,47 @@ describe("the ui5 xml views validations API", () => {
...
+    });
     expect(issues).to.have.lengthOf(2);
     const issueTypes = map(issues, (_) => _.kind);
     expect(issueTypes).to.include.members([
       "UnknownEnumValue",
       "UseOfDeprecatedClass",
     ]);
   });
+
+  it("will detect semantic UI5 xml view issues (smoke) when flexEnabled is true", () => {
```
bd82, line 50: perhaps reduce this snippet to only test the stableID issue?
Do we need to duplicate large parts of the previous test?

packages/xml-views-validation/src/validators/elements/non-stable-id.ts
```diff
@@ -0,0 +1,123 @@
...
+export function validateNonStableId(
+  xmlElement: XMLElement,
+  model: UI5SemanticModel
+): NonStableIDIssue[] {
+  // Can't give an error if there is no position or name
+  if (xmlElement.name === null || xmlElement.syntax.openName === undefined) {
+    return [];
+  }
+
+  if (isWhiteListed(xmlElement, model)) {
```
bd82, line 23: If all scenarios of the `whiteList` are UI5 classes edge cases
perhaps we can give it a more accurate name, e.g: `isWhiteListClass`
or maybe the more general `isWhiteListedTag` 
or `isTagExcludedFromStableIDCheck`?

WDYT?

packages/xml-views-validation/src/validators/elements/non-stable-id.ts
```diff
@@ -0,0 +1,123 @@
...
+    return [];
+  }
+
+  if (!isCustomClass(xmlElement)) {
```
bd82, line 27:
```typescript
if (!isCustomClass(xmlElement)) {
    return [];
  }
```

This reads like the check is only for custom Classes.
But logically the check is for either classes we identify (in model) OR customClasses.

So there is an assumption that `is[Possible]CustomClass` will return true for all classes in the model, which is probably correct now, but not necessarily in the future.

I'd rather have a more robust check:
```typescript
if (!isPossibleCustomClass(xmlElement) && !isKnownUI5Control/Element/Class) {
    return [];
  }
```

packages/xml-views-validation/src/validators/elements/non-stable-id.ts
```diff
@@ -0,0 +1,123 @@
...
+    offsetRange: {
+      start: xmlElement.syntax.openName.startOffset,
+      end: xmlElement.syntax.openName.endOffset,
+    },
+  };
+
+  return [nonStableIDIssue];
+}
+
+function isWhiteListed(
```
bd82, line 55: - Do we need to use two APIs for checking the names?
  - `ui5NodeToFQN` vs resolveNS + includes
  - could/should we use only the resolve + includes for consistency?

- Do we need to distinguish between things a the root or not?
  - because `["Fragment", "CustomData", "ExtensionPoint"]` would still return true if they are in the root if I read this correctly.

So maybe this function could be refactored, maybe even simplified greatly.

packages/xml-views-validation/src/validators/elements/non-stable-id.ts
```diff
@@ -0,0 +1,123 @@
...
+  );
+}
+
+function hasNonAdaptableTreeMetaData(xmlElement: XMLElement): boolean {
+  let currElement = xmlElement;
+  while (currElement.parent.type !== "XMLDocument") {
+    const hasNonAdaptableTreeMetaData = some(
+      currElement.attributes,
+      (attribute) =>
+        attribute.key === "sap.ui.dt:designtime" &&
```
bd82, line 101: add a TODO here to inspect if we need to properly resolve the attribute "NS"

bd82: or if a plain string matcher is sufficent

packages/xml-views-validation/src/validators/elements/non-stable-id.ts
```diff
@@ -0,0 +1,123 @@
...
+        attribute.value === "not-adaptable-tree"
+    );
+
+    if (hasNonAdaptableTreeMetaData) {
+      return true;
+    }
+
+    currElement = currElement.parent;
```
bd82, line 109: does `"not-adaptable-tree"` affect the immediate tag or only the children tags?

packages/xml-views-validation/src/validators/elements/non-stable-id.ts
```diff
@@ -0,0 +1,123 @@
...
+  return false;
+}
+
+function isElementWithStableID(xmlElement: XMLElement): boolean {
+  return some(
+    xmlElement.attributes,
+    (attribute) =>
+      attribute.key === "id" &&
+      attribute.value !== null &&
+      /\S/.test(attribute.value)
```
bd82, line 121: /\S/.test(attribute.value) means there is some none whitespace character present.

- WS definition in XML may be different than JS regExp.
- We are dealing with WS inside string literal / value so the definition may
  once again be different.
- https://www.w3.org/TR/REC-xml/#sec-common-syn

I suggest inspect the stringLiteral / Value tokens in the XML-Tools/Parser
and align the check with that.
- https://github.com/SAP/xml-tools/blob/master/packages/parser/lib/lexer.js#L140-L143
- Although I am not 100% that is accurate either :)

It may be more **clear** to test **NOT all characters are whitespace**.

packages/xml-views-validation/test/validators/element/non-stable-id-spec.ts
```diff
@@ -0,0 +1,172 @@
...
+            <🢂m:Panel🢀>
+            </m:Panel>
+        </mvc:View>`,
+        getMessage(NON_STABLE_ID, "Panel")
+      );
+    });
+
+    it("will detect missing stable id in custom control", () => {
+      assertSingleIssue(
+        `<🢂custom:View🢀
```
bd82, line 50: won't it be better to provide a more "real life" example where the custom class
replaces the Button instead of the top level view?

packages/xml-views-validation/test/validators/element/non-stable-id-spec.ts
```diff
@@ -0,0 +1,172 @@
...
+        getMessage(NON_STABLE_ID, "View")
+      );
+    });
+
+    it("will detect missing stable id in whitelisted root class when it's not in the root", () => {
+      assertSingleIssue(
+        `<mvc:View xmlns:uxap="sap.uxap" xmlns:m="sap.m"
+            xmlns:mvc="sap.ui.core.mvc"
+            xmlns="sap.ui.commons">
+              <🢂m:Panel🢀>
```
bd82, line 65: since when is `Panel` whitelisted?

packages/xml-views-validation/test/validators/element/non-stable-id-spec.ts
```diff
@@ -0,0 +1,172 @@
...
+              </m:Panel>
+          </mvc:View>`,
+        getMessage(NON_STABLE_ID, "Panel")
+      );
+    });
+
+    it("will detect missing stable in sub element when the element has attribute sap.ui.dt:designtime='not-adaptable'", () => {
```
bd82, line 72: `the element` -> `the parent element`

packages/xml-views-validation/test/validators/element/non-stable-id-spec.ts
```diff
@@ -0,0 +1,172 @@
...
+  let ui5SemanticModel: UI5SemanticModel;
+
+  before(async () => {
+    ui5SemanticModel = await generateModel({
+      version: "1.74.0",
+      modelGenerator: generate,
+    });
+  });
+
+  context("true positive scenarios", () => {
```
bd82, line 22: add tests for empty id value and null id value cases?

packages/xml-views-validation/test/validators/element/non-stable-id-spec.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  context("negative edge cases", () => {
+    let assertNoIssues: (xmlSnippet: string) => void;
+    before(() => {
+      assertNoIssues = partial(assertNoIssuesBase, ui5SemanticModel, {
+        element: [validateNonStableId],
+      });
+    });
+
+    it("will not detect an issue when there is a root-whitelisted UI5 class", () => {
```
bd82, line 95: `when there is a` --> `for a`

packages/xml-views-validation/test/validators/element/non-stable-id-spec.ts
```diff
@@ -0,0 +1,172 @@
...
+      assertNoIssues(
+        `<!-- sap.ui.core.mvc.View is whitlisted root class -->
+        <mvc:View xmlns:uxap="sap.uxap" xmlns:m="sap.m"
+          xmlns:mvc="sap.ui.core.mvc"
+          xmlns="sap.ui.commons">
+        </mvc:View>`
+      );
+    });
+
+    it("will not detect an issue when there is a root-whitelisted UI5 class - core ns", () => {
```
bd82, line 105: `when there is` -> `for a` because the existence of the whitelisted element
does not mean all other elements would not be checked

packages/xml-views-validation/test/validators/element/non-stable-id-spec.ts
```diff
@@ -0,0 +1,172 @@
...
+            <m:Button id="yyy">
+                <m:customData>
+                    <core:CustomData></core:CustomData>
+                </m:customData>
+            </m:Button>
+        </mvc:View>`
+      );
+    });
+
+    it("will not detect an issue when the control has attribute attribute sap.ui.dt:designtime='not-adaptable'", () => {
```
bd82, line 127: ` attribute attribute` - >  `attribute`

packages/xml-views-validation/test/validators/element/non-stable-id-spec.ts
```diff
@@ -0,0 +1,172 @@
...
+            xmlns:mvc="sap.ui.core.mvc"
+            xmlns="sap.ui.commons"
+            xmlns:sap.ui.dt="sap.ui.dt">
+              <m:Panel sap.ui.dt:designtime="not-adaptable">
+              </m:Panel>
+        </mvc:View>`
+      );
+    });
+
+    it("will not detect an issue when the control has attribute attribute sap.ui.dt:designtime='not-adaptable'", () => {
```
bd82, line 139: `attribute attribute` -> `attribute`

bd82, line 139: am I missing something or is this test copy/pasted twice?

packages/xml-views-validation/test/validators/element/non-stable-id-spec.ts
```diff
@@ -0,0 +1,172 @@
...
+          xmlns:mvc="sap.ui.core.mvc"
+          xmlns="sap.ui.commons"
+          xmlns:sap.ui.dt="sap.ui.dt">
+            <m:Panel sap.ui.dt:designtime="not-adaptable">
+            </m:Panel>
+        </mvc:View>`
+      );
+    });
+
+    it("will not detect an issue when the control has attribute attribute sap.ui.dt:designtime='not-adaptable-tree'", () => {
```
bd82, line 151: - attribute word * 2
- emphasize it would not detect issues for **sub-elements**

packages/xml-views-validation/src/validators/elements/non-stable-id.ts
sapirpol: both

packages/xml-views-validation/src/utils/messages.ts
```diff
@@ -13,6 +13,7 @@ export const UNKNOWN_TAG_NAME_NO_NS = `The "{0}" name is neither a class name no
 export const INVALID_AGGREGATION_CARDINALITY = `The aggregation "{0}" has cardinality of 0..1 and may only contain one element`;
 export const INVALID_AGGREGATION_TYPE = `The class "{0}" is under the aggregation "{1}" and must match the type "{2}"`;
 export const NON_UNIQUE_ID = `Duplicate ID: "{0}" found.`;
+export const NON_STABLE_ID = `The class "{0}" must declare a none empty ID attribute when flexEnabled is "true"`;
```
bd82, line 16: I think the term is: "https://www.merriam-webster.com/dictionary/non-empty"

"non-empty"

packages/xml-views-validation/src/utils/ui5-classes.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const whiteListedNamespaces: Record<string, boolean> = {
+  [SVG_NS]: true,
+  [TEMPLATING_NS]: true,
+  [XHTML_NS]: true,
+};
+
+Object.freeze(whiteListedNamespaces);
+
+function isPartOfUI5Namespace(xmlElement: XMLElement): boolean {
```
bd82, line 35: `isPartOfUI5Namespace` is misleading, its not **part of** UI5.

But I am having trouble with finding a good name, maybe:
- `isCommonExternalNamespace` ?

packages/xml-views-validation/src/validators/document/non-unique-id.ts
```diff
@@ -77,11 +73,7 @@ class IdsCollectorVisitor implements XMLAstVisitor {
       attrib.value !== "" &&
       attrib.syntax.value !== undefined &&
       attrib.parent.name !== null &&
-      // Heuristic to limit false positives by only checking tags starting with upper
-      // case names, This would **mostly** limit the checks for things that can actually be
-      // UI5 Elements / Controls.
-      /^[A-Z]/.test(attrib.parent.name) &&
-      !isNoneUI5id(attrib)
+      isPossibleCustomClass(attrib.parent as XMLElement & { name: string })
```
bd82, line 76: why is casting needed, should not `attrib.parent.name !== null &&` cause narrowing of the type?

packages/xml-views-validation/src/validators/elements/non-stable-id.ts
```diff
@@ -0,0 +1,128 @@
...
+  if (xmlElement.name === null || xmlElement.syntax.openName === undefined) {
+    return [];
+  }
+
+  if (isWhiteListedClass(xmlElement)) {
+    return [];
+  }
+
+  if (
+    !isPossibleCustomClass(xmlElement as XMLElement & { name: string }) &&
```
bd82, line 27: why is casting needed when we have `(xmlElement.name === null` before ?

packages/xml-views-validation/src/validators/elements/non-stable-id.ts
```diff
@@ -0,0 +1,128 @@
...
+    offsetRange: {
+      start: xmlElement.syntax.openName.startOffset,
+      end: xmlElement.syntax.openName.endOffset,
+    },
+  };
+
+  return [nonStableIDIssue];
+}
+
+function isWhiteListedClass(xmlElement: XMLElement): boolean {
```
bd82, line 57: I am not sure these are all `classes`

can we rename to "isWhiteListedTag" ?

packages/xml-views-validation/test/validators/element/non-stable-id-spec.ts
bd82: this comment does not seem to e resolved

sapirpol: we need this test for root-level scenario

PR #234, opened 2020-07-15
fix: ignore unknown namespace message for whitelisted namespaces

sapirpol:


packages/xml-views-validation/src/validators/attributes/unknown-xmlns-namespace.ts
```diff
@@ -19,9 +19,14 @@ export function validateUnknownXmlnsNamespace(
...
   const attributeValue = attribute.value;
   const attributeValueToken = attribute.syntax.value;
+  const whiteListedNamespaces = ["sap.ui.dt"];
 
   // TODO empty namespaces aren't valid but this should be handled in xml-tools because it's a general xml issue.
-  if (attributeValueToken === undefined || attributeValue === null) {
+  if (
+    attributeValueToken === undefined ||
+    attributeValue === null ||
+    includes(whiteListedNamespaces, attributeValue)
```
bd82, line 28: put the check for whitelist in a different if statement

packages/xml-views-validation/src/validators/attributes/unknown-xmlns-namespace.ts
```diff
@@ -19,9 +19,14 @@ export function validateUnknownXmlnsNamespace(
 
   const attributeValue = attribute.value;
   const attributeValueToken = attribute.syntax.value;
+  const whiteListedNamespaces = ["sap.ui.dt"];
```
bd82, line 22: move this const just before the new if statement

PR #266, opened 2020-07-30
feat: non stable id quick fix

sapirpol:


packages/language-server/src/quick-fix.ts
```diff
@@ -0,0 +1,103 @@
...
+  CodeActionKind,
+  TextDocumentEdit,
+  TextEdit,
+} from "vscode-languageserver-types";
+import { getQuickFixIdInfo } from "@ui5-language-assistant/xml-views-quick-fix";
+import { parse, DocumentCstNode } from "@xml-tools/parser";
+import { buildAst, XMLDocument } from "@xml-tools/ast";
+import { LSPRangeToOffsetRange, offsetRangeToLSPRange } from "./range-utils";
+
+export function getQuickFixCodeAction(
```
bd82, line 17: Can we decouple this form `vscode-languageserver-types`
What if someone wants to use this quick-fix package in implementing some other language-server?

packages/language-server/src/quick-fix.ts
```diff
@@ -0,0 +1,103 @@
...
+  document: TextDocument,
+  diagnostic: Diagnostic
+): CodeAction | undefined {
+  const documentText = document.getText();
+  // We prefer to parse the document again to avoid cache state handling
+  const { cst, tokenVector } = parse(documentText);
+  const xmlDocAst = buildAst(cst as DocumentCstNode, tokenVector);
+  switch (diagnostic.code) {
+    case 666: {
```
bd82, line 26: we need a better diagnostic code number

packages/language-server/src/quick-fix.ts
```diff
@@ -0,0 +1,100 @@
...
+import { parse, DocumentCstNode } from "@xml-tools/parser";
+import { buildAst, XMLDocument } from "@xml-tools/ast";
+import { LSPRangeToOffsetRange, offsetRangeToLSPRange } from "./range-utils";
+
+export function getQuickFixCodeAction(
+  document: TextDocument,
+  diagnostic: Diagnostic
+): CodeAction | undefined {
+  const documentText = document.getText();
+  // We prefer to parse the document again to avoid cache state handling
```
bd82, line 22: Should we do the parsing at the language server side to enable
another language server to potential implement the caching on its side?

Right now this implementation **forces** no caching.

packages/xml-views-quick-fix/api.d.ts
```diff
@@ -0,0 +1,12 @@
+import { XMLDocument } from "@xml-tools/ast";
+import { OffsetRange } from "@ui5-language-assistant/xml-views-validation";
```
bd82, line 2: `OffsetRange` should probably be in logic utils

packages/language-server/src/quick-fix.ts
bd82: Examples of error status codes.
- https://github.com/Microsoft/TypeScript/blob/v1.6.2/src/compiler/diagnosticMessages.json

packages/language-server/src/quick-fix.ts
```diff
@@ -0,0 +1,100 @@
...
+  const { cst, tokenVector } = parse(documentText);
+  const xmlDocAst = buildAst(cst as DocumentCstNode, tokenVector);
+  switch (diagnostic.code) {
+    case 666: {
+      // non stable id
+      return getCodeActionForQuickFixId({
```
bd82, line 28: `getCodeActionForQuickFixId` 

- Maybe the word `stable` should appear here?
- Maybe the word `get` is not 100% accurate, perhaps its `compute`?

packages/language-server/src/quick-fix.ts
bd82: The diagnostic code should be a cost exported from the validations package as a public API.

packages/language-server/src/range-utils.ts
```diff
@@ -0,0 +1,25 @@
+import { TextDocument } from "vscode-languageserver";
+import { Range as LSPRange } from "vscode-languageserver-types";
+import { OffsetRange } from "@ui5-language-assistant/xml-views-validation";
```
bd82, line 3: `OffsetRange` should probably be in `logic-utils`

packages/language-server/src/server.ts
```diff
@@ -67,6 +69,10 @@ connection.onInitialize((params: InitializeParams) => {
         triggerCharacters: ['"', "'", ":", "<"],
       },
       hoverProvider: true,
+      codeActionProvider: true,
+      executeCommandProvider: {
```
bd82, line 73: I could use an explanation how this  `executeCommandProvider` works

packages/language-server/src/quick-fix.ts
```diff
@@ -0,0 +1,100 @@
...
+        xmlDocument: xmlDocAst,
+        nonStableIdDiagnostic: diagnostic,
+      });
+    }
+    default:
+      return undefined;
+  }
+}
+
+export function executeCommand(
```
bd82, line 39: `executeCommand` seems like a **generic** flow that should be in server.ts or at least not in quick-fix.ts

packages/language-server/src/quick-fix.ts
```diff
@@ -0,0 +1,100 @@
...
+  textDocument: TextDocument,
+  params: ExecuteCommandParams
+): TextDocumentEdit[] | undefined {
+  switch (params.command) {
+    case "nonStableIdQuickFix": {
+      return executeQuickFixIdCommand({
+        textDocument,
+        // @ts-expect-error - we already checked arguments exist
```
bd82, line 47: where did we previously check the argument lists?

packages/language-server/src/quick-fix.ts
```diff
@@ -0,0 +1,100 @@
...
+      return undefined;
+  }
+}
+
+export function executeCommand(
+  textDocument: TextDocument,
+  params: ExecuteCommandParams
+): TextDocumentEdit[] | undefined {
+  switch (params.command) {
+    case "nonStableIdQuickFix": {
```
bd82, line 44: this (`nonStableIdQuickFix`) looks like a const that should be extracted somewhere, it is used in **three** places.

packages/xml-views-quick-fix/test/quick-fix-id-spec.ts
```diff
@@ -0,0 +1,171 @@
...
+      const { document, start, end, idStartOffest } = getXmlSnippet(
+        testXmlSnippet
+      );
+      const testXmlDoc = getXmlDocument(document);
+      const quickFixInfo = getQuickFixIdInfo(testXmlDoc, { start, end });
+      expectExists(quickFixInfo, "Quick Fix Info");
+      expect(quickFixInfo.suggestion).to.equal(expectedSuggestion);
+      expect(quickFixInfo.offsetRange.start).to.equal(idStartOffest);
+    });
+  });
```
bd82, line 107: what happens if there are multiple attributes each in its own line?
would the insertion respect the existing indentation levels?

packages/xml-views-quick-fix/test/quick-fix-id-spec.ts
```diff
@@ -0,0 +1,171 @@
...
+      const testXmlDoc = getXmlDocument(document);
+      const quickFixInfo = getQuickFixIdInfo(testXmlDoc, {
+        start,
+        end,
+      });
+
+      expect(quickFixInfo).to.not.exist;
+    });
+
+    it("will not suggest quick fix id when the tag is empty", () => {
```
bd82, line 128: the desciption string `will not suggest quick fix id when the tag is empty` appears twice in this file

packages/xml-views-quick-fix/api.d.ts
```diff
@@ -0,0 +1,12 @@
...
+
+type quickFixIdInfo = {
```
bd82, line 4: - type name starts with upper case.
- add `Stable` inside this name?

packages/xml-views-quick-fix/api.d.ts
```diff
@@ -0,0 +1,12 @@
...
+  suggestion: string;
+  offsetRange: OffsetRange;
+};
+
+export declare function getQuickFixIdInfo(
```
bd82, line 9: it is a `getter` or does it `compute` / `calculate it?

bd82, line 9: add `Stable` inside this function name?

packages/vscode-ui5-language-assistant/test/suite/validations/flex-enabled/flex-enabled-spec.ts
```diff
@@ -57,6 +57,7 @@ describe("the Language Server Client Validations Integration Tests - Flex Enable
         message: getMessage(NON_STABLE_ID, "Panel"),
         range: getRanges(xmlSnippet)[0],
         source: UI5LANG_ERROR_MSG,
+        code: 666,
```
bd82, line 60: once you extract this error code to a const, use it here.

packages/language-server/src/xml-view-diagnostics.ts
```diff
@@ -56,9 +55,13 @@ function validationIssuesToLspDiagnostics(
       case "UnknownTagName":
       case "InvalidAggregationCardinality":
       case "InvalidAggregationType":
+        return {
+          ...commonDiagnosticPros,
+        };
       case "NonStableIDIssue":
         return {
           ...commonDiagnosticPros,
+          code: 666,
```
bd82, line 64: - extract to const,
- expose from validations package?
- maybe expose from messages package we will create in the future?

packages/xml-views-quick-fix/package.json
```diff
@@ -0,0 +1,43 @@
...
+  ],
+  "main": "lib/src/api.js",
+  "repository": "https://github.com/sap/ui5-language-assistant/",
+  "license": "Apache-2.0",
+  "typings": "./api.d.ts",
+  "engines": {
+    "node": ">=10.0.0"
+  },
+  "dependencies": {
+    "@ui5-language-assistant/xml-views-validation": "^0.6.0",
```
bd82, line 21: remove the dep to `validations` once you extract the interface fomr it

packages/xml-views-quick-fix/test/quick-fix-id-spec.ts
```diff
@@ -0,0 +1,171 @@
...
+  const end = xmlSnippet.indexOf("🢀");
+  xmlText = xmlText.replace("🢀", "");
+  const idStartOffest = xmlText.indexOf("$");
+  xmlText = xmlText.replace("$", "");
+  const document: TextDocument = createTextDocument("xml", xmlText);
+  return { document, start, end, idStartOffest };
+}
+
+function createTextDocument(languageId: string, content: string): TextDocument {
+  return TextDocument.create("uri", languageId, 0, content);
```
bd82, line 170: see other packages, they have an "implementation-matches-api" pseudo test.

packages/language-server/src/server.ts
```diff
@@ -164,6 +170,43 @@ documents.onDidChangeContent(async (changeEvent) => {
...
+connection.onCodeAction((params) => {
+  const docUri = params.textDocument.uri;
+  const codeActions: CodeAction[] = [];
+  const textDocument = documents.get(docUri);
+  if (textDocument === undefined) {
+    return undefined;
+  }
+
+  const diagnostics = params.context.diagnostics;
+  forEach(diagnostics, (_) => {
```
bd82, line 182: try lodash `map` + `compact` instead

packages/language-server/src/server.ts
```diff
@@ -164,6 +170,43 @@ documents.onDidChangeContent(async (changeEvent) => {
...
+  if (params.arguments === undefined) {
+    return;
+  }
+
+  const textDocument = documents.get(params.arguments[0]);
+  if (textDocument === undefined) {
+    return;
+  }
+
+  const textEdit = executeCommand(textDocument, params);
```
bd82, line 202: I think there may be too many assumptions here.
- Should we check more clearely what type of command it before executing it?
- Would all commands we ever implement will return textEdits?

packages/language-server/src/quick-fix.ts
```diff
@@ -0,0 +1,100 @@
...
+  if (quickFixIdInfo === undefined) {
+    return undefined;
+  }
+
+  const title = "Generate ID";
+  return CodeAction.create(
+    title,
+    Command.create(
+      title,
+      "nonStableIdQuickFix",
```
bd82, line 78: `nonStableIdQuickFix` is a constat?

packages/language-server/src/quick-fix.ts
```diff
@@ -0,0 +1,100 @@
...
+    opts.nonStableIdDiagnostic.range,
+    opts.document
+  );
+
+  const quickFixIdInfo = getQuickFixIdInfo(opts.xmlDocument, errorOffset);
+  if (quickFixIdInfo === undefined) {
+    return undefined;
+  }
+
+  const title = "Generate ID";
```
bd82, line 73: This seems like another **user facing** constant

packages/language-server/src/quick-fix.ts
sapirpol: in server.ts - `connection.onExecuteCommand`

packages/xml-views-quick-fix/test/quick-fix-id-spec.ts
sapirpol: in this case, the id will be generated in the tag line and other attributes will not be affected

packages/language-server/src/xml-view-diagnostics.ts
sapirpol: I'll add & expose error code for all validations in a different PR

packages/language-server/src/quick-fix.ts
```diff
@@ -0,0 +1,86 @@
...
+  CodeActionKind,
+  TextDocumentEdit,
+  TextEdit,
+} from "vscode-languageserver-types";
+import { computeQuickFixStableIdInfo } from "@ui5-language-assistant/xml-views-quick-fix";
+import { parse, DocumentCstNode } from "@xml-tools/parser";
+import { buildAst, XMLDocument } from "@xml-tools/ast";
+import { LSPRangeToOffsetRange, offsetRangeToLSPRange } from "./range-utils";
+
+export const QUICK_FIX_STABLE_ID = "nonStableIdQuickFix";
```
bd82, line 16: command prefix?

packages/language-server/src/quick-fix.ts
```diff
@@ -0,0 +1,86 @@
...
+const QUICK_FIX_STABLE_ID_TITLE = "Generate ID";
+
+export function getQuickFixCodeAction(
```
bd82, line 19: `diagnosticToCodeActionFix`?

packages/language-server/src/quick-fix.ts
```diff
@@ -0,0 +1,86 @@
...
+  document: TextDocument,
+  diagnostic: Diagnostic
+): CodeAction | undefined {
+  const documentText = document.getText();
+  // We prefer to parse the document again to avoid cache state handling
+  const { cst, tokenVector } = parse(documentText);
+  const xmlDocAst = buildAst(cst as DocumentCstNode, tokenVector);
+  switch (diagnostic.code) {
+    case 1000: {
+      // non stable id
```
bd82, line 29: redundant comment

packages/language-server/src/quick-fix.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  const quickFixIdInfo = computeQuickFixStableIdInfo(
+    opts.xmlDocument,
+    errorOffset
+  );
+  if (quickFixIdInfo === undefined) {
+    return undefined;
+  }
+
+  const title = QUICK_FIX_STABLE_ID_TITLE;
```
bd82, line 59: redundant const `title`

packages/language-server/src/quick-fix.ts
```diff
@@ -0,0 +1,86 @@
...
+  return CodeAction.create(
+    title,
+    Command.create(
```
bd82, line 62: This is a "bad" API because of 5 params.
Its not easy to understand what each param is.

- Ensure your const names are "optimal".
- `offsetRangeToLSPRange(quickFixIdInfo.offsetRange, opts.document)` to temp variable with meaningful name?

bd82: a word with `replace` may fit the extracted variable imho.

packages/language-server/src/quick-fix.ts
```diff
@@ -0,0 +1,86 @@
...
+      QUICK_FIX_STABLE_ID,
+      opts.document.uri,
+      offsetRangeToLSPRange(quickFixIdInfo.offsetRange, opts.document),
+      quickFixIdInfo.suggestion
+    ),
+    CodeActionKind.QuickFix
+  );
+}
+
+export function executeQuickFixIdCommand(opts: {
```
bd82, line 73: `stable` word missing?

packages/language-server/src/server.ts
```diff
@@ -67,6 +77,11 @@ connection.onInitialize((params: InitializeParams) => {
         triggerCharacters: ['"', "'", ":", "<"],
       },
       hoverProvider: true,
+      codeActionProvider: true,
+      // Each command executes a different code action scenario
+      executeCommandProvider: {
+        commands: [QUICK_FIX_STABLE_ID],
```
bd82, line 83: Do we need to ensure uniqueness of command names?
If our name is not very unique, could the command ever be executed due to another cause (e.g enother plugin)?

How do other extensions provide commands? is there a unique prefix, e.g: `ui5_language_assistant.quick_fix_stable_id` ?

packages/xml-views-quick-fix/test/quick-fix-id-spec.ts
bd82: o.k, that is acceptable and may be upgraded in the future.

Lets add a couple of tests with multi line element:
- attributes exists on other lines.
- invalid id attribute exists on a different line than the opening tag

packages/xml-views-quick-fix/src/api.ts
```diff
@@ -0,0 +1 @@
+export { computeQuickFixStableIdInfo } from "./quick-fix-id";
```
bd82, line 1: stable-id?

packages/language-server/src/server.ts
```diff
@@ -211,3 +250,26 @@ connection.listen();
...
+  switch (params.command) {
+    case QUICK_FIX_STABLE_ID_COMMAND: {
+      const change = executeQuickFixStableIdCommand({
+        textDocument,
+        // @ts-expect-error - we already checked arguments exist
+        quickFixRange: params.arguments[1],
+        // @ts-expect-error - we already checked arguments exist
+        quickFixIDSuggestion: params.arguments[2],
+      });
+      connection.workspace.applyEdit({
```
bd82, line 267: we discussed online that the applyEdit may need to happen inside `executeQuickFixStableIdCommand`

packages/xml-views-quick-fix/api.d.ts
```diff
@@ -0,0 +1,8 @@
+import { XMLDocument } from "@xml-tools/ast";
+import { OffsetRange } from "@ui5-language-assistant/logic-utils";
```
bd82, line 2: if you have a dependency to logic-utils in the api.d.ts
it also has to be in dependency section in package.json (not devDep)

packages/language-server/src/quick-fix.ts
```diff
@@ -0,0 +1,89 @@
...
+      QUICK_FIX_STABLE_ID_COMMAND,
+      opts.document.uri,
+      suggestionRange,
+      quickFixIdInfo.suggestion
+    ),
+    CodeActionKind.QuickFix
+  );
+}
+
+export function executeQuickFixStableIdCommand(opts: {
```
bd82, line 76: This function may need to actually execute the textEdit because
other commands may do other things unrelated to textEdits, so its a concern of the QuickFix.

packages/xml-views-quick-fix/src/quick-fix-id.ts
```diff
@@ -0,0 +1,107 @@
+import { find, some } from "lodash";
```
bd82, line 1: I will review this file tomorrow

packages/language-server/src/server.ts
```diff
@@ -164,6 +175,34 @@ documents.onDidChangeContent(async (changeEvent) => {
...
+
+  return codeActions;
+});
+
+connection.onExecuteCommand(async (params) => {
+  if (params.arguments === undefined) {
+    return;
+  }
+
+  const textDocument = documents.get(params.arguments[0]);
```
bd82, line 198: - document assumption that all commands have textDocument URI in the first argument
- document doing runtime checks here.

packages/language-server/src/server.ts
```diff
@@ -211,3 +250,26 @@ connection.listen();
...
+
+function executeCommand(
+  textDocument: TextDocument,
+  params: ExecuteCommandParams
+): void {
+  switch (params.command) {
+    case QUICK_FIX_STABLE_ID_COMMAND: {
+      const change = executeQuickFixStableIdCommand({
+        textDocument,
+        // @ts-expect-error - we already checked arguments exist
```
bd82, line 262: instead of many (with many commands) `ts-expect-error` comments
consider treating the params.argument as `any` type, e.g:

```
  params: ExecuteCommandParams & { arguments: any}

```

packages/language-server/src/server.ts
sapirpol: as discussed, because we don't want to pass the connection to another `executeQuickFixStableIdCommand` we'll keep it this way

packages/xml-views-quick-fix/src/quick-fix-id.ts
```diff
@@ -0,0 +1,107 @@
...
+export type QuickFixStableIdInfo = {
+  suggestion: string;
+  offsetRange: OffsetRange;
+};
+
+export function computeQuickFixStableIdInfo(
+  xmlDoc: XMLDocument,
+  errorOffset: OffsetRange
+): QuickFixStableIdInfo | undefined {
+  const biggestIdsCollector = new IdsByElementNameCollectorVisitor();
```
bd82, line 24: the line:
`const biggestIdsCollector = new IdsByElementNameCollectorVisitor();`

Hints the class name of the visitor is inaccurate,
Perhaps it should be more similar to the variable name on the left hand side

packages/xml-views-quick-fix/src/quick-fix-id.ts
```diff
@@ -0,0 +1,107 @@
...
+  accept(xmlDoc, biggestIdsCollector);
+  const biggestIdsOfElements = biggestIdsCollector.biggestIdsOfElements;
+  const astNode = astPositionAtOffset(xmlDoc, errorOffset.start);
+  if (astNode === undefined || astNode.kind !== "XMLElementOpenName") {
```
bd82, line 28: `astNode?.kind !== "XMLElementOpenName"`?

packages/xml-views-quick-fix/src/quick-fix-id.ts
```diff
@@ -0,0 +1,107 @@
...
+    return undefined;
+  }
+
+  const xmlElement = astNode.astNode;
```
bd82, line 32: discussion: 
if we know that for `XMLEelementOpenName` there would always be:
- .syntax.openName
- .name

Should we create a branch in the code that is never reached due to type system limitations? or should we define / cast the xmlElement's type more precisely?

 
```typescript
const xmlElement = astNode.astNode as XMLElement & { ... }
```

packages/xml-views-quick-fix/src/quick-fix-id.ts
```diff
@@ -0,0 +1,107 @@
...
+  /* istanbul ignore if - ast node of kind "XMLElementOpenName" will always have name  */
+  if (xmlElement.name === null || xmlElement.syntax.openName === undefined) {
+    return undefined;
+  }
+
+  const isExistIdKey = some(
```
bd82, line 38: `isExistIdKey` -> `hasIdAttribute` ?

bd82, line 38: isExistIdKey --> `hasIdAttribute` ?

packages/xml-views-quick-fix/src/quick-fix-id.ts
```diff
@@ -0,0 +1,107 @@
...
+    xmlElement.attributes,
+    (attrib) => attrib.key === "id"
+  );
+
+  const suggestion = getQuickFixIdSuggestion(
```
bd82, line 43: `get` -> `compute` ?

packages/xml-views-quick-fix/src/quick-fix-id.ts
```diff
@@ -0,0 +1,107 @@
...
+    biggestIdsOfElements,
+    xmlElement.name,
+    isExistIdKey
+  );
+
+  const offsetRange = getQuickFixIdRange(xmlElement);
```
bd82, line 49: could we be more specific here? is this the **replace** range?
It may help someone who is unfamiliar with this code understand flow better

packages/xml-views-quick-fix/src/quick-fix-id.ts
```diff
@@ -0,0 +1,107 @@
...
+  return { suggestion, offsetRange };
+}
+
+class IdsByElementNameCollectorVisitor implements XMLAstVisitor {
+  public biggestIdsOfElements: Record<string, number> = Object.create(null);
```
bd82, line 54: Provide an in-depth comment about this class and what we are collecting.
We have a none trivial logic, a few comments and maybe even micro example could come in handy

packages/xml-views-quick-fix/src/quick-fix-id.ts
```diff
@@ -0,0 +1,107 @@
...
+        this.biggestIdsOfElements[className] = suffix;
+      }
+    }
+  }
+}
+
+function getQuickFixIdSuggestion(
+  biggestIdsOfElements: Record<string, number>,
+  elementName: string,
+  isExistIdKey: boolean
```
bd82, line 78: hasIdAttribute?

packages/xml-views-quick-fix/src/quick-fix-id.ts
```diff
@@ -0,0 +1,107 @@
...
+): string {
+  const suffix = biggestIdsOfElements[elementName]
+    ? biggestIdsOfElements[elementName] + 1
+    : 1;
+  let suggestion = `id="${ID_PREFIX_PATTERN}${elementName}${suffix}"`;
+  if (!isExistIdKey) {
+    // We want extra space if there is no id key
```
bd82, line 85: This comment could be more useful if it explained the **why** 😄

packages/xml-views-quick-fix/src/quick-fix-id.ts
```diff
@@ -0,0 +1,107 @@
...
+    suggestion = " " + suggestion;
+  }
+
+  return suggestion;
+}
+
+function getQuickFixIdRange(xmlElement: XMLElement): OffsetRange {
+  const idAttrib = find(xmlElement.attributes, (attrib) => attrib.key === "id");
+  if (idAttrib !== undefined && idAttrib.syntax.key !== undefined) {
```
bd82, line 94: `idAttrib.?syntax.key !== undefined`?

packages/xml-views-quick-fix/src/quick-fix-id.ts
```diff
@@ -0,0 +1,107 @@
...
+    return {
+      start: idAttrib.position.startOffset,
+      end: idAttrib.position.endOffset,
+    };
+  } else {
+    return {
+      //@ts-expect-error - we already checked the element has open name
```
bd82, line 101: we can't assume that `syntax.openName` exists just because we checked it in another flow.
Who is to say that this function would **only ever be executed from** said other flow.

Can you model the assumption in the type signature of the function?
That way it won't be callable unless those checks were performed by the caller first.

packages/xml-views-quick-fix/src/quick-fix-id.ts
```diff
@@ -0,0 +1,107 @@
...
+  let suggestion = `id="${ID_PREFIX_PATTERN}${elementName}${suffix}"`;
+  if (!isExistIdKey) {
+    // We want extra space if there is no id key
+    suggestion = " " + suggestion;
+  }
+
+  return suggestion;
+}
+
+function getQuickFixIdRange(xmlElement: XMLElement): OffsetRange {
```
bd82, line 92: This function includes special semantic quick fix logic, which may not be trivial.
Could we add a short comment on each branch to help a reader understand our logic.

e.g:
- branch1 - replace whole ID attribute area.
- branch2 - no id, insert immediately after open tag

packages/language-server/src/quick-fix.ts
sapirpol: as discussed, because we don't want to pass the connection to another executeQuickFixStableIdCommand we'll keep it this way

PR #275, opened 2020-08-05
chore: remove old copy of manifest.json schema

bd82:
the schema store reference is used instead (see package.json).


bd82:
Waiting for confirmation that this resource is no longer used anywhere before mergining.

bd82:
> Waiting for confirmation that this resource is no longer used anywhere before merging.

Confirmation received, merging.

PR #283, opened 2020-08-11
feat: quick fix stable id for entire file

sapirpol:


packages/language-server/src/quick-fix.ts
```diff
@@ -12,78 +12,182 @@ import { computeQuickFixStableIdInfo } from "@ui5-language-assistant/xml-views-q
 import { parse, DocumentCstNode } from "@xml-tools/parser";
 import { buildAst, XMLDocument } from "@xml-tools/ast";
 import { LSPRangeToOffsetRange, offsetRangeToLSPRange } from "./range-utils";
+import {
+  validateXMLView,
+  validateNonStableId,
+} from "@ui5-language-assistant/xml-views-validation";
+import { UI5SemanticModel } from "@ui5-language-assistant/semantic-model-types";
+import { map, forEach } from "lodash";
```
bd82, line 20: import order
- external deps.
- mono repo deps
- relative deps.

packages/language-server/src/quick-fix.ts
```diff
@@ -12,78 +12,182 @@ import { computeQuickFixStableIdInfo } from "@ui5-language-assistant/xml-views-q
...
+import { Range } from "vscode-languageserver-textdocument";
 
 export const QUICK_FIX_STABLE_ID_COMMAND = "ui5_lang.quick_fix_stable_id";
+export const QUICK_FIX_FILE_STABLE_ID_COMMAND =
+  "ui5_lang.quick_fix_file_stable_id";
 const QUICK_FIX_STABLE_ID_COMMAND_TITLE = "Generate ID";
+const QUICK_FIX_FILE_STABLE_ID_COMMAND_TITLE = "Generate IDs for all file";
```
bd82, line 27: We need UA, but maybe
> "Generate IDs for entire file"; ?

packages/language-server/src/quick-fix.ts
```diff
@@ -12,78 +12,182 @@ import { computeQuickFixStableIdInfo } from "@ui5-language-assistant/xml-views-q
...
 }
 
-function computeCodeActionForQuickFixStableId(opts: {
+function computeCodeActionsForQuickFixStableId(opts: {
   document: TextDocument;
   xmlDocument: XMLDocument;
   nonStableIdDiagnostic: Diagnostic;
-}): CodeAction | undefined {
+  ui5Model: UI5SemanticModel;
+}): CodeAction[] | undefined {
```
bd82, line 63: do we still need `| undefined` in the return type here?

packages/language-server/src/quick-fix.ts
```diff
@@ -12,78 +12,182 @@ import { computeQuickFixStableIdInfo } from "@ui5-language-assistant/xml-views-q
...
+  xmlDocument: XMLDocument;
+  ui5Model: UI5SemanticModel;
+}): CodeAction[] {
+  const validators = {
+    document: [],
+    element: [validateNonStableId],
+    attribute: [],
+  };
+
+  const nonStableIdFileIssues = validateXMLView({
```
bd82, line 118: document that we are re-validating intentionally to keep the flow simple & stateless

packages/language-server/src/quick-fix.ts
```diff
@@ -12,78 +12,182 @@ import { computeQuickFixStableIdInfo } from "@ui5-language-assistant/xml-views-q
...
+    validators,
+    model: opts.ui5Model,
+    xmlView: opts.xmlDocument,
+  });
+
+  if (nonStableIdFileIssues.length === 1) {
```
bd82, line 124: - This has to be **documented**, `=== 1` means there is only one issue, so the original "solo" quick fix while suffix.

packages/language-server/src/quick-fix.ts
```diff
@@ -12,78 +12,182 @@ import { computeQuickFixStableIdInfo } from "@ui5-language-assistant/xml-views-q
...
+    return [];
+  }
+
+  const errorsOffset = map(nonStableIdFileIssues, (_) => _.offsetRange);
```
bd82, line 128: instead of checking if some length === 1,
Could we remove the known diagnostic range from `nonStableIdFileIssues` before doing the map? or is it too complicated because we have different kinds of ranges here?

packages/language-server/src/quick-fix.ts
```diff
@@ -12,78 +12,182 @@ import { computeQuickFixStableIdInfo } from "@ui5-language-assistant/xml-views-q
...
+  const nonStableIdFileIssuesInfo = computeQuickFixStableIdInfo(
+    opts.xmlDocument,
+    errorsOffset
+  );
+  const nonStableIdFileIssuesLSPInfo: QuickFixStableIdLSPInfo[] = map(
+    nonStableIdFileIssuesInfo,
+    (_) => {
+      return {
```
bd82, line 136: you can do `(_) => ({ prop1: "", prop2: 666})` and avoid the `return` keyword

packages/language-server/src/quick-fix.ts
```diff
@@ -12,78 +12,182 @@ import { computeQuickFixStableIdInfo } from "@ui5-language-assistant/xml-views-q
...
   return documentEdit;
 }
+
+export function executeQuickFixFileStableIdCommand(opts: {
+  documentUri: string;
+  documentVersion: number;
+  nonStableIdIssues: QuickFixStableIdLSPInfo[];
+}): TextDocumentEdit[] {
+  const textEdits: TextEdit[] = [];
+  forEach(opts.nonStableIdIssues, (_) => {
```
bd82, line 181: `forEach + push === map`

packages/language-server/src/quick-fix.ts
```diff
@@ -12,78 +12,182 @@ import { computeQuickFixStableIdInfo } from "@ui5-language-assistant/xml-views-q
...
   ];
 
   return documentEdit;
 }
+
+export function executeQuickFixFileStableIdCommand(opts: {
+  documentUri: string;
+  documentVersion: number;
+  nonStableIdIssues: QuickFixStableIdLSPInfo[];
+}): TextDocumentEdit[] {
```
bd82, line 179: Do we need `TextDocumentEdit[]` **array** in the return type?
it seems to always be an array sized **1**?

Or is that to support some generic pattern by the caller?

packages/xml-views-validation/src/validators/index.ts
```diff
@@ -11,7 +11,20 @@ import { validateUnknownTagName } from "./elements/unknown-tag-name";
...
+export { validateUseOfDeprecatedAggregation } from "./elements/use-of-depracated-aggregation";
+export { validateUseOfDeprecatedAttribute } from "./attributes/use-of-depracated-attribute";
+export { validateNonUniqueID } from "./document/non-unique-id";
+export { validateUnknownAttributeKey } from "./attributes/unknown-attribute-key";
+export { validateUnknownTagName } from "./elements/unknown-tag-name";
+export { validateExplicitAggregationCardinality } from "./elements/cardinality-of-aggregation";
+export { validateAggregationType } from "./elements/type-of-aggregation";
+export { validateNonStableId } from "./elements/non-stable-id";
+
+export const defaultValidators: UI5Validators = {
```
bd82, line 27: now that we expose this `defaultValidators` object can we **deepFreeze** it to avoid externals mutating it?

packages/xml-views-validation/api.d.ts
```diff
@@ -93,3 +102,60 @@ export interface NonUniqueIDIssue extends BaseUI5XMLViewIssue {
 export interface NonStableIDIssue extends BaseUI5XMLViewIssue {
   kind: "NonStableIDIssue";
 }
+
+export function validateUnknownEnumValue(
```
bd82, line 106: This public API is a little verbose.
It makes me think we want to expose all these validators
inside some object? and sub properties?

bd82: we discussed this on voice

packages/xml-views-validation/api.d.ts
```diff
@@ -1,22 +1,31 @@
-import { XMLDocument } from "@xml-tools/ast";
+import { XMLDocument, XMLElement, XMLAttribute } from "@xml-tools/ast";
 import { UI5SemanticModel } from "@ui5-language-assistant/semantic-model-types";
 import { OffsetRange } from "@ui5-language-assistant/logic-utils";
+import { UI5Validators } from "./src/validate-xml-views";
 
 export function validateXMLView(opts: {
+  validators: UI5Validators;
```
bd82, line 7: maybe call this type: `ValidatorsConfig`?

packages/xml-views-quick-fix/src/quick-fix-stable-id.ts
```diff
@@ -84,6 +92,7 @@ function computeQuickFixIdSuggestion(
   const suffix = biggestIdsByElementName[elementName]
     ? biggestIdsByElementName[elementName] + 1
     : 1;
+  biggestIdsByElementName[elementName] = suffix;
```
bd82, line 95: document that we are mutating that data structure to support multiple fixes in the same file without conflicts

packages/language-server/src/server.ts
```diff
@@ -174,18 +178,29 @@ documents.onDidChangeContent(async (changeEvent) => {
...
   const textDocument = documents.get(docUri);
   if (textDocument === undefined) {
     return undefined;
   }
 
   const diagnostics = params.context.diagnostics;
-  const codeActions = compact(
-    map(diagnostics, (_) => diagnosticToCodeActionFix(textDocument, _))
-  );
+  const nonStableIdIssue = find(diagnostics, (_) => _.code === 1000);
```
bd82, line 194: feels like this condition is too specific for stableID fixes.
Should/could we preform this check inside `diagnosticToCodeActionFix`?

packages/vscode-ui5-language-assistant/test/test-fixtures/validations/flex-enabled-spec/test.view.xml
```diff
@@ -0,0 +1,17 @@
+<mvc:View xmlns:core="sap.ui.core" xmlns:uxap="sap.uxap" xmlns:m="sap.m"
```
bd82, line 1: where is this fixture used?

packages/xml-views-validation/api.d.ts
sapirpol: UI5ValidatorsConfig?

packages/language-server/src/quick-fix.ts
sapirpol: It's a genetic pattern by the caller

bd82:
We need to add `BREAKING CHANGE:` section in the squashed commit before we merge

PR #290, opened 2020-08-16
test: add tests for quick fix

sapirpol:


packages/vscode-ui5-language-assistant/test/suite/validations/flex-enabled/flex-enabled-spec.ts
```diff
@@ -62,6 +64,79 @@ describe("the Language Server Client Validations Integration Tests - Flex Enable
     ]);
   });
 
+  it("will quick fix a non-stable id issue", async () => {
```
bd82, line 67: it looks like you copied the previous test to add a new part at the "end" of the test.
Perhaps just add the new part to the pre-existing test?

bd82: on second thought it may be best to define each expected functionality differently.

packages/vscode-ui5-language-assistant/test/suite/validations/flex-enabled/flex-enabled-spec.ts
```diff
@@ -62,6 +64,79 @@ describe("the Language Server Client Validations Integration Tests - Flex Enable
...
+    await vscode.commands.executeCommand(
+      //@ts-expect-error - test assumption
+      fixes[0].command.command,
+      //@ts-expect-error - test assumption
+      ...fixes[0].command.arguments
+    );
+
+    await sleep(EXECUTE_COMMAD_TIMEOUT);
+    expectProblemView(xmlUri, []);
+    await vscode.commands.executeCommand("workbench.action.closeActiveEditor");
```
bd82, line 103: Why is `await vscode.commands.executeCommand("workbench.action.closeActiveEditor");` needed at the end?

Would follow up tests not work if this was not executed?
If so perhaps it would be in an `afterEach` section.

packages/vscode-ui5-language-assistant/test/suite/validations/flex-enabled/flex-enabled-spec.ts
```diff
@@ -62,6 +64,79 @@ describe("the Language Server Client Validations Integration Tests - Flex Enable
...
+            xmlns="sap.ui.commons">
+                <⭲m:Panel⭰>
+                  <m:Button></m:Button>
+                </m:Panel>
+                <m:Panel>
+                </m:Panel>
+        </mvc:View>`;
+    await setFileTextContents(xmlSnippet, xmlPath);
+    const actualDiagnostics = vscode.languages.getDiagnostics(xmlUri);
+    expect(actualDiagnostics).to.not.be.empty;
```
bd82, line 119: `expect(actualDiagnostics).to.not.be.empty;`

This does **not** prove there are **multiple** stableID issues in the file.
So the test does **not** prove "entire file" issues were fixed.

packages/vscode-ui5-language-assistant/test/suite/validations/flex-enabled/flex-enabled-spec.ts
```diff
@@ -62,6 +64,79 @@ describe("the Language Server Client Validations Integration Tests - Flex Enable
...
+    await vscode.commands.executeCommand(
+      //@ts-expect-error - test assumption
+      fixes[1].command.command,
+      //@ts-expect-error - test assumption
+      ...fixes[1].command.arguments
+    );
+
+    await sleep(EXECUTE_COMMAD_TIMEOUT);
+    expectProblemView(xmlUri, []);
+    await vscode.commands.executeCommand("workbench.action.closeActiveEditor");
```
bd82, line 137: same comment regarding `afterEach` as above

packages/vscode-ui5-language-assistant/test/suite/validations/flex-enabled/flex-enabled-spec.ts
```diff
@@ -62,6 +64,79 @@ describe("the Language Server Client Validations Integration Tests - Flex Enable
...
+      "vscode.executeCodeActionProvider",
+      xmlUri,
+      getRanges(xmlSnippet)[0],
+      vscode.CodeActionKind.QuickFix
+    );
+
+    expect(fixes).to.exist;
+    await vscode.commands.executeCommand(
+      //@ts-expect-error - test assumption
+      fixes[1].command.command,
```
bd82, line 130: Does `fixes[1].command` have an identifying properties? could we add
an assertion on those so it would be clearer which command is activated?

packages/vscode-ui5-language-assistant/test/suite/validations/flex-enabled/flex-enabled-spec.ts
```diff
@@ -62,6 +64,79 @@ describe("the Language Server Client Validations Integration Tests - Flex Enable
...
+      "vscode.executeCodeActionProvider",
+      xmlUri,
+      getRanges(xmlSnippet)[0],
+      vscode.CodeActionKind.QuickFix
+    );
+
+    expect(fixes).to.exist;
+    await vscode.commands.executeCommand(
+      //@ts-expect-error - test assumption
+      fixes[0].command.command,
```
bd82, line 96: Does `fixes[0].command` have an identifying properties? could we add
an assertion on those so it would be clearer which command is activated?

packages/xml-views-quick-fix/test/quick-fix-stable-id-spec.ts
```diff
@@ -5,6 +5,8 @@ import { buildAst, XMLDocument } from "@xml-tools/ast";
 import { expectExists } from "@ui5-language-assistant/test-utils";
 import { computeQuickFixStableIdInfo } from "../src/quick-fix-stable-id";
 
+const UNFOUND_CHAR = -1;
```
bd82, line 8: indexOf !== -1 is idiomatic in JavaScript.
I actually found: `const UNFOUND_CHAR = -1;` to be confusing. (pun intended).

That is because:
- you probably meant: "**NOT_FOUND**_CHAR" because `unfounded` means something else:
  - https://www.dictionary.com/browse/unfounded
- The `_CHAR` is likely redundant because:
  - We don't care what we are searching for or on what object we are searching on, e.g:
    - Lodash's `findIndex` also returns `-1` when **NOT_FOUND**.
 
Options:
1. Use const.
  - call it `NOT_FOUND`
  - define it directly above the **single** function it is used in.
2. Use `-1` directly.

packages/xml-views-quick-fix/test/quick-fix-stable-id-spec.ts
```diff
@@ -16,19 +18,59 @@ describe("the UI5 language assistant QuickFix Service", () => {
...
       ]);
       expectExists(quickFixInfo, "Quick Fix Info");
       expect(quickFixInfo[0].newText).to.equal(expectedSuggestion);
-      expect(quickFixInfo[0].replaceRange.start).to.equal(idStartOffest);
+      expect(quickFixInfo[0].replaceRange.start).to.equal(
+        quickFixStableIdTestInfo[0].idStartOffest
+      );
+    });
+
+    it("will get quick fix suggestions when classes are missing id attribute key", () => {
```
bd82, line 39: The test title `"will get quick fix suggestions when classes are missing id attribute key"` is too similar to the previous test title.

Emphasize the difference, e.g by saying `multiple classes`

packages/xml-views-quick-fix/test/quick-fix-stable-id-spec.ts
```diff
@@ -16,19 +18,59 @@ describe("the UI5 language assistant QuickFix Service", () => {
...
+      const testXmlSnippet = `<mvc:View
+                                xmlns:mvc="sap.ui.core.mvc"
+                                xmlns="sap.m"> 
+                                  <mvc:content>
+                                    <🢂List🢀$></List>
+                                    <🢂List🢀$></List>
+                                  </mvc:content>
+                              </mvc:View>`;
+      const expectedFirstSuggestion = ' id="_IDGenList1"';
```
bd82, line 48: we do not need these consts.
They actually reduce readability of the test, because, when a developer **reads** the test they will need to resolve the reference and scroll up to understand what is actually expected.

`      expect(quickFixInfo[0].newText).to.equal(expectedFirstSuggestion);`

PR #318, opened 2020-10-12
fix(xml-views-validation): rename insensitive terms

michawai:


bd82:
1. you need to sign the DCO.
2. the commits messages needs to be in conventional commits format

PR #337, opened 2020-12-15
chore: usage analytics (BAS only)

bd82:


bd82:
Still a few TODOs left in the code. 
I need to investigate those.

packages/language-server/src/commads.ts
```diff
@@ -28,6 +29,7 @@ export function executeCommand(
       connection.workspace.applyEdit({
         documentChanges: change,
       });
+      track("MANIFEST_STABLE_ID", "single");
```
allaVolkov, line 32: i think you can use constant TRACK_EVENTS .MANIFEST_STABLE_ID here

packages/language-server/src/commads.ts
```diff
@@ -39,6 +41,7 @@ export function executeCommand(
       connection.workspace.applyEdit({
         documentChanges: change,
       });
+      track("MANIFEST_STABLE_ID", "multiple");
```
allaVolkov, line 44: and here

packages/language-server/src/hover.ts
```diff
@@ -28,6 +29,7 @@ export function getHoverResponse(
   if (astPosition !== undefined) {
     const ui5Node = findUI5HoverNodeAtOffset(astPosition, model);
     if (ui5Node !== undefined) {
+      track("XML_UI5_DOC_HOVER", ui5Node.kind);
```
allaVolkov, line 32: and here

allaVolkov (approved):
please see minor comments

packages/language-server/src/commads.ts
bd82: I made my own `track` function with a different API.
See:
```typescript
type TrackEventsKeys = keyof typeof TRACK_EVENTS;

export function track(
  eventType: TrackEventsKeys,
  ...custom_events: string[]
): void {
  try {
    getSWA().track(TRACK_EVENTS[eventType], custom_events);
  } catch (e) {
    console.error(`failed usage analytics tracking: ${e.message}`);
  }
}

So it accepts the key for the track event, not the value as the first argument
```

packages/language-server/src/hover.ts
bd82: see answer above

allaVolkov (approved):

packages/language-server/src/hover.ts
bd82: I must have not posted the answer above...
Anyhow the `track` method I am using here is a different API that accepts the **key** of the relevant event not the value. See:

```typescript
type TrackEventsKeys = keyof typeof TRACK_EVENTS;

export function track(
  eventType: TrackEventsKeys,
  ...custom_events: string[]
): void {
  try {
    getSWA().track(TRACK_EVENTS[eventType], custom_events);
  } catch (e) {
    console.error(`failed usage analytics tracking: ${e.message}`);
  }
}
```

bd82: oh it was just marked as resolved, never mind :)

bd82:
The coveralls report of -0.2% is related to two tracking calls in a file that is currently not tested at all.
So I cannot resolve it in the scope of this PR.

PR #348, opened 2020-12-24
feat: logging for language server

bd82:


packages/language-server/src/ui5-model.ts
```diff
@@ -54,14 +56,20 @@ export async function getSemanticModelWithFetcher(
...
       if (apiJson === undefined) {
+        getLogger().info("No cache found for UI5 lib", { libName });
         const url = baseUrl + libName.replace(/\./g, "/") + suffix;
         const response = await fetcher(url);
         if (response.ok) {
           apiJson = await response.json();
           await writeToCache(cacheFilePath, apiJson);
         } else {
-          console.error(`Could not read UI5 resources from ${url}`);
+          getLogger().error("Could not read UI5 resources from", { url });
```
tal-sapan, line 66: Currently with the noop logger, we won't write anything anywhere on these errors. Maybe instead of noop it should at least print to the console? Until we have the implementation that prints to a file (and maybe also the console so the user can easily see such errors)

packages/language-server/src/ui5-model.ts
```diff
@@ -102,7 +110,10 @@ async function writeToCache(
     try {
       await writeJson(filePath, apiJson);
     } catch (err) {
-      console.warn(`Could not cache UI5 resources to file ${filePath}`, err);
+      getLogger().warn("Could not read cache file For UI5 lib", {
```
tal-sapan, line 113:
```suggestion
      getLogger().warn("Could not write cache file For UI5 lib", {
```

packages/settings/api.d.ts
```diff
@@ -9,10 +9,36 @@ export interface CodeAssistSettings {
...
+  off: true;
+  messages: true;
+  verbose: true;
+}
+
+export const validTraceServerValues: IValidTraceServerValues;
+
+export interface LoggingSettings {
+  logging: {
+    level: "off" | "fatal" | "error" | "warn" | "info" | "debug" | "trace";
```
tal-sapan, line 26: Why not use `keyof IValidLoggingLevelValues` like you did in `TraceSettings`?

packages/settings/test/settings-spec.ts
```diff
@@ -1,13 +1,13 @@
...
-  resetSettings,
   getSettingsForDocument,
   getDefaultSettings,
   setGlobalSettings,
   setSettingsForDocument,
   hasSettingsForDocument,
   clearDocumentSettings,
   clearSettings,
-} from "../src/settings";
+} from "../src/api";
```
tal-sapan, line 9: Why did you change these imports?

packages/vscode-ui5-language-assistant/test/unit/.mocharc.js
```diff
@@ -1,3 +1,7 @@
+const chai = require("chai");
+const deepEqualInAnyOrder = require("deep-equal-in-any-order");
+chai.use(deepEqualInAnyOrder);
```
tal-sapan, line 3: Should this really be in the configuration file? Wouldn't it be better to do it in the test file?

tal-sapan (commented):
See comments

packages/language-server/src/logger.ts
```diff
@@ -0,0 +1,44 @@
+import { cloneDeep } from "lodash";
+import omitDeep from "omit-deep-lodash";
+
+// TODO: This interface should be provided by the logging library.
+// Initially we are only providing a fairly simple API, additional variants can be added as needed...
+export interface ILogger {
+  /* eslint-disable @typescript-eslint/no-explicit-any -- meta is an object with string type key and any type value */
```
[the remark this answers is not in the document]
bd82: Probably, but this interface should be (and is) part of the logging library.
Currently we cannot consume the logging library here so it is copy/pasted.

I may be able to remove as I will use the vscode-logging library.

packages/language-server/src/logger.ts
```diff
@@ -0,0 +1,44 @@
...
+  warn: (msg: string, meta?: Record<string, any>) => void;
+  info: (msg: string, meta?: Record<string, any>) => void;
+  debug: (msg: string, meta?: Record<string, any>) => void;
+  trace: (msg: string, meta?: Record<string, any>) => void;
+  /* eslint-enable @typescript-eslint/no-explicit-any -- meta is an object with string type key and any type value */
+}
+
+const NOOP_LOGGER: ILogger = {
+  /* eslint-disable @typescript-eslint/no-empty-function, @typescript-eslint/no-unused-vars, @typescript-eslint/no-explicit-any -- NOOP */
+  fatal: (msg: string, meta?: Record<string, any>) => {},
```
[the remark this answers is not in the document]
bd82: yeah it seems I can. Interesting...

packages/language-server/src/logger.ts
```diff
@@ -0,0 +1,44 @@
...
+  warn: (msg: string, meta?: Record<string, any>) => {},
+  info: (msg: string, meta?: Record<string, any>) => {},
+  debug: (msg: string, meta?: Record<string, any>) => {},
+  trace: (msg: string, meta?: Record<string, any>) => {},
+  /* eslint-enable @typescript-eslint/no-empty-function, @typescript-eslint/no-unused-vars, @typescript-eslint/no-explicit-any -- NOOP */
+};
+
+let logger = NOOP_LOGGER;
+export function getLogger(): ILogger {
+  return NOOP_LOGGER;
```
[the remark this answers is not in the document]
bd82: mistake

packages/language-server/src/logger.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export function initBasFileLogger(): void {
```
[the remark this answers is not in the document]
bd82: Only in BAS there will be FileLogger.
Otherwise there will be OutputChannelLogger

packages/language-server/src/logger.ts
```diff
@@ -0,0 +1,44 @@
...
+  const possibleSensitiveProps = ["uri"];
```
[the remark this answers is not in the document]
bd82: - `password, credentials` don't appear on any of the interfaces of `vscode-languageserver-protocol`.
- URIs however  appear may times.
- We are **not** allowed to collect any folder / file names due to GDPR... and if in some cases there would be file loggers (BAS) which may be **collected centrally** it means we must avoid logging any such information.

packages/language-server/src/logger.ts
```diff
@@ -0,0 +1,44 @@
...
+  // TODO: use this to transform the meta argument to the log methods and remove possible personal information
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  function removePossibleUserInformation(obj: Record<string, unknown>) {
+    const clonedObj = cloneDeep(obj);
+    return omitDeep(clonedObj, possibleSensitiveProps);
+  }
+  // TODO: init FILE Based BAS Logger
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
bd82: Thanks

packages/language-server/src/server.ts
```diff
@@ -130,29 +141,41 @@ connection.onHover(
...
     if (!isManifestDoc(uri)) {
       return;
     }
-
     await updateManifestData(uri, change.type);
   });
 });
 
 documents.onDidChangeContent(async (changeEvent) => {
+  getLogger().trace("`onDidChangeWatchedFiles` event", { changeEvent });
```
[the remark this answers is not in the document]
bd82: thanks

packages/language-server/src/server.ts
```diff
@@ -173,11 +196,13 @@ documents.onDidChangeContent(async (changeEvent) => {
       ui5Model,
       flexEnabled,
     });
+    getLogger().trace("computed diagnostics", { diagnostics });
```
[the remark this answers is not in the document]
bd82: - I will use a console Logger so we will get **structured** logging to the output channel.
- I agree that we do not necessarily need centralized logging for the scenario of LSP server / process
  as 99% of issues on it should be easily reproduced on a local dev machine.
- However, that is the **requirement** we have, so at least at the same time we will also enjoy
  **structured** logging, and more granular control of logging levels (once this PR is finished).

packages/language-server/src/server.ts
```diff
@@ -220,26 +247,33 @@ function ensureDocumentSettingsUpdated(resource: string): void {
...
+  getLogger().debug("`onDidChangeConfiguration` event");
   if (hasConfigurationCapability) {
-    // Reset all cached document settings
+    getLogger().trace("Reset all cached document settings");
     clearSettings();
   } else {
     if (change.settings.UI5LanguageAssistant !== undefined) {
-      setGlobalSettings(change.settings.UI5LanguageAssistant);
+      const ui5LangAssistSettings = change.settings.UI5LanguageAssistant;
+      getLogger().trace("Reset all cached document settings", {
```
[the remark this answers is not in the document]
bd82: thanks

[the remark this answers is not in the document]
bd82: it get logged at line 250

packages/language-server/src/ui5-model.ts
bd82: print to the console will be included in this PR (hopefully tonight)

packages/language-server/src/ui5-model.ts
bd82: thanks

packages/settings/api.d.ts
bd82: copy / pasta mistake at 10pm 😄

packages/settings/test/settings-spec.ts
bd82: importing the `main` of the package.json
causes that file to be executed, thus it gets a coverage report.

Otherwise I got a coverage KPI error.
Regardless it is a good practice to import **Public** APIs the same way a **consumer** would during tests. It in effect creates a bit of both design and runtime "micro test".

Here is how the generated api.**js** looks like:
Note the use of Object.defineProperty (with a **function** in the getter).
So its not enough to load the file, all the exported properties must be **accessed**.

```javascript
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.validLoggingLevelValues = exports.validTraceServerValues = void 0;
var settings_1 = require("./settings");
Object.defineProperty(exports, "setGlobalSettings", { enumerable: true, get: function () { return settings_1.setGlobalSettings; } });
Object.defineProperty(exports, "setSettingsForDocument", { enumerable: true, get: function () { return settings_1.setSettingsForDocument; } });
Object.defineProperty(exports, "getSettingsForDocument", { enumerable: true, get: function () { return settings_1.getSettingsForDocument; } });
Object.defineProperty(exports, "hasSettingsForDocument", { enumerable: true, get: function () { return settings_1.hasSettingsForDocument; } });
Object.defineProperty(exports, "clearSettings", { enumerable: true, get: function () { return settings_1.clearSettings; } });
Object.defineProperty(exports, "clearDocumentSettings", { enumerable: true, get: function () { return settings_1.clearDocumentSettings; } });
Object.defineProperty(exports, "getDefaultSettings", { enumerable: true, get: function () { return settings_1.getDefaultSettings; } });
exports.validTraceServerValues = {
    off: true,
    messages: true,
    verbose: true,
};
exports.validLoggingLevelValues = {
    off: true,
    fatal: true,
    error: true,
    warn: true,
    info: true,
    debug: true,
    trace: true,
};
//# sourceMappingURL=api.js.map
```

packages/vscode-ui5-language-assistant/test/unit/.mocharc.js
bd82: Its a **plugin** that effects the chai **global** state.
If I put it in each test file it would only have the desired side effect the first time it is invoked.
And we would have to track which test use/does not use/no longer uses these plugin's assertions.

Yeah we only have 1-2 test files here, so it does not really matter, but I think its a better practice to manage the Chai plugins "**centrally**".

bd82:
@tal-sapan CR fixes are in:
- https://github.com/SAP/ui5-language-assistant/pull/348/commits/5fe38e7f2ecdd810a5a8abdaefee6291d31c5e98

Other comments I've replied to in-depth.

bd82:
The reduced coverage (0.04%) is due to missing tests (or hard to build?) for the manifest-handling logic.
- https://coveralls.io/builds/35974103/source?filename=packages%2Flanguage-server%2Fsrc%2Fmanifest-handling.ts#L7

packages/vscode-ui5-language-assistant/test/unit/.mocharc.js
tal-sapan: The reason it bothers me is that you are changing the state of chai by importing (reading) the mocha configuration file. It doesn't sound like it's an operation that should have side effects. I would prefer to add the `chai.use` line in every test file that uses these plugins.
But if you think this is the better approach I won't prevent merging it, it's not that critical.

packages/language-server/src/logger.ts
```diff
@@ -1,44 +1,72 @@
...
+const meta = require("../../package.json");
+
+export type ILogger = Omit<IChildLogger, "getChildLogger">;
+
+let logLevel: LogLevel = "error";
+/**
+ * We are using the VSCode Logging library right now as it is:
+ * 1. The only one of our logging libraries available on npmjs.com (currently).
+ * 2. Supports console logging which would be re-directed to the VSCode extension's output Channel
+ *    - Assuming this LSP server processes was spawned from the VSCode Extension
```
tal-sapan, line 16: Even if it wasn't spawned from the vscode extension, it will write the logs to the console which is the best we can do (at least until we print to a file)

packages/language-server/src/logger.ts
```diff
@@ -1,44 +1,72 @@
...
+  trace: buildSafeLoggerMethod("trace"),
 };
 
-let logger = NOOP_LOGGER;
 export function getLogger(): ILogger {
-  return NOOP_LOGGER;
+  return loggerWrapper;
+}
+
+const possibleSensitiveProps = { uri: true };
```
tal-sapan, line 46: Why is this an object instead of string array like before? You only use its keys

packages/language-server/test/logger-spec.ts
```diff
@@ -0,0 +1,73 @@
...
+import * as sinon from "sinon";
+import { getLogger, getLogLevel, setLogLevel } from "../src/logger";
+import { LogLevel } from "@vscode-logging/logger";
+
+describe("the Language Server Logger", () => {
+  let errorSpy;
+  let sandbox;
+
+  before(() => {
+    sandbox = sinon.createSandbox();
```
tal-sapan, line 11: I think you don't have to create a sandbox anymore since sinon has a default sandbox (you can call `sinon.spy`, `sinon.restore` etc instead of using a specific sandbox)

tal-sapan (commented):
Please add integration with the preferences for changing the log level

packages/language-server/src/logger.ts
bd82: not sure :)

packages/language-server/test/logger-spec.ts
bd82: Thanks

packages/language-server/package.json
```diff
@@ -24,6 +24,7 @@
     "node": ">=10.0.0"
   },
   "dependencies": {
+    "@vscode-logging/types": "0.1.3",
```
bd82, line 27: imports in `api.d.ts` need to appear in deps instead of dev-deps to enable end user consumption

packages/vscode-ui5-language-assistant/src/configuration.ts
```diff
@@ -0,0 +1,24 @@
+import { ExtensionContext, workspace } from "vscode";
+import { LanguageClient } from "vscode-languageclient";
+
+// TODO: can we get this from meta?
```
bd82, line 4: TBD create test

packages/vscode-ui5-language-assistant/src/configuration.ts
```diff
@@ -0,0 +1,30 @@
...
+  WorkspaceConfiguration,
+} from "vscode";
+import { CHANGE_LOG_LEVEL_REQUEST } from "@ui5-language-assistant/language-server";
+
+export const LOGGING_LEVEL_CONFIG_PROP = "UI5LanguageAssistant.logging.level";
+
+export function listenToLogLevelChanges<R>(opts: {
+  /* eslint-disable-next-line @typescript-eslint/no-explicit-any -- signature with `any` originates from VSCode APIs */
+  subscriptions: { dispose(): any }[];
+  /* eslint-disable-next-line @typescript-eslint/no-explicit-any -- signature with `any` originates from VSCode APIs */
```
tal-sapan, line 13: It might originate from vscode but you can still use `unknown` if you only send the parameter to the function

packages/vscode-ui5-language-assistant/src/configuration.ts
```diff
@@ -0,0 +1,30 @@
...
+  sendRequest: (method: string, param: any) => Promise<R>;
+  onDidChangeConfiguration: Event<ConfigurationChangeEvent>;
+  getConfiguration: (section?: string) => WorkspaceConfiguration;
+}): void {
+  opts.subscriptions.push(
+    opts.onDidChangeConfiguration((e) => {
```
tal-sapan, line 19: I wonder if this is necessary. We listen to the `onDidChangeConfiguration` event in the server, doesn't it cover this?

tal-sapan (commented):
I didn't review `configuration-spec.ts` because I think it might be unnecessary, please check (I added a comment about it)

packages/vscode-ui5-language-assistant/package.json
```diff
@@ -79,15 +79,15 @@
     "vscode": "^1.47.0"
   },
   "scripts": {
-    "ci": "npm-run-all clean compile lint test:unit coverage bundle package",
+    "ci": "npm-run-all clean compile lint coverage:* bundle package",
```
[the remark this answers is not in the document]
bd82: Yes it would be but currently the integration tests are disabled and run "empty".
So i made sure the `unit` ones are the "last" to run so we at least get the report for that.
Fixing the coverage (that never worked for the VSCode E2E tests) is outside the scope of this PR.

packages/vscode-ui5-language-assistant/package.json
```diff
@@ -79,15 +79,15 @@
...
     "clean": "rimraf ./dist ./lib ./coverage ./nyc_output *.vsix NOTICE LICENSE",
     "compile": "yarn run clean && tsc -p .",
     "compile:watch": "tsc -p . --watch",
     "lint": "eslint . --ext .ts --max-warnings=0 --ignore-path=../../.gitignore",
-    "test": "npm-run-all test:*",
-    "test:unit": "mocha --config=./test/unit/.mocharc.js",
     "test:integration": "node lib/test/run-integration-tests.js",
-    "coverage": "nyc node lib/test/run-integration-tests.js",
+    "test:unit": "mocha --config=./test/unit/.mocharc.js",
+    "coverage:integration:": "nyc node lib/test/run-integration-tests.js",
```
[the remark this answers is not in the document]
bd82: thanks

packages/vscode-ui5-language-assistant/src/configuration.ts
```diff
@@ -0,0 +1,30 @@
...
+  Event,
+  WorkspaceConfiguration,
+} from "vscode";
+import { CHANGE_LOG_LEVEL_REQUEST } from "@ui5-language-assistant/language-server";
+
+export const LOGGING_LEVEL_CONFIG_PROP = "UI5LanguageAssistant.logging.level";
+
+export function listenToLogLevelChanges<R>(opts: {
+  /* eslint-disable-next-line @typescript-eslint/no-explicit-any -- signature with `any` originates from VSCode APIs */
+  subscriptions: { dispose(): any }[];
```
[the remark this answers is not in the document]
bd82: I copied from the VSCode index.d.ts
- 
image

allaVolkov (approved):
reviewed 7 last commits

packages/vscode-ui5-language-assistant/src/configuration.ts
bd82: thanks

packages/vscode-ui5-language-assistant/src/configuration.ts
```diff
@@ -0,0 +1,30 @@
...
+}): void {
+  opts.subscriptions.push(
+    opts.onDidChangeConfiguration((e) => {
+      if (e.affectsConfiguration(LOGGING_LEVEL_CONFIG_PROP)) {
+        const newLogLevel = opts
+          .getConfiguration()
+          .get(LOGGING_LEVEL_CONFIG_PROP);
+        // no validation done here as the server performs such validation
+        // and will ignore invalid values.
+        opts.sendRequest(CHANGE_LOG_LEVEL_REQUEST, newLogLevel);
```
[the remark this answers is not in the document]
bd82: Yeah I am not interested in the rejection or resolving of the promise.
Can you add said eslint rule to this template too?
- https://github.com/SAP-samples/ecmascript-monorepo-template/blob/main/.eslintrc.js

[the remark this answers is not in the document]
bd82: will add void

packages/vscode-ui5-language-assistant/src/configuration.ts
bd82: it is not 😢 there goes my examples for DI in VSCode ext to avoid sinon mocking

bd82:
coveralls fails due to a file that had logging calls  added but has no coverage report, that is outside the scope of this PR to fix.
- https://coveralls.io/builds/36010081/source?filename=packages%2Flanguage-server%2Fsrc%2Fmanifest-handling.ts#L7

bd82:
I removed the "possible user information" filtering.
I don't think we should log anywhere centrally, only to the output channel / console.
So this makes this requirement redundant. 

Also it is very difficulty to be sure any and all logs are free from "potential user information" if we want to really leverage the
power of structured logging and add extra meta/info objects on the logs.

packages/vscode-ui5-language-assistant/package.json
```diff
@@ -33,6 +33,21 @@
...
+            "fatal",
+            "error",
+            "warn",
+            "info",
+            "debug",
+            "trace"
+          ],
+          "default": "error",
+          "description": "The verbosity of logging according to the following order: trace > debug > info > warn > error > fatal > off.",
+          "scope": "resource"
```
tal-sapan, line 49: The scope should be `"window"`. Resource-level scope means each workspace folder could have a different value, and when you access the configuration property you have to send a resource URI. The log level however is not in the scope of a resource/workspace folder, it's general for the entire extension (in the current running instance).

packages/language-server/src/server.ts
```diff
@@ -260,23 +259,31 @@ connection.onDidChangeConfiguration((change) => {
   if (hasConfigurationCapability) {
     getLogger().trace("Reset all cached document settings");
     clearSettings();
+    // intentionally, ignoring the results of this promise
+    // and avoiding using an async function as the API of the changeHandler is `synced`
+    // Note the change object appears to be empty of VSCode which is why we attempt to obtain
```
tal-sapan, line 264: Please check if after changing the scope to `window` you still need this call. Maybe it will be sent on the event.

packages/vscode-ui5-language-assistant/package.json
bd82: I am pretty sure we made this mistake in all our extensions :(

tal-sapan: Yes, probably because this is what the [example](https://github.com/SAP/vscode-logging/blob/33b09d75f368d34ce5d5fc92199928c270a21786/examples/extension/package.json#L39) says. But I fixed it in hana-explorer at least.

packages/language-server/src/server.ts
bd82: Still not sent on the event after fixing the scope.
- https://github.com/SAP/ui5-language-assistant/commit/a47abd00c72a16a7cce960ddce4ad0a400c4c236

tal-sapan: Try sending the section in `onInitializaed`
```typescript
    connection.client.register(
      DidChangeConfigurationNotification.type,
      {section: "UI5LanguageAssistant"}
    );
```
I checked it and it seems to be sending the configuration when I send it

bd82: Thanks.
- https://github.com/SAP/ui5-language-assistant/commit/0eaa2791e4316e5aaa2f264683f1124d3f85e3de

PR #413, opened 2021-06-07
feat: multiple ui5 versions support

bd82:


the-docta:
is there a way to vote for features?

petermuessig:
Implemented with #469 and #472

Issue #285, opened 2020-08-11
Auto complete doesn't work with corporate proxy


uxkjaer:
Hi,

Excellent to see SAP taking ownership of these extensions. However the extension doesn't work through corporate proxy, even though it has been added in VS code settings and windows environment variables.

See output below

c:\Users\KjaerJ\AppData\Roaming\Code\User\globalStorage\saposs.vscode-ui5-language-assistant\ui5-resources-cache\1.71.14 will be used to cache UI5 resources
[Error - 9:46:22 am] Request textDocument/hover failed.
Message: Request textDocument/hover failed with message: request to https://sapui5.hana.ondemand.com/1.71.14/test-resources/sap/ca/scfld/md/designtime/api.json failed, reason: connect ETIMEDOUT 23.48.30.95:443
Code: -32603
(node:13616) UnhandledPromiseRejectionWarning: FetchError: request to https://sapui5.hana.ondemand.com/1.71.14/test-resources/sap/ca/scfld/md/designtime/api.json failed, reason: connect ETIMEDOUT 23.48.30.95:443
at ClientRequest. (c:\Users\KjaerJ.vscode\extensions\saposs.vscode-ui5-language-assistant-1.5.1\node_modules@ui5-language-assistant\language-server\dist\server.js:48289:11)
at ClientRequest.emit (events.js:203:13)
at TLSSocket.socketErrorListener (_http_client.js:399:9)
at TLSSocket.emit (events.js:203:13)
at emitErrorNT (internal/streams/destroy.js:91:8)
at emitErrorAndCloseNT (internal/streams/destroy.js:59:3)
at processTicksAndRejections (internal/process/task_queues.js:77:11)
(node:13616) UnhandledPromiseRejectionWarning: FetchError: request to https://sapui5.hana.ondemand.com/1.71.14/test-resources/sap/ca/scfld/md/designtime/api.json failed, reason: connect ETIMEDOUT 23.48.30.95:443
at ClientRequest. (c:\Users\KjaerJ.vscode\extensions\saposs.vscode-ui5-language-assistant-1.5.1\node_modules@ui5-language-assistant\language-server\dist\server.js:48289:11)
at ClientRequest.emit (events.js:203:13)
at TLSSocket.socketErrorListener (_http_client.js:399:9)
at TLSSocket.emit (events.js:203:13)
at emitErrorNT (internal/streams/destroy.js:91:8)
[...]

PR #434, opened 2021-11-30
fix: replace node-fetch with axios

uxkjaer:
use axios instead of note-fetch to better support proxy

"re #285"


uxkjaer:
I've committed some code with another user appearantly. All tests are passing now, however they fail intermitently when running via the ci script. I think my node is struggling with the parallel processes. Would be good if you could take a look

uxkjaer:
found a better solution

PR #472, opened 2022-08-16
fix(language-server): detect the UI5 version from closest manifest.json

petermuessig:
To support mono-repos having different UI5 projects and to ensure that
the correct minUI5Version of the UI5 application/library is being used
the UI5 version will now be derived from the closest manifest.json of
the XMLView to be edited.

`manifest.json`:

```json
{
  [...]
  "sap.ui5": {
    "dependencies": {
      "minUI5Version": "${anySemVer}"
    }
  }
  [...]
}
```

The detection of the used framework has been removed so far and would
require an additional metadata in some UI5 application/library related
metadata. This can be re-added for future versions.

### DIFF
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -48,6 +48,7 @@
   },
   "devDependencies": {
     "@types/node-fetch": "2.5.10",
+    "@types/semver": "7.3.12",
     "@types/tmp": "0.2.0",
     "@ui5-language-assistant/semantic-model-types": "^3.2.0",
     "@ui5-language-assistant/test-utils": "^3.2.0",
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -41,7 +41,7 @@
     "fs-extra": "9.0.1",
     "lodash": "4.17.21",
     "node-fetch": "2.6.1",
-    "semver": "^7.3.7",
+    "semver": "7.3.7",
     "vscode-languageserver": "6.1.1",
     "vscode-languageserver-textdocument": "1.0.1",
     "vscode-uri": "2.1.2"
--- a/packages/language-server/package.json
+++ b/packages/language-server/package.json
@@ -41,10 +41,10 @@
     "fs-extra": "9.0.1",
     "lodash": "4.17.21",
     "node-fetch": "2.6.1",
+    "semver": "^7.3.7",
     "vscode-languageserver": "6.1.1",
     "vscode-languageserver-textdocument": "1.0.1",
-    "vscode-uri": "2.1.2",
-    "yaml": "1.10.2"
+    "vscode-uri": "2.1.2"
   },
   "devDependencies": {
     "@types/node-fetch": "2.5.10",
--- a/packages/language-server/src/manifest-handling.ts
+++ b/packages/language-server/src/manifest-handling.ts
@@ -7,7 +7,10 @@ import { FileChangeType } from "vscode-languageserver";
 import { getLogger } from "./logger";
 
 type AbsolutePath = string;
-type ManifestData = Record<AbsolutePath, { flexEnabled: boolean }>;
+type ManifestData = Record<
+  AbsolutePath,
+  { flexEnabled: boolean; minUI5Version: string }
+>;
 const manifestData: ManifestData = Object.create(null);
 
 export function isManifestDoc(uri: string): boolean {
@@ -22,13 +25,11 @@ export async function initializeManifestData(
   );
 
   const readManifestPromises = map(manifestDocuments, async (manifestDoc) => {
-    const isFlexEnabled = await readFlexEnabledFlagFromManifestFile(
-      manifestDoc
-    );
+    const response = await readManifestFile(manifestDoc);
 
     // Parsing of manifest.json failed because the file is invalid
-    if (isFlexEnabled !== "INVALID") {
-      manifestData[manifestDoc] = { flexEnabled: isFlexEnabled };
+    if (response !== "INVALID") {
+      manifestData[manifestDoc] = response;
     }
   });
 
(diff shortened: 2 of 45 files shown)


packages/language-server/package.json
```diff
@@ -41,10 +41,10 @@
     "fs-extra": "9.0.1",
     "lodash": "4.17.21",
     "node-fetch": "2.6.1",
+    "semver": "^7.3.7",
```
bd82, line 44: fixed version

packages/language-server/src/manifest-handling.ts
```diff
@@ -54,6 +55,26 @@ export function getFlexEnabledFlagForXMLFile(xmlPath: string): boolean {
...
+  const closestManifestPath = maxBy(
+    manifestFilesForCurrentFolder,
+    (manifestPath) => manifestPath.length
+  );
+
+  if (closestManifestPath === undefined) {
+    return undefined;
+  }
+
+  return manifestData[closestManifestPath].minUI5Version;
```
bd82, line 75: `minUI5Version`comes from the manifest and this is a user controlled file.

Do we want / need special handling / logging if the user did not specify `minUI5Version`? or just implicitly return undefined?

packages/language-server/package.json
petermuessig: Ahh damn

packages/language-server/src/manifest-handling.ts
bd82: what happens if the user specified an invalid minUi5Version?
e.g:
- 666

or one that is not supported by the semantic model (very old one?).

petermuessig: I tend for `undefined` because we default the version later and log the usage of the default version when the semantic model is being created. Otherwise we need to do the same for `flexEnabled` as well

packages/language-server/src/ui5-model.ts
```diff
@@ -21,95 +16,82 @@ import {
...
+  null
+);
+function createSemanticModelCacheKey(
+  framework: UI5Framework,
+  version: string
+): string {
+  return `${framework}:${version}`;
+}
+
 export async function getSemanticModel(
```
bd82, line 37: `getSemanticModel` reached 4 parameters.
is it time for a single `opts` param?

packages/language-server/src/server.ts
```diff
@@ -111,27 +103,33 @@ connection.onCompletion(
...
+      textDocumentPosition,
+    });
+
+    const documentUri = textDocumentPosition.textDocument.uri;
+    const document = documents.get(documentUri);
+    if (document) {
+      const documentPath = URI.parse(documentUri).fsPath;
+      const minUI5Version = getMinUI5VersionForXMLFile(documentPath);
+      const model = await getSemanticModel(
+        initializationOptions?.modelCachePath,
```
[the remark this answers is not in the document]
petermuessig: This is still the old logic - with the memory cache of the semantic model it will at least be read from memory during the runtime.

packages/language-server/src/ui5-model.ts
petermuessig: could do so

packages/language-server/src/ui5-model.ts
```diff
@@ -173,13 +155,14 @@ export async function getSemanticModelWithFetcher(
...
 
-  return generate({
+  const model = generate({
     version: version,
     libraries: jsonMap,
     typeNameFix: getTypeNameFix(),
     strict: false,
     printValidationErrors: false,
   });
+  return (semanticModelCache[key] = model);
```
bd82, line 165: this is subjective but I am not a fan of performing two actions:
- store cache
- return value

in the same statement, but maybe this is idiomatic style when dealing with caches?

packages/language-server/src/ui5-model.ts
petermuessig: Looking into the follow-up tasks, I would keep it for now - not that I am against it, but there are many places to be adopted then... 😓

packages/language-server/src/ui5-model.ts
petermuessig: If you prefer, I change this

packages/language-server/src/manifest-handling.ts
petermuessig: When an unsupported version is being used, then the model is empty and will be in memory cached. I can also verify this by reading the version information from CDN - I had this in mind but didn't do it so far...

petermuessig: Now, the version is being negotiated using the version mapping information:
https://ui5.sap.com/version.json

petermuessig:
Finally, the change is ready for review

packages/vscode-ui5-language-assistant/src/extension.ts
```diff
@@ -63,6 +69,36 @@ export async function activate(context: ExtensionContext): Promise<void> {
     clientOptions
   );
 
+  // keep a reference to the current UI5 version
+  let currentVersion;
+
+  // create and register the command to open the SAPUI5 demokit
+  context.subscriptions.push(
```
bd82, line 76: the activate method used to only handle the server process startup.
Now it also handles UI/UX concerns.

I would split these concerns to separate helper functions and avoid giving too much (direct) responsibility to the activate method.

packages/language-server/test/ui5-model-spec.ts
```diff
@@ -184,10 +213,278 @@ describe("the UI5 language assistant ui5 model", () => {
...
+
+    afterEach(async () => {
+      await cleanup();
+    });
+
+    it("resolve the default version", async () => {
+      expect(
+        await negotiateVersionWithFetcher(
+          async (url: string): Promise<FetchResponse> => {
+            if (url.endsWith("/version.json")) {
```
bd82, line 284: why does a test have a branching `if/else`?

packages/language-server/test/ui5-model-spec.ts
```diff
@@ -184,10 +213,278 @@ describe("the UI5 language assistant ui5 model", () => {
...
+              // request for version info (needs to be just an object => found!)
+              return createResponse(true, 200, versionInfo);
+            }
+          },
+          cachePath,
+          FRAMEWORK,
+          "1.105.0"
+        )
+      ).to.be.equal("1.105.0");
+      expect(
```
bd82, line 316: should not this be two separate tests? one for 105 and one for 104?

petermuessig: I don't mind on this

bd82 (approved):

packages/language-server/test/ui5-model-spec.ts
petermuessig: Using now 2 separate fetchers

packages/vscode-ui5-language-assistant/src/extension.ts
petermuessig: Ok, will be split in separate helpers to create the server and to create the statusbaritem

packages/language-server/test/ui5-model-spec.ts
petermuessig: Splitted

PR #486, opened 2022-08-25
ci: update cimg/go for ghr build

bd82:


marufrasully:
This is obselete. We have not github action in place.

PR #491, opened 2022-09-06
feat: add XML validation and quickfix for hardcoded UI text

fausto-m:
This is a validator that warns the user of hardcoded UI texts in XML views/fragments that could be externalized to a resource bundle. Includes a quick fix for the warning, where externalized strings from the project's i18n.properties file (if existent) may be suggested as replacements.

### DIFF
--- a/packages/language-server/src/commands.ts
+++ b/packages/language-server/src/commands.ts
@@ -3,6 +3,7 @@ import { commands } from "@ui5-language-assistant/user-facing-text";
 import {
   executeQuickFixStableIdCommand,
   executeQuickFixFileStableIdCommand,
+  executeQuickFixHardcodedI18nStringCommand,
 } from "./quick-fix";
 import { track } from "./swa";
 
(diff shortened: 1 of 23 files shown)


bd82:
@fausto-m the build matrix fails, can you please fix so I can start the review process.

@petermuessig WDYT of this feature?

fausto-m:
Hi @bd82, the build matrix issues have been solved now. Node16 failed due to something unrelated to my changes (timeout error from other test). Is there a way to rerun the test without additional commits? I've tried from CircleCI, but the option is disabled for me.

uxkjaer:
Hi @fausto-m,

I checked out this code and the validation is working fine. However I couldn't get the quick fix to work. Can you explain in a bit more detail how that works? Maybe we should add that to the readme.

I wonder if we should have a setting as well to enable or disable this validation, some projects are just in one language and it might seem as a nuisance for them to get these squiggly lines

voicis:
Hi, I have a few more points to add regarding this feature.

Similar feature is implemented for annotation files in [SAP Fiori Tools - XML Annotation Language Server](https://marketplace.visualstudio.com/items?itemName=SAPSE.sap-ux-annotation-modeler-extension) and I think it would make sense to keep the behaviour as similar as possible across the extensions, since users could be using both in the same project.

1. Properly resolving i18n model to its resource file - `manifest.json` contains information about the model names and where the resources are located at. Users can change the name of the i18n model or the path to the resource bundle and in that case current implementation may generate a code that actually doesn't work in runtime. Here is a sample fragment of `manifest.json` with the definition. This #523 PR might help with getting such information from the context.
```json
"sap.ui5": {
        "models": {
            "i18n": {
                "type": "sap.ui.model.resource.ResourceModel",
                "settings": {
                    "bundleName": "sap.fe.cap.travel.i18n.i18n"
                }
            }
      }
}
```
2. Current use case seems rather narrow, because it requires the translated text to exist which probably isn't the case most of the time. I would expect that a more common scenario would be one where you generate a new text (in i18n.properties file) for values that are not yet translated as well and save the user time for having to define one. 
3. I'm concerned that maintaining `ui5-user-facing-attributes.ts` could prove troublesome, because we need to track the changes done to control libraries. It would be good if we could get information which properties should be translated from the ui5 metadata, but it looks like currently it is missing.

PR #492, opened 2022-09-06
WIP: Refactor: replaced testing with Vitest

uxkjaer:
We can now test directly from TS files without compiling and Vitest also takes care of coverage. 

All tests executes in about 20 seconds, majority of this is from  packages/semantic-model/test/api-spec.ts as this downloads and tests multiple version of SAPUI5. This is done sequentially right now.


uxkjaer:
still having an issue with node version 14. But can see the vitest team are working on something as we speak, so I'll revisit this tomorrow.

PR #496, opened 2022-09-12
Feat: implement webview to the api docs as go to definition  

uxkjaer:


bd82:
Hi @uxkjaer 

I am a bit conflicted about this feature:

**Pros:**
- it is quite cool
- and may be useful.

Cons:
- It is not quite the regular "go to definition" behavior I am used to as it does not navigate to **source code**.
   It feels more like it is "go to docs" which is already implemented when you hover and then click on "more-information".
- So we have a bit of duplicate feature, (go-to-docs) which is implemented in two different ways (browser redirect vs webview inside vscode).

@petermuessig  WDYT?

I think I should also consult with the product person responsible for this.

uxkjaer:
My take on it. 
- youis  often needs to read the api and check something. 
- The hover scroll down and click isn't really useful for me. 
- we currently don't have any other useful use case for the go to definition in an XML view
- further possibilities are:
- to pick up where there cursor is placed ie event or aggregation and handle deep navigation
- implement a go to samples option as well. This could be maybe implementations.

I know we aren't navigating source code, but we don't have any other useful use case for these options. We could simply add our own as well and a different short cut. My idea was that you were already used to the f12 shortcut to go and read a definition.

bd82:
Good points I will setup a meeting with you and the relevant product person 👍

marianfoo:
Hi Jacob,
I think this is a good solution.
There is already a similar plugin from Jacek:
https://marketplace.visualstudio.com/items?itemName=jacek-wozniczak.vscode-ui5-api-reference
I tried this plugin and personally I prefer to have a separate browser window.
But that could also have to do with the fact that it was formatted differently here.

But you are right, I didn't know the function of "more information".

I think the use case of Jakob is a good one and can help nicely, even if I probably won't use it.
But maybe it is well implemented and I will use it :)

vobu:
just also wanted to bring up @wozjac's extension.
and there is also a `.docset` from @monavari-lebrecht at http://monavari.de/docset/ .
sooo...in best Ghostbuster's style 👻, why not combine the streams and ...
- incorporate @wozjac's extension here, into the language assistant itself (and award him eternal fame)
- provide an option to export to `.docset`, taking up @monavari-lebrecht's work
- have @uxkjaer's API ref in the webview as a "read more"-link of sorts, being available from the incorporated @wozjac's extension

uxkjaer:
We could also consider just using a custom menu option and command instead, if that makes more sense?

bd82:
> We could also consider just using a custom menu option and command instead, if that makes more sense?

Perhaps:
- Would it ever make sense to ever implement a go-to-definition that would open the source code of UI5?
- So if we keep it as currently implemented would it conflict with such a future feature?

I am generally in favor of this (or similar) feature, But it a major feature and I'd like to demonstrate it to the
relevant product person, but with the holidays season in Israel now, this will add some delay...

uxkjaer:
We have types for [openui5](https://www.npmjs.com/package/@openui5/Ts-types) and [Sapui5](https://www.npmjs.com/package/@sapui5/ts-types) so maybe navigate to those instead?

Happy to just add this as another menu option instead. Mainly I'm just keen to get a possibility not to leave vscode when needing to search the api. 
We could also have another one for going to samples.

Happy to discuss this with relevant people.

uxkjaer:
I've changed the shortcut and context menu to UI5-Language-Assistant: View API Reference. The go to definition no longer shows the api reference. I've also added a setting with options to either show API reference inside vscode or in browser.

packages/language-server/api.d.ts
```diff
@@ -25,6 +25,13 @@ export type ServerInitializationOptions = {
   logLevel?: LogLevel;
 };
 
+export function getNodeName(
```
bd82, line 28: why does the language server exposes a logical utility function?
It generally runs in a different process and communicates via JSON_RPC

packages/language-server/api.d.ts
```diff
@@ -25,6 +25,13 @@ export type ServerInitializationOptions = {
...
+  text: string,
```
bd82, line 29: with so many arguments, some of which are optional you should probably use a config object in this case.

packages/language-server/src/api.ts
```diff
@@ -19,3 +20,25 @@ export const SERVER_PATH: string = isDevelopmentRun
...
+): Promise<string | undefined> {
+  const ui5Node = await getUI5NodeName(
+    offsetAt,
+    text,
+    undefined,
+    cachePath,
+    framework,
+    ui5Version
+  );
+  return ui5Node
```
bd82, line 39: I think too much is done here in a single expression.
Can you split this up into multiple statements for readability?

packages/language-server/api.d.ts
bd82: Generally the language server should use logical utilities from other packages.
Maybe getNodeName should be extracted out of the language server sub-package.

packages/language-server/src/documentation.ts
```diff
@@ -108,3 +113,30 @@ export function getNodeDetail(node: BaseUI5Node): string {
       return node.name;
   }
 }
+
+export async function getUI5NodeName(
```
bd82, line 117: I am not sure this function belongs in this sub-package

packages/language-server/src/documentation.ts
```diff
@@ -108,3 +113,30 @@ export function getNodeDetail(node: BaseUI5Node): string {
...
+  offsetAt: number,
+  text: string,
+  model?: UI5SemanticModel,
+  cachePath?: string,
+  framework?: string,
+  ui5Version?: string
+): Promise<BaseUI5Node | undefined> {
```
bd82, line 124: with so many params and optional params a config object may be warranted

packages/language-server/src/documentation.ts
```diff
@@ -108,3 +113,30 @@ export function getNodeDetail(node: BaseUI5Node): string {
...
+    model = await fetchModel(cachePath, framework, ui5Version);
+  }
+  if (astPosition !== undefined) {
+    return findUI5HoverNodeAtOffset(astPosition, model);
+  }
+  return undefined;
+}
+
+async function fetchModel(cachePath, framework, ui5Version) {
+  const ui5Model = await getSemanticModel(cachePath, framework, ui5Version);
```
bd82, line 140: I do not recall:
Does every functional flow invoke `getSemanticModel` ?
Or does it happen once on language server startup?

Is this different or the same as other functional flows?

packages/language-server/src/documentation.ts
```diff
@@ -108,3 +113,30 @@ export function getNodeDetail(node: BaseUI5Node): string {
...
+export async function getUI5NodeName(
+  offsetAt: number,
+  text: string,
+  model?: UI5SemanticModel,
+  cachePath?: string,
+  framework?: string,
+  ui5Version?: string
+): Promise<BaseUI5Node | undefined> {
+  const documentText = text;
+  const { cst, tokenVector } = parse(documentText);
```
bd82, line 126: Again I do not recall.
Does every flow functional flow re-parses the text each time?
Or is there a cache of documents / CST?

packages/language-server/src/hover.ts
```diff
@@ -5,33 +5,30 @@ import {
...
-import { getNodeDocumentation, getNodeDetail } from "./documentation";
+import {
+  getNodeDocumentation,
+  getNodeDetail,
+  getUI5NodeName,
+} from "./documentation";
 import { track } from "./swa";
 
-export function getHoverResponse(
+export async function getHoverResponse(
```
bd82, line 19: could changing this to `async` break anything?

packages/language-server/src/server.ts
```diff
@@ -128,6 +130,7 @@ connection.onCompletion(
         minUI5Version
       );
       connection.sendNotification("UI5LanguageAssistant/ui5Model", {
+        cachePath: initializationOptions?.modelCachePath,
```
bd82, line 133: why does this feature require adding the cache path to the notifications?
Perhaps I am rusty in regards to this code base, but why was it not needed for other features, but needed here?

yarn.lock
```diff
@@ -8841,11 +8841,6 @@ tr46@^1.0.1:
   dependencies:
```
bd82, line 8841: why did the lock file change if no changes were made to the package.json?

packages/settings/api.d.ts
```diff
@@ -1,5 +1,8 @@
 import Thenable from "vscode-languageserver";
-export type Settings = CodeAssistSettings & TraceSettings & LoggingSettings;
+export type Settings = CodeAssistSettings &
+  TraceSettings &
+  LoggingSettings &
+  API_Reference;
```
bd82, line 5: The naming of `API_Reference` type does not seem consistent 
with the existing pattern in this code base / file

packages/settings/api.d.ts
```diff
@@ -14,12 +17,25 @@ export interface TraceSettings {
...
   messages: true;
   verbose: true;
 }
 
+export interface IValidApiReferenceValues {
+  editor: true;
+  browser: true;
+}
+
+export const validViewApiReferenceValues: IValidApiReferenceValues;
```
bd82, line 37: remove `view` from the name to have consistent `type` and `const` names?

packages/vscode-ui5-language-assistant/README.md
```diff
@@ -124,6 +124,16 @@ The feature is available in the following:
 
 ![](https://raw.githubusercontent.com/SAP/ui5-language-assistant/master/packages/vscode-ui5-language-assistant/resources/readme/preview-manifest-json.gif)
 
+### Quick navigation to API Reference
+
+Right click a tag in the XML file and use the View API reference shortcut to navigate directly to the API reference. Use the setting
```
bd82, line 129: Creating a demo GIF for this feature may be a good idea to "publicize" it.

packages/vscode-ui5-language-assistant/README.md
```diff
@@ -124,6 +124,16 @@ The feature is available in the following:
...
+
+```
+"UI5LanguageAssistant.view.API_Reference": "browser" | "editor"
+```
+
+The default setting is editor which opens the API reference in a new editor to the side.
```
bd82, line 135: clearer explanation needed.
- `editor` settings is described by the **same** word "editor"
- "browser" flow is not described.

packages/vscode-ui5-language-assistant/README.md
bd82, line 129: View API reference --> View API **R**eference

bd82:
Does the ui5 logo belong in this PR?

packages/language-server/test/documentation-spec.ts
```diff
@@ -89,5 +98,83 @@ describe("The @ui5-language-assistant/language-server <getNodeDocumentation> fun
...
+
+      const result = await getUI5NodeName(
+        document.offsetAt(position),
+        document.getText(),
+        undefined,
+        cachePath.path,
+        "SAPUI5",
+        ui5Model.version
+      );
+      if (result) expect(getNodeDetail(result)).to.equal("sap.m.Input");
```
bd82, line 157: why is the test none deterministic ?
why is an if / else needed?

packages/language-server/test/documentation-spec.ts
```diff
@@ -89,5 +98,83 @@ describe("The @ui5-language-assistant/language-server <getNodeDocumentation> fun
...
+      </Page>
+    </mvc:View>`;
+      const { document, position } = getXmlSnippet(xmlSnippet);
+
+      const result = await getUI5NodeName(
+        document.offsetAt(position),
+        document.getText(),
+        ui5SemanticModel
+      );
+      if (result) {
```
bd82, line 120: why is the test none deterministic ?
why is an if / else needed?

packages/language-server/test/documentation-spec.ts
```diff
@@ -89,5 +98,83 @@ describe("The @ui5-language-assistant/language-server <getNodeDocumentation> fun
...
+        expect(getNodeDetail(result)).to.equal("sap.m.Input");
+      } else {
+        expect(result).to.equal("sap.m.Input");
+      }
+    });
+
+    it("get the UI5 node name without a model", async () => {
+      const xmlSnippet = `<mvc:View displayBlock="true"
```
bd82, line 128: these snippets normally need a "⇶" symbol to mark the cursor position.
Why is this not needed here?

packages/vscode-ui5-language-assistant/src/constants.ts
```diff
@@ -1,2 +1,5 @@
 export const COMMAND_OPEN_DEMOKIT = "UI5LanguageAssistant.command.openDemokit";
+export const COMMAND_OPEN_WEBVIEW = "UI5LanguageAssistant.command.webView";
 export const LOGGING_LEVEL_CONFIG_PROP = "UI5LanguageAssistant.logging.level";
+export const VIEW_API_REF = "UI5LanguageAssistant.View.API_Reference";
+// export const REPLACE_GO_TO_DEFINITION = "editor.action.goToDeclaration";
```
bd82, line 5: remove new commented out code?

packages/vscode-ui5-language-assistant/src/extension.ts
```diff
@@ -125,6 +150,101 @@ function updateCurrentModel(model: UI5Model | undefined) {
   }
 }
 
+async function showWebView() {
```
bd82, line 153: this logic should be in a separate file.
The "extension.ts" responsibility is for initializing the language server.
new concerns should be in a separate file.

packages/vscode-ui5-language-assistant/src/extension.ts
```diff
@@ -125,6 +150,101 @@ function updateCurrentModel(model: UI5Model | undefined) {
...
+      document.offsetAt(window.activeTextEditor?.selection.active),
+      currentModel.cachePath,
+      currentModel.framework,
+      currentModel.version
+    ).catch((Error) => {
+      console.log(Error);
+      return undefined;
+    });
+  }
+  const ui5Url = `${new URL(apiRefUrl).href}${name ? "#/api/" + name : ""}`;
```
bd82, line 180: should not there be re-use in build the API ref link?
This is also done in the "more details" in hover feature...

packages/vscode-ui5-language-assistant/src/extension.ts
bd82, line 153: try to refactor this 70 lines function by extracting logical sub-parts
so the high level logical flow would be easily visible

bd82:
I think we need to review this in a "pair programing" session.
I get the feeling this PR has broken some assumptions on the code structure in this mono-repo.
But I am a bit rusty in regards to this project...

PR #498, opened 2022-09-12
feat: added turborepo for speed

uxkjaer:


bd82:
why are there so many source code changes for a **tool** introduction?

uxkjaer:
This is to come after #496, i think this PR has both of them.

bd82:
I don't think it is a good idea to change the build tools in a project that is not very actively maintained...

bd82:
I'll close this for now.
We can discuss it on a voice chat if you wish.

uxkjaer:
Well the plan is that I would pick up the maintenance and make it more regular. But I'm still not a maintainer.

PR #499, opened 2022-09-22
fix: code completion in default aggregation

voicis:
Inherited default aggregations were not taken into account when providing code completion for element names.

Added logic to find the first parent class with default aggregation and to use it (if it exists).


packages/xml-views-completion/src/providers/elementName/classes.ts
```diff
@@ -156,7 +156,19 @@ function handleDefaultAggregationScenario({
   parentXMLElement: XMLElement;
   parentUI5Class: UI5Class;
 }): classSuggestionContext | null {
-  const defaultAggregation = parentUI5Class.defaultAggregation;
+  let defaultAggregation: UI5Aggregation | undefined = undefined;
+
+  // handle inheritance
+  let currentClass: UI5Class | undefined = parentUI5Class;
+  while (currentClass) {
```
bd82, line 163: @petermuessig : This seems to make sense
Can you verify that this is the correct behavior in UI5?

bd82, line 163: @voicis 

Could we wrap this new logic in a function with a clear name and invoke said function?

e.g:
`searchForClosestDefaultAggregation" (better name?).

Basically when I read this function I first want to understand what we are trying to achieve (logical flow)
and only later (optionally) want to see **how** we implemented it.

PR #500, opened 2022-09-22
fix: duplicate code completion items

voicis:
Controls that override the inherited properties had duplicate completion items for attribute names e.g. https://ui5.sap.com/1.102.0/#/api/sap.fe.macros.FilterBar.

Changed `flattenMembers` function to return only unique members (by name).

### DIFF
--- a/packages/logic-utils/src/utils/flatten-members.ts
+++ b/packages/logic-utils/src/utils/flatten-members.ts
@@ -1,4 +1,4 @@
-import { partial } from "lodash";
+import { partial, filter, concat } from "lodash";
 import { UI5Class } from "@ui5-language-assistant/semantic-model-types";
 
 function flattenMembers<T extends { name: string }>(
@@ -10,13 +10,13 @@ function flattenMembers<T extends { name: string }>(
   if (ui5SuperClass !== undefined) {
     // UI5 SDK refers to inherited members (properties, events, aggregations. ...) as "borrowed" ...
     const borrowedMembers = flattenMembers(membersGetter, ui5SuperClass);
-    return directMembers.concat(
-      borrowedMembers.filter(
-        (borrowed) =>
-          directMembers.find((direct) => direct.name === borrowed.name) ===
-          undefined
-      )
+    const borrowedMembersWithoutOverrides = filter(
+      borrowedMembers,
+      (borrowed) =>
+        directMembers.find((direct) => direct.name === borrowed.name) ===
+        undefined
     );
+    return concat(directMembers, borrowedMembersWithoutOverrides);
   }
   return directMembers;
 }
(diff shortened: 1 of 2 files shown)


bd82:
@voicis please sign the CLA: https://github.com/SAP/ui5-language-assistant/pull/500#issuecomment-1254992500

packages/logic-utils/src/utils/flatten-members.ts
```diff
@@ -10,7 +10,13 @@ function flattenMembers<T>(
   if (ui5SuperClass !== undefined) {
     // UI5 SDK refers to inherited members (properties, events, aggregations. ...) as "borrowed" ...
     const borrowedMembers = flattenMembers(membersGetter, ui5SuperClass);
-    return directMembers.concat(borrowedMembers);
+    return directMembers.concat(
```
bd82, line 13: Could you please refactor this logic with two separate steps
1. **one** step to created "borrowedAndNotOverridden" (better name?) list and assign it to a **separate** variable..
2. **second** step to add this list to the result

Also in this project we tend to use `lodash` utilities and not built-ins (.find / .filter) even
if those built-ins exist and are safe to use.

This is to ensure consistent style of writing logic ([].find(() => {}) vs _.find([], () => {})
You can see example in the rest of the code (e.g in logic-utils)

packages/logic-utils/src/utils/flatten-members.ts
```diff
@@ -10,7 +10,13 @@ function flattenMembers<T>(
   if (ui5SuperClass !== undefined) {
     // UI5 SDK refers to inherited members (properties, events, aggregations. ...) as "borrowed" ...
     const borrowedMembers = flattenMembers(membersGetter, ui5SuperClass);
-    return directMembers.concat(borrowedMembers);
+    const borrowedMembersWithoutOverrides = filter(
```
bd82, line 13: Should we removed the borrowed that are also "overrides"
Or should we remove the original definitions that have been overridden?

Which is more correct? and does it even matter?

bd82 (approved):

packages/logic-utils/src/utils/flatten-members.ts
voicis: We should remove original definitions that have been overridden. There is also documentation information in the completion items, which may differ.

bd82: Are not we doing the opposite here by removing the "borrowed" && "overriden"
instead of "original" && "overriden"?

Or am I missing something here?

voicis: With "original" definition I meant the ones that are coming from the parent class, maybe my explanation was misleading. If we have the following example:
```Typescript
interface A {
   one: () => void;
   two: string;
}
interface B extends A {
   one: () => void;
   three: string;
}
```

When considering `B` the function `one` from `A` should be removed and the `one` from `B` should be used. The code here should be doing that as far as I can see.

PR #506, opened 2022-10-14
chore: introduce app context

mr25mr:
- app context only contains UI5 model and manifest details
- fix getting correct UI5 version on initial load
- place caching logic in a single file

### DIFF
--- a/packages/language-server/src/cache.ts
+++ b/packages/language-server/src/cache.ts
@@ -17,10 +17,10 @@ class ProjectCache {
   }
 
   /**
-   * Get all cached manifest details
+   * Get entries of cached manifest details
    */
-  getAllManifestDetails(): Map<AbsolutePathToManifest, ManifestDetails> {
-    return this._manifestDetails;
+  getManifestDetailsEntries(): string[] {
+    return [...this._manifestDetails.keys()];
   }
   getManifestDetails(documentPath: string): ManifestDetails | undefined {
     return this._manifestDetails.get(documentPath);
@@ -32,10 +32,10 @@ class ProjectCache {
     return this._manifestDetails.delete(documentPath);
   }
   /**
-   * Get all cached yaml details
+   * Get entries of cached yaml details
    */
-  getAllYamlDetails(): Map<AbsolutePathToYaml, YamlDetails> {
-    return this._ui5YamlDetails;
+  getYamlDetailsEntries(): string[] {
+    return [...this._ui5YamlDetails.keys()];
   }
   getYamlDetails(documentPath: string): YamlDetails | undefined {
     return this._ui5YamlDetails.get(documentPath);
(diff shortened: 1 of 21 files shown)


packages/language-server/src/manifest-handling.ts
```diff
@@ -135,6 +131,8 @@ async function readManifestFile(
   const flexEnabled = manifestJsonObject["sap.ui5"]?.flexEnabled;
   const minUI5Version =
     manifestJsonObject["sap.ui5"]?.dependencies?.minUI5Version;
-
-  return { flexEnabled: flexEnabled, minUI5Version: minUI5Version };
+  return {
+    flexEnabled: flexEnabled,
+    minUI5Version: minUI5Version,
+  };
```
UweSchwerk, line 137: Maybe revert if it is only a change in formatting?

yarn.lock
```diff
@@ -8841,11 +8841,6 @@ tr46@^1.0.1:
   dependencies:
     punycode "^2.1.0"
 
-tr46@~0.0.3:
-  version "0.0.3"
-  resolved "https://registry.yarnpkg.com/tr46/-/tr46-0.0.3.tgz#8184fd347dac9cdc185992f3a6622e14b9d9ab6a"
-  integrity sha512-N3WMsuqV66lT30CrXNbEjx4GEwlow3v6rr4mCcv6prnfwhS01rkgyFdjPNBYd9br7LpXV1+Emh01fHnq2Gdgrw==
-
```
UweSchwerk, line 8848: any explanation why this is removed from yarn.lock ?

UweSchwerk (commented):
overall this looks good to me (I was just checking code style, not functional correctness). 
Pls also sign Contributor License Agreement before removing draft flag.

packages/language-server/src/cache.ts
```diff
@@ -0,0 +1,70 @@
...
+  constructor() {
+    this._manifestDetails = new Map();
+    this._ui5YamlDetails = new Map();
+    this._ui5Model = new Map();
+  }
+
+  /**
+   * Get all cached manifest details
+   */
+  getAllManifestDetails(): Map<AbsolutePathToManifest, ManifestDetails> {
```
vadson71, line 22: I'm not sure we need all these getters and setters. 
They usually not needed unless we have some internal data transformation/validation inside the class.
You allow to access _manifestDetails directly and at the same time offer to update it via setManifestDetails method. IMHO it is fake security. Much simpler just to have public class member

packages/language-server/src/manifest-handling.ts
mr25mr: with current formatting and running ```format:validate```  shows every thing is compliant with Prettier code style.. So it should be fine,

packages/language-server/src/cache.ts
mr25mr: well we are not very much concern about security here, but anyway to to make private class members not to be publicly exposed, redefine it. Thanks anyway.

yarn.lock
mr25mr: well. It is managed by yarn and I am not sure if that is needed to be in yarn,lock file.

mr25mr:
covered in this PR https://github.com/SAP/ui5-language-assistant/pull/523

PR #523, opened 2022-11-21
feat: introduce context package

mr25mr:


packages/context/src/loader.ts
```diff
@@ -0,0 +1,252 @@
...
+  }
+  const manifest = await getUI5Manifest(manifestPath);
+  if (!manifest) {
+    return;
+  }
+  const manifestDetails = await getManifestDetails(documentPath);
+  if (!manifestDetails) {
+    return;
+  }
+  const projectInfo = await getProjectInfo(projectRoot);
```
vadson71, line 132: I'd suggest reject promise with specific result (code+description) instead of returning undefined in all above cases. This would help the consumer to understand where the problem actually is.

packages/context/src/loader.ts
```diff
@@ -0,0 +1,252 @@
...
+        service: service.name,
+        odataForeignKeys: true,
+        odataContainment: false,
+      });
+
+      services.set(service.urlPath, metadataContent);
+    }
+  } catch (error) {
+    console.log(error);
+    cache.deleteCapServices(projectRoot);
```
vadson71, line 186: Why you are deleting cache entry when it doesn't exist (in line 167 we already checked it)?
Why not to return rejected promise in case of exception? You are putting into the cache and return to consumer potentially incomplete services in case of exception.

packages/context/src/manifest.ts
```diff
@@ -0,0 +1,227 @@
...
+import { ManifestDetails } from "./types";
+import { Manifest, FileName } from "@sap-ux/project-access";
+import findUp from "find-up";
+import { findAppRoot } from "./utils";
+import { cache } from "./cache";
+
+export async function readManifestFile(
+  manifestUri: string
+): Promise<Manifest | undefined> {
+  const manifestContent = await readFile(
```
vadson71, line 16: Reading file may fail with exception which isn't captured with try block. Consumer should take care. Better to have a unified approach and reject promise with corresponding reason in all potential failures instead of returning resolved to undefined promise.

packages/context/src/watcher.ts
```diff
@@ -0,0 +1,193 @@
...
+  manifestUri: string,
+  changeType: FileChangeType
+): Promise<void> => {
+  console.debug("`reactOnManifestChange` function called", {
+    manifestUri,
+    changeType,
+  });
+  const manifestPath = URI.parse(manifestUri).fsPath;
+  switch (changeType) {
+    case 1: //created
```
vadson71, line 78: No reaction on creation? Since we handle manifest deletion, I'd expect handler for its creation

packages/context/src/watcher.ts
```diff
@@ -0,0 +1,193 @@
...
+  ui5YamlUri: string,
+  changeType: FileChangeType
+): Promise<void> => {
+  console.debug("`updateUI5YamlData` function called", {
+    ui5YamlUri,
+    changeType,
+  });
+  const ui5YamlPath = URI.parse(ui5YamlUri).fsPath;
+  switch (changeType) {
+    case 1: //created
```
vadson71, line 118: Same here

packages/context/src/manifest.ts
```diff
@@ -0,0 +1,227 @@
...
+ * Id of a `.view.xml` or `.fragment.xml` is calculated based on manifest app id and relative file from app root folder
+ */
+export const getCustomViewId = async (
+  documentPath: string
+): Promise<string> => {
+  const appRoot = await findAppRoot(documentPath);
+  if (!appRoot) {
+    return "";
+  }
+  const manifestPath = (await findManifestPath(documentPath)) as string;
```
voicis, line 211: Why do we need to cast here? We should avoid casting, because this can lead to runtime errors.

packages/context/src/types.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export const DEFAULT_UI5_FRAMEWORK = "SAPUI5";
+export const DEFAULT_UI5_VERSION = "1.71.49";
+export const UI5_FRAMEWORK_CDN_BASE_URL = {
+  OpenUI5: "https://sdk.openui5.org/",
+  SAPUI5: "https://ui5.sap.com/",
+};
+export const UI5_PROJECT_TYPE = "UI5";
+export const CAP_PROJECT_TYPE = "CAP";
+export enum DirName {
```
voicis, line 16: Perhaps this should be added to the open-ux-tools as this is needed for other extensions as well.

packages/context/src/ui5-yaml.ts
```diff
@@ -30,35 +21,35 @@ export async function initializeUI5YamlData(
...
     (ui5YamlPath) => ui5YamlPath.length
   );
-
-  return closestUI5YamlPath
-    ? ui5YamlData[closestUI5YamlPath].framework
-    : "SAPUI5";
+  if (closestUI5YamlPath) {
+    return cache.getYamlDetails(closestUI5YamlPath)?.framework ?? "OpenUI5";
+  }
+  return "OpenUI5";
```
voicis, line 47: Why did we change the default framework here?

packages/context/src/ui5-yaml.ts
```diff
@@ -30,35 +21,35 @@ export async function initializeUI5YamlData(
...
-    if (response !== "INVALID") {
-      ui5YamlData[ui5YamlDoc] = response;
+    if (response) {
+      cache.setYamlDetails(ui5YamlDoc, response);
+      console.info("ui5.yaml data initialized", { ui5YamlDoc });
     }
   });
 
-  getLogger().info("ui5.yaml data initialized", { ui5YamlDocuments });
+  console.info("list of ui5.yaml files", { ui5YamlDocuments });
```
voicis, line 30: We probably should keep the logger usage, because it is used to control log level via user setting.

packages/context/src/manifest.ts
mr25mr: wrap it in `try-cache`. On exception, log trace with a message. Regarding promise rejection, as said, should be done in a separate PR as per need.

packages/context/src/loader.ts
mr25mr: Thanks adapted.

packages/context/src/loader.ts
mr25mr: That is a good idea and further improvement maybe as per need.

packages/context/src/manifest.ts
```diff
@@ -0,0 +1,227 @@
...
+ * @param manifest manifest of an app
+ * @param serviceName name of data source
+ */
+export function getServicePath(
+  manifest: Manifest,
+  serviceName: string
+): string | undefined {
+  const dataSources = manifest["sap.app"]?.dataSources;
+
+  if (dataSources && serviceName !== undefined) {
```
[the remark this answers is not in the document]
mr25mr: copy paste effect. :)

packages/context/src/watcher.ts
mr25mr: in case of ui5 yaml, we simply empty cache on deletion. it is handled when next time user open an xml view. Meaning an yaml file is not found and default data is returned which is default framework and version = undefined.

packages/context/src/types.ts
mr25mr: yes. I will convey this message and once averrable in `open-ux-tools` we can clean up.

packages/context/src/ui5-yaml.ts
mr25mr: Thanks. We should not. :) This method is not used, so removed it completely. It is handled in `getYamlDetails` methods and default framework is `SAPUI5`

packages/context/src/ui5-yaml.ts
mr25mr: Thanks. Done

test-packages/framework/projects/cap/app/manage_travels/webapp/localService/metadata.xml
```diff
@@ -1,1820 +1,1392 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8" ?>
```
voicis, line 1: Do we need this file for a CDS project?

test-packages/framework/src/framework.ts
```diff
@@ -134,23 +134,32 @@ export class TestFramework implements TestFrameworkAPI {
   ): Promise<ReadFileResult> {
     const root = this.getProjectRoot();
     const filePath = join(root, ...pathSegments);
-    if (!fileExits(filePath)) {
+    if (!(await fileExits(filePath))) {
```
voicis, line 137: Instead of checking if file exists it is better to check file read failed, because these operations are async it could be that a file exists for the first call, but doesn't for the second and you end up with an unhandled error anyways. https://nodejs.org/api/fs.html#fsexistspath-callback

packages/context/src/watcher.ts
```diff
@@ -1,95 +1,74 @@
...
+    return;
+  }
+  // remove manifest cache
+  cache.deleteManifest(manifestPath);
+  const cachedProject = cache.getProject(projectRoot);
+  if (!cachedProject) {
+    return;
+  }
+  if (cachedProject.type == "CAP") {
+    for (const [, app] of cachedProject.apps) {
```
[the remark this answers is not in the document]
mr25mr: good point. adapted.

packages/context/src/watcher.ts
```diff
@@ -1,95 +1,74 @@
...
-      return;
+  const projectRoot = await getProjectRoot(manifestPath);
+  if (!projectRoot) {
+    return;
+  }
+  // remove manifest cache
+  cache.deleteManifest(manifestPath);
+  const cachedProject = cache.getProject(projectRoot);
+  if (!cachedProject) {
+    return;
```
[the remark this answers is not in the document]
mr25mr: I think no.

test-packages/framework/projects/cap/app/manage_travels/webapp/localService/metadata.xml
mr25mr: for cds project no, but needed for ui5 project. [Here](https://github.com/SAP/ui5-language-assistant/pull/523/files#diff-47b1cdd6a07cb38e18cc1ae6229e9bb5026db16f4cd01e8a5071676f9fb2a76fR48) use it for ui5 project.

packages/context/src/loader.ts
```diff
@@ -155,6 +159,12 @@ export async function getProject(
   return ui5Project;
 }
 
+const trimSeparator = (path: string): string => {
+  return path
+    .split(sep)
```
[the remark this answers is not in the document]
mr25mr: right. idea was to unify key. anyway adapted.

packages/vscode-ui5-language-assistant/src/extension.ts
```diff
@@ -82,6 +82,8 @@ function createLanguageClient(context: ExtensionContext): LanguageClient {
       fileEvents: [
         workspace.createFileSystemWatcher("**/manifest.json"),
         workspace.createFileSystemWatcher("**/ui5.yaml"),
+        workspace.createFileSystemWatcher("**/*.xml"),
```
petermuessig, line 85: maybe this can also be centralized in the `context` to just provide a map of file extensions next to the `watcher.js` in which the reaction to changes could also be moved to ?

petermuessig (commented):
Some minor comments, nothing critical...

packages/context/src/loader.ts
```diff
@@ -0,0 +1,261 @@
...
+async function getCapServices(
+  projectRoot: string
+): Promise<Map<string, string>> {
+  const cachedCapServices = cache.getCapServices(projectRoot);
+  if (cachedCapServices) {
+    return cachedCapServices;
+  }
+  const services = new Map<string, string>();
+  try {
+    const data = await getCapModelAndServices(projectRoot);
```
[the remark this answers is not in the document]
mr25mr: done.

packages/context/src/ui5-model.ts
```diff
@@ -38,11 +41,11 @@ export async function getSemanticModel(
   );
 }
 
-// cache the semantic model creation promise to ensure unique instances per version
-const semanticModelCache: Record<
-  string,
-  Promise<UI5SemanticModel>
-> = Object.create(null);
+const isUi5Model = (
```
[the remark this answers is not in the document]
mr25mr: done

packages/context/src/cache.ts
```diff
@@ -0,0 +1,133 @@
+import type { Manifest } from "@sap-ux/project-access";
+import type { UI5SemanticModel } from "@ui5-language-assistant/semantic-model-types";
+
+import type { App, Project, YamlDetails } from "./types";
+
+type AbsoluteAppRoot = string;
+type AbsoluteProjectRoot = string;
+
+class ProjectCache {
```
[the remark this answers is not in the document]
mr25mr: right. change it to `Cache`

packages/context/src/loader.ts
```diff
@@ -0,0 +1,257 @@
...
+    app,
+  };
+}
+
+/**
+ * Get CAP or UI5 project
+ *
+ * @param documentPath path to a file i.e absolute/path/webapp/ext/main/Main.view.xml
+ */
+export async function getProject(
```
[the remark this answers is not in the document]
mr25mr: yes for getting project.

packages/context/src/types.ts
```diff
@@ -0,0 +1,121 @@
+import type {
+  UI5SemanticModel,
+  UI5Framework,
+} from "@ui5-language-assistant/semantic-model-types";
+import { ConvertedMetadata } from "@sap-ux/vocabularies-types";
+import type { Manifest } from "@sap-ux/project-access";
+
+export const DEFAULT_UI5_FRAMEWORK = "SAPUI5";
+export const DEFAULT_UI5_VERSION = "1.71.49";
+export const UI5_FRAMEWORK_CDN_BASE_URL = {
```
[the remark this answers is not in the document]
mr25mr: well that requires some infestation and analysis ...

packages/context/src/utils/project.ts
```diff
@@ -0,0 +1,180 @@
...
+import { getPackageName } from "./package";
+
+const packageName = getPackageName();
+
+/**
+ * Unify service path by adding forward slash to beginning and end of a service path
+ *
+ * @param servicePath path to a service
+ */
+export const unifyServicePath = (servicePath: string): string => {
```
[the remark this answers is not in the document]
mr25mr: The issue is that sometimes a path in cds is defined as `/processor` but in manifest.json  as`/processor/` and we use this key to store and retrieve CAP service. This method simply unify it that we always get forward slashes. [See test](https://github.com/SAP/ui5-language-assistant/pull/523/files#diff-2e6ccb448b4211e2da2c815ff42529ea1b1208fc777c2b878ad390b8fa1c2cfbR117)
using `path.resolve` API does not unified it as far as I test it.

packages/context/src/watcher.ts
```diff
@@ -0,0 +1,255 @@
...
+ */
+export const reactOnManifestChange = async (
+  manifestUri: string,
+  changeType: FileChangeType
+): Promise<void> => {
+  getLogger(packageName).debug("`reactOnManifestChange` function called", {
+    manifestUri,
+    changeType,
+  });
+  const manifestPath = URI.parse(manifestUri).fsPath;
```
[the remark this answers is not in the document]
mr25mr: The idea is to have cache on function level which may have performance bottleneck like `getCAPServices` Though it is related to CAP project but handle individually this is because CAP can have multiple apps where services can be same.  This avoid double compilation. Maybe in the feature we can think of path being hierarchical or similar approach and give away performance over code readability. But that is a matter of a different PR.

packages/language-server/src/server.ts
```diff
@@ -191,19 +189,27 @@ connection.onHover(
 );
 
 connection.onDidChangeWatchedFiles(async (changeEvent) => {
-  getLogger().debug("`onDidChangeWatchedFiles` event", { changeEvent });
+  getLogger(packageName).debug("`onDidChangeWatchedFiles` event", {
+    changeEvent,
+  });
+  const cdsFileEvents: FileEvent[] = [];
```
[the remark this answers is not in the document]
mr25mr: currently all branches are handled from context package, but in the future there may come cases where we do not and logging here may make sense.

packages/context/src/cache.ts
```diff
@@ -0,0 +1,133 @@
...
+import type { App, Project, YamlDetails } from "./types";
+
+type AbsoluteAppRoot = string;
+type AbsoluteProjectRoot = string;
+
+class Cache {
+  private manifest: Map<AbsoluteAppRoot, Manifest>;
+  private app: Map<AbsoluteAppRoot, App>;
+  private project: Map<AbsoluteAppRoot, Project>;
+  private CAPServices: Map<AbsoluteProjectRoot, Map<string, string>>;
```
petermuessig, line 13: minor comment: I would avoid to change the case of variables starting with `CAP` and keep them lower case - otherwise it looks like a class!

`capServices` would be the better member name...

packages/context/src/loader.ts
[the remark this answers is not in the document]
petermuessig: Really? Still seems to be `Cap` here... But this is just a minor thing - it's mainly for you to be consistent...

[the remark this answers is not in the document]
petermuessig: Ah, ok this comes from an external NPM package... Ok, ignore it...

packages/context/src/utils/ui5.ts
```diff
@@ -1,12 +1,5 @@
 import { UI5Framework } from "@ui5-language-assistant/semantic-model-types";
```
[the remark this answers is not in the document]
petermuessig: The `version.json` is soon available for `OpenUI5`:
https://openui5nightly.hana.ondemand.com/version.json (nightly version) - will keep you posted...

petermuessig (approved):
In general, I'm OK with this change. So, you can go ahead...

PR #543, opened 2023-01-24
fix: make modules private

mr25mr:
All published projects [here](https://www.npmjs.com/search?q=%40ui5-language-assistant) will be first deprecated and will be removed finally in case it is feasible.


bd82:
1. This is not a "fix" (commit message header), look in https://www.conventionalcommits.org/en/v1.0.0/
    maybe this is a `chore` or `build` prefix.

2. Consider having this `private` more at the top of the file, [example](https://github.com/SAP/ui5-language-assistant/blob/master/test-packages/framework/package.json#L4) as this metadata (specifically when private) is very important.

mr25mr:
covered in this PR https://github.com/SAP/ui5-language-assistant/pull/546

PR #576, opened 2023-03-21
feat: enable offline mode

mr25mr:
Enable offline mode support for both XML views and manifest.json
* add (user / workspace) settings for SAPUI5 web server
* read content from SAPUI5 web server if available
* add text document content provider to support LSP for `mainfest.json`


voicis (commented):
If the web server config is not set and I'm offline the language server still crashes, which I don't think is the correct behaviour. We should handle these errors more gracefully without actually crashing the server.

Also with setting it crashes when trying to retrieve `https://sdk.openui5.org/1.111.00/resources/sap-ui-version.json` url.

We should document somewhere also the new setting and basic server configuration - what kind of requests it should support.

packages/logic-utils/package.json
```diff
@@ -20,7 +20,10 @@
     "@ui5-language-assistant/settings": "4.0.5",
     "@vscode-logging/logger": "1.2.2",
     "@xml-tools/ast": "5.0.0",
-    "lodash": "4.17.21"
+    "lodash": "4.17.21",
+    "node-fetch": "3.2.10",
+    "https-proxy-agent": "5.0.1",
```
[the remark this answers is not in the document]
mr25mr: they are moved from here https://github.com/SAP/ui5-language-assistant/pull/576/files#diff-15d790c3c6b4e4beab168603a33f16f2515c3f9b56d4695aca9871da084ceaf4L25 and were use only in `fetch` module / file. That file along its dependencies are moved.

packages/vscode-ui5-language-assistant/src/manifest/README.md
```diff
@@ -0,0 +1,3 @@
+# LSP for `manifest.json` file incase of offline mode
+
+To get complete LSP support for `manifest.json` file incase of offline mode, content of [schema](https://raw.githubusercontent.com/SAP/ui5-manifest/master/schema.json) _MUST_ be downloaded and saved in `manifest>schema.json` file. Furthermore, its content _MUST_ be scan to identify any `http(s)` references. In case any reference is found, its content _MUST_ be downloaded and saved in `manifest><any-reference>.json` and its reference _MUST_ be replaced by `"/manifest/<any-reference>.json"` in `manifest>schema.json`. Currently there is only one such case e.g `"https://adaptivecards.io/schemas/adaptive-card.json#/definitions/AdaptiveCard"` for which `adaptive-card.json` file is created and its content is downloaded and paste in that file. Moreover its reference in `manifest>schema.json` is replaced by `"/manifest/adaptive-card.json"`
```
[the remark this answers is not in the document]
mr25mr: well .....

mr25mr:
> If the web server config is not set and I'm offline the language server still crashes, which I don't think is the correct behaviour. We should handle these errors more gracefully without actually crashing the server.
> 
> Also with setting it crashes when trying to retrieve `https://sdk.openui5.org/1.111.00/resources/sap-ui-version.json` url.
> 
> We should document somewhere also the new setting and basic server configuration - what kind of requests it should support.

documentation is done in change long file https://github.com/SAP/ui5-language-assistant/pull/577/files?short_path=8e1ed70#diff-8e1ed702e9f0be5b0a84cd7f6145913bceac7ef53b054397c346e8c546102df9

packages/fe/test/services/utils.ts
```diff
@@ -52,7 +53,7 @@ export const getViewCompletionProvider = (
       content,
       offset
     );
-    const context = await getContext(documentPath);
+    const context = (await getContext(documentPath)) as Context;
```
voicis, line 56: I don't think its safe to assume that `context` is not going to of type `Error`.  If its an error then we'll have a runtime error when trying to work with a different type of object.

packages/fe/test/services/utils.ts
```diff
@@ -102,7 +103,7 @@ export const getViewValidator = (
       insertAfter: "<content>",
     });
     const { ast } = await framework.readFile(viewFilePathSegments);
-    const context = await getContext(documentPath);
+    const context = (await getContext(documentPath)) as Context;
```
voicis, line 106: Same as above.

packages/fe/test/services/utils.ts
mr25mr: yes, but in case of test, it is fine to make that assumption.

voicis (approved):

PR #581, opened 2023-03-24
fix: replace webpack with esbuild which solved crashing

mr25mr:
Replace `webpack` with `esbuild` which solved crashing of ui-language-assitance. Currently this solves the issue, but we should have a more close look at `esbuild`.


voicis (commented):
I checked the vsix and it works correctly also with `trace` logging enabled. Watch mode does not work, but we can fix it later. We should remove the webpack config files, otherwise the new config looks good.

PR #587, opened 2023-03-30
fix: environment for publishing

vadson71:
Environment variable added to npm deployment script


kranthie-sap (approved):
The change to the workflow looks good to me. It should skip the husky git hooks.

PR #594, opened 2023-04-14
fix: upgrade prettier

mr25mr:


vadson71 (approved):
Changes look reasonable

PR #608, opened 2023-05-24
fix: fallback to 1.71.49 for old supported UI5 versions

vadson71:


voicis (commented):
I tested this with 1.38 version in manifest.json and it works well. However, the tooltip message  of the version in the status bar is misleading - it says that version is not found in the manifest.json.

vadson71:
> I tested this with 1.38 version in manifest.json and it works well. However, the tooltip message of the version in the status bar is misleading - it says that version is not found in the manifest.json.

Ok. Thanks! I forgot about this message. Will adapt it

voicis (approved):

voicis (approved):

PR #610, opened 2023-05-26
feat: semantic model adaptation

vadson71:
POC of semantic model adaptation


voicis (commented):
We should have some test cases with code completion and diagnostics for this.

Issue #563, opened 2023-02-17
LSP for PropertyBindingInfo


MN-public:
Provide a minimal code completion and syntax check for the simple binding notation. This should include

- Code completion for property names in `PropertyBindingInfo`
- Validation for property names in `PropertyBindingInfo`
- Syntax highlighting for `PropertyBindingInfo` definition

Binding info object structure is defined by [sap.ui.base.ManagedObject.PropertyBindingInfo](https://sapui5.hana.ondemand.com/#/api/sap.ui.base.ManagedObject.PropertyBindingInfo)

MN-public:
Update: Code completion and diagnostics for property names provided. For more information and short "demo", see the last chapter here: https://blogs.sap.com/2023/07/04/sapui5-language-assistant-boosts-developer-productivity-with-building-blocks-offline-work-and-more/

MN-public:
Syntax highlighting for PropertyBindingInfo definition provided

PR #611, opened 2023-05-31
feat: code completion and diagnostics for property binding info

mr25mr:
This PR Provides: 

* Code completion PropertyBindingInfo
* Validation for  PropertyBindingInfo

Related issue: https://github.com/SAP/ui5-language-assistant/issues/563


packages/binding-parser/src/types/property-binding-info.ts
```diff
@@ -0,0 +1,150 @@
...
+export interface CollectionValue {
+  leftSquare?: LeftSquare;
+  elements: (PrimitiveValue | StructureValue)[];
+  range?: Range; // range which include left bracket, element and right bracket
+  rightSquare?: RightSquare;
+  commas?: Comma[];
+}
+export type Value = PrimitiveValue | StructureValue | CollectionValue;
+
+export interface AstElement {
```
voicis, line 115: The name is too generic and it is not really clear what it represents. Since the grammar is basically the same as for  JavaScript object literals and they are later used as such, then perhaps we can use the same names - objects and properties.

packages/language-server/src/xml-view-diagnostics.ts
```diff
@@ -41,63 +46,75 @@ export function getXMLViewDiagnostics(opts: {
...
-  }
->(
-  issues: (UI5XMLViewIssue | ExternalIssueType)[],
+function validationIssuesToLspDiagnostics<T extends ExternalIssueType>(
+  issues: (UI5XMLViewIssue | T)[],
   document: TextDocument,
-  isExternalIssue: (
-    issue: UI5XMLViewIssue | ExternalIssueType
-  ) => issue is ExternalIssueType
+  externalIssues: ((issue: UI5XMLViewIssue | T) => issue is T)[]
```
voicis, line 96: `externalIssues` is misleading and implies that it is an array of issues, but its actually an array of functions.

packages/language-server/src/xml-view-diagnostics.ts
```diff
@@ -142,7 +159,7 @@ function validationIssuesToLspDiagnostics<
         };
       /* istanbul ignore next - defensive programming */
       default:
-        assertNever(issueKind);
+        assertNever(issueKind as never);
```
voicis, line 162: The point of this line was to check if all the cases are handled and we shouldn't use type cast here.

packages/language-server/src/xml-view-diagnostics.ts
```diff
@@ -41,63 +46,75 @@ export function getXMLViewDiagnostics(opts: {
...
+  issues: (UI5XMLViewIssue | T)[],
   document: TextDocument,
-  isExternalIssue: (
-    issue: UI5XMLViewIssue | ExternalIssueType
-  ) => issue is ExternalIssueType
+  externalIssues: ((issue: UI5XMLViewIssue | T) => issue is T)[]
 ): Diagnostic[] {
   const diagnostics: Diagnostic[] = map(issues, (currIssue) => {
+    const range = isBindingIssue(currIssue)
+      ? currIssue.range!
```
voicis, line 100: Non-null assertions should be avoided.

packages/binding/src/types/issues.ts
```diff
@@ -0,0 +1,82 @@
...
+  | MissMatchValue
+  | NotAllowedProperty
+  | DuplicateProperty
+  | RequiredDependency
+  | RecursiveProperty
+  | MissingLeftCurly;
+
+interface BaseUI5XMLViewBindingIssue extends BaseUI5XMLViewIssue {
+  issueType: typeof BINDING_ISSUE_TYPE;
+  range?: Range;
```
voicis, line 25: Shouldn't this be a mandatory property?

packages/binding/src/types/issues.ts
```diff
@@ -0,0 +1,82 @@
...
+  | UnknownChar
+  | UnknownPropertyBindingInfo
+  | MissMatchValue
+  | NotAllowedProperty
+  | DuplicateProperty
+  | RequiredDependency
+  | RecursiveProperty
+  | MissingLeftCurly;
+
+interface BaseUI5XMLViewBindingIssue extends BaseUI5XMLViewIssue {
```
voicis, line 23: Since we have a different handling for external validation issues, do we really need to extend the base type? This forces us to convert back and forth between ranges and offsets which adds unnecessary coding.

voicis (commented):
#### Parser and AST

The benefit of having AST is that you don't need to have a all the grammar related information in the data structure, so it can be simpler and easier to work with. In current implementation I think it still has a bit too much grammar data which has no semantic meaning e.g brackets, whitespace etc.

The name `Property`Binding is use quite extensively here, but it might be too narrow as the syntax for `Aggregation` binding would be exactly the same, it would just have slightly different properties.


#### Validation

Currently there is a bit of a mix of having grammar based validation and semantic checks. It would be better if we can separate them and have all the grammar related checks done during parsing, then this would prevent grammar information bleeding into the AST.

packages/binding-parser/src/types/property-binding-info.ts
```diff
@@ -0,0 +1,150 @@
...
+  | typeof STRING_VALUE
+  | typeof NULL_VALUE
+  | typeof NUMBER_VALUE
+  | typeof COMMA
+  | typeof BOOLEAN_VALUE
+  | typeof PARSE_ERROR;
+
+export interface Base {
+  text: string;
+  range: Range;
```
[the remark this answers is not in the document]
mr25mr: Currently we do not have such use case. Maybe later, if needed, it should be adapted.

packages/binding/src/definition/definition.ts
```diff
@@ -0,0 +1,251 @@
+import type { PropertyBindingInfoElement } from "../types";
+export const propertyBindingInfoElements: PropertyBindingInfoElement[] = [
```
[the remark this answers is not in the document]
mr25mr: yes. We need to enhance 'semantic models' package for property definition. Will be done in a different PR.

packages/binding/src/services/completion/providers/property-binding-info.ts
```diff
@@ -0,0 +1,113 @@
...
+      continue;
+    }
+    for (const binding of ast.bindings) {
+      if (!isPropertyBindingInfo(text, binding)) {
+        continue;
+      }
+      completionItems.push(...getCompletionItems(context, binding, ast.spaces));
+    }
+  }
+  const nonDuplicateItems = completionItems.reduce(
```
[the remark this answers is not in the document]
mr25mr: well. Let's assume we have such binding syntax. `some text {} other text {} some here`. When we extract we build out of such syntax `[{}, {}]` an array of binding syntaxes. For first element we provided all completion items [added to completion items list], for second one same data is added [duplicate data]. I hade this situation only here, and had to remove duplicate entries. This could be also done before sending data to user on client.

packages/binding-parser/src/types/property-binding-info.ts
mr25mr: change it to `StructureValue` and `StructureElement`

packages/binding-parser/src/ast/binding-parser-visitor.ts
```diff
@@ -0,0 +1,201 @@
...
+      range,
+      commas,
+      type: "structure-value",
+    };
+  }
+  [OBJECT_ITEM](
+    node: ObjectItemChildren,
+    param: VisitorParam
+  ): StructureElement {
+    const key = this.createToken(KEY, param, node[KEY] ?? node[STRING_VALUE]);
```
voicis, line 93: Can't we handle the quotes already here? It's rather error prone to have to check this every time when you try to use the key value.

packages/binding/src/utils/element.ts
```diff
@@ -52,11 +52,30 @@ export const valueTypeMap = new Map([
...
+/**
+ * Change key from
+ * a. 'key'
+ * b. "key"
+ * c. &apos;key&apos;
+ * d. &quot;key&quot;
+ * to key without any quotes
+ */
+export const clearKey = (key = ""): string => {
+  return (key = key.split(/'|"|&apos;|&quot;/).filter((i) => !!i)[0] ?? "");
```
voicis, line 64: Why are we assigning to the local variable here? It has no effect.

packages/xml-views-validation/src/validators/document/non-unique-id.ts
```diff
@@ -55,7 +56,7 @@ function buildIssuesForSingleID(
           start: _.syntax.value.startOffset,
           end: _.syntax.value.endOffset,
         })),
-      };
+      } as NonUniqueIDIssue;
```
voicis, line 59: Use return type of the anonymous function instead of type cast.

voicis:
There also seems to be an issue with mismatching quotes e.g `{"abc: "xyz"}`, but I think is something that could be addressed in a followup PR. Error is not reported on UI and CST has two object items which seems wrong.

voicis (approved):

PR #619, opened 2023-06-15
fix: tab stop

mr25mr:


voicis (commented):
Tested different cases where completion items have enclosing symbols (quotes, curly brackets etc.) and the typing experience is now a lot smoother.

Other than the typo the PR looks ready to be merged.

voicis (approved):

PR #624, opened 2023-06-24
fix: cannot assign to read only property 'document' of object

mr25mr:


voicis (approved):
Tested and diagnostics are cleared correctly when file is closed.

PR #626, opened 2023-06-26
Fix/25141/semantic model

mr25mr:


packages/binding/src/definition/definition.ts
```diff
@@ -1,251 +1,167 @@
...
-    },
-  },
-  {
-    name: "targetType",
-    type: [
-      {
-        kind: "string",
+        default: {
+          values: [
+            "sap.ui.model.BindingMode.Default",
```
voicis, line 77: We have this information in the semantic model, I don't think we should hard code it here.

packages/binding/src/types/binding-info.ts
```diff
@@ -1,39 +1,50 @@
...
+  "formatOptions" = "formatOptions",
+}
 export interface Dependents {
   name: BindingInfoName;
   type: PropertyType[];
 }
 export interface PropertyType {
-  kind: TypeKind;
+  kind: string;
+  default?: {
```
voicis, line 32: `default` seems a bit misleading here. It's a list of "all" possible values, not the one that would be used in case there is no other value.

packages/binding/src/definition/definition.ts
```diff
@@ -1,251 +1,167 @@
...
+import { UI5TypedefProp } from "@ui5-language-assistant/semantic-model-types";
+import { PROPERTY_BINDING_INFO } from "../constant";
+import {
+  BindContext,
+  PropertyBindingInfoElement,
+  BindingInfoName,
+  PropertyType,
+  TypeKind,
+} from "../types";
+const buildType = (property: UI5TypedefProp): PropertyType[] => {
```
voicis, line 10: This part seems to be essentially the same - It's still is hardcoded against a specific list of properties.

packages/binding-parser/src/lexer/token.ts
```diff
@@ -60,7 +61,7 @@ const colon = createToken({
 });
 const stringValue = createToken({
   name: STRING_VALUE,
-  pattern: /('|"|&apos;|&quot;)(.*?)('|"|&apos;|&quot;)/,
+  pattern: /(?:'|"|&apos;|&quot;)(?:.*?)(?:'|"|&apos;|&quot;)/,
```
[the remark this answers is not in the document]
mr25mr: memory and performance.

packages/semantic-model/src/api.ts
```diff
@@ -41,6 +41,7 @@ export function generate({
   model.version = version;
   model.isFallback = isFallback;
   model.isIncorrectVersion = isIncorrectVersion;
+  addTypedefType(model);
```
voicis, line 44: Why couldn't we use similar logic to class properties?

packages/semantic-model/src/resolve.ts
```diff
@@ -298,3 +298,88 @@ const typesToIgnore: Record<string, undefined> = {
 function ignoreType(typeName: string): boolean {
   return has(typesToIgnore, typeName);
 }
+
+export function resolveTypeDefPropertyType(
```
voicis, line 302: This looks very similar to `resolveType`. We should have only one function for resolving types unless there is a good reason to have multiple. As far as I know all `type` strings should have the same structure and should be processed the same way.

packages/semantic-model-types/api.d.ts
```diff
@@ -78,6 +78,7 @@ export interface UI5Namespace extends BaseUI5Node {
 export interface UI5Typedef extends BaseUI5Node {
   kind: "UI5Typedef";
   properties: UI5TypedefProp[];
+  symbolProperties: (Ui5Property | ObjProperty)[];
```
voicis, line 81: We should avoid adding temporary data into the model, because it adds additional clutter and confusion which of the similar properties should be used.

voicis (approved):

Issue #637, opened 2023-07-18
Unknown property binding info on sap:label.


vanthax:
Using `<Label text="{oData>/SomeValue/#@sap:label} id="some_label" />` in an XMLView inside SAP BAS results in several problems showing up:
- Unknown property binding info
- Unkown character
- Missing comma
- Expect a value

This seems to be related to the colon character.

MN-public:
@vanthax  
Thanks for reporting the issue. 
The fix is prepared. You can check it before the official release to VSC and SAP Business Application Studio by installing the extension from https://github.com/SAP/ui5-language-assistant/releases/tag/v4.0.33

vanthax:
I wasn't able to reproduce the issue with the provided version. Thanks for quickly resolving the issue.

PR #639, opened 2023-07-19
fix: unjustified errors for property binding info

mr25mr:
fix unjustified error for following points.

* escape diagnostic for metadata binding that contains colon e.g `{/SomeValue/#@sap:label}`
    * consider also HTML equivalent e.g. `&#47;` or `&#x2F;`
* escape diagnostic for model that contains colon e.g https://github.com/SAP/ui5-language-assistant/issues/637
    *   consider also HTML equivalent `&gt;`

Refactoring
* move some logic to `binding-parser` package for reuse

### DIFF
--- a/.changeset/shiny-maps-guess.md
+++ b/.changeset/shiny-maps-guess.md
@@ -1,0 +1,8 @@
+---
+"@ui5-language-assistant/vscode-ui5-language-assistant-bas-ext": patch
+"vscode-ui5-language-assistant": patch
+"@ui5-language-assistant/binding-parser": patch
+"@ui5-language-assistant/binding": patch
+---
+
+fix unjustified errors for property binding info
--- a/packages/binding-parser/src/api.ts
+++ b/packages/binding-parser/src/api.ts
@@ -7,7 +7,11 @@ export {
   isBeforeAdjacentRange,
   positionContained,
   rangeContained,
-} from "./utils/position";
+  isBindingExpression,
+  isMetadataPath,
+  isModel,
+  isPropertyBindingInfo,
+} from "./utils";
 
 import {
   Value,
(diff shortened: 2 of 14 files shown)


voicis:
Wouldn't it be easier and safer to check if its simple binding (only model and binding path) via regex? UI5 also does something similar https://github.com/SAP/openui5/blob/master/src/sap.ui.core/src/sap/ui/base/BindingParser.js#L396. 

Checking it by looking at errors seems to be rather error prone, because you need to check many things to make sure that the errors are actually representing the particular case you're looking for and not something else.

The reported issue is no longer reproducible, however here are some examples that seems to be still wrong:
```{'path'>: ''}``` This should be handled as binding, because quotes are not allowed in simple binding. (currently no errors are shown)
```{oData>#@sap:label}``` We still complain about the relative paths with simple binding.

mr25mr:
> Wouldn't it be easier and safer to check if its simple binding (only model and binding path) via regex? UI5 also does something similar https://github.com/SAP/openui5/blob/master/src/sap.ui.core/src/sap/ui/base/BindingParser.js#L396.
> 
> Checking it by looking at errors seems to be rather error prone, because you need to check many things to make sure that the errors are actually representing the particular case you're looking for and not something else.
> 
> The reported issue is no longer reproducible, however here are some examples that seems to be still wrong: `{'path'>: ''}` This should be handled as binding, because quotes are not allowed in simple binding. (currently no errors are shown) `{oData>#@sap:label}` We still complain about the relative paths with simple binding.

Using regex makes life easier, but it also has its own weakness for our use case. That reg expression assumes that it always starts with  `{key:` [bracket, key and colon], which means for partially valid content, it may produce undesired result.   e.g  "{path '', suspended: false}"  Here colon is missing in first key and causing to ignore complete binding expression. Anyway, I enhance the logic which also covers cases you detected.

voicis (approved):
- Tested locally and the issues are no longer reproducible
- Wide range of cases are covered with tests (there seems to be an issue with SonarQube coverage reporting)

PR #683, opened 2023-12-12
fix: use logger instead of console.log

mr25mr:
This PR
* introduce a logger package for reuse
* adapt test to use latest patch set of fallback e.g `1.71.61`
* adapt latest patch of UI5 in test project
* rewrite tests

This PR  should stabilize pipeline and git ride of console spam.


vadson71 (approved):
Changes look good. Tested locally.

PR #686, opened 2023-12-22
feat: updated ui5 maintenance versions

vadson71:
Adaptation to the latest UI5 maintenance changes

### DIFF
--- a/packages/semantic-model/src/utils.ts
+++ b/packages/semantic-model/src/utils.ts
@@ -20,7 +20,7 @@ export function error(message: string, throwError: boolean): void {
   if (throwError) {
     throw new Error(message);
   } else {
-    // console.error(message);
+    console.error(message);
   }
 }
 
--- a/packages/semantic-model/src/utils.ts
+++ b/packages/semantic-model/src/utils.ts
@@ -20,7 +20,7 @@ export function error(message: string, throwError: boolean): void {
   if (throwError) {
     throw new Error(message);
   } else {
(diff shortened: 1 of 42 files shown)


packages/semantic-model/src/utils.ts
```diff
@@ -20,7 +20,7 @@ export function error(message: string, throwError: boolean): void {
   if (throwError) {
     throw new Error(message);
   } else {
-    console.error(message);
+    // console.error(message);
```
voicis, line 23: Should this be commented out?

packages/semantic-model/test/unit/api-negative.test.ts
```diff
@@ -17,12 +17,12 @@ describe("The ui5-language-assistant semantic model package API negative tests",
     ): () => UI5SemanticModel {
       const generateCall = (): UI5SemanticModel => {
         return generate({
-          version: "1.71.49",
+          version: "1.71.61",
```
voicis, line 20: Wouldn't it be better to have a constant for the version here as well?

vadson71: Created a constant

packages/semantic-model/src/utils.ts
vadson71: reverted

mr25mr (approved):
LGTM

voicis (approved):

PR #689, opened 2024-01-09
fix: no code completion or diagnostic for contextPath or metaPath of ...

marufrasully:
…macros namespace


packages/binding/src/services/completion/providers/binding.ts
```diff
@@ -70,64 +74,73 @@ export function bindingSuggestions({
...
-        binding,
-        ast.spaces,
-        ui5Aggregation,
-        propBinding
-      )
-    );
+    return completionItems;
+  } catch (error) {
+    getLogger().debug("bindingSuggestions failed:", error);
+    return completionItems;
```
vadson71, line 144: An exception may occur when some completionItems have already been collected. Would you like to return this incomplete list of items or is it better to return an empty list?

packages/binding/src/services/completion/providers/binding.ts
```diff
@@ -70,64 +74,73 @@ export function bindingSuggestions({
...
+    );
+    if (!ui5Property && !ui5Aggregation) {
+      return completionItems;
     }
-    if (!isBindingAllowed(text, binding, errors, properties)) {
-      continue;
+    const propBinding = getBindingElements(context, ui5Aggregation, false);
+    const properties = propBinding.map((i) => i.name);
+    const value = attribute.syntax.value;
+    const startChar = value && value.image.charAt(0);
```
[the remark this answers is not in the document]
marufrasully: using ? causes code coverage not to be considered.

packages/binding/src/services/completion/providers/binding.ts
marufrasully: it is very rare that such situation occur with current code base, but anyway,  I will return empty.

marufrasully: it is very rare that such situation occur with current code base, but anyway,  I will return empty.

vadson71 (approved):
Code looks good. Comments addressed.

PR #698, opened 2024-03-13
fix: exclude test-package from sonar reporting

marufrasully:


vadson71 (approved):

vadson71 (changes requested):
Changeset is missing. But seems like not required for sonar config changes.

vadson71 (approved):

PR #704, opened 2024-04-10
fix: use file uri for definition

marufrasully:


packages/xml-views-definition/test/unit/utils/file.test.ts
```diff
@@ -91,13 +80,9 @@ describe("file", () => {
...
       const namespace = "sap.ui.demo.walkthrough";
       const value = "sap.ui.demo.walkthrough.controller.Handler";
-      const expectedFileUri = join(
-        __dirname,
-        "..",
-        "data",
-        "controller",
-        "Handler.ts"
-      );
+      const expectedFileUri = URI.file(
```
vadson71, line 83: Would be nice to have some reusable function for this path transformation

vadson71 (approved):
Code changes look good. Excellent code coverage

vadson71 (approved):

Issue #707, opened 2024-04-22
Tooltip property validation 


mateusz-matyjaszczyk:
Hey,

The code is working perfectly fine but it is showing an error "Mandatory property path must be defined"
![CleanShot 2024-04-22 at 13 48 22](https://github.com/SAP/ui5-language-assistant/assets/26802754/ff3b5c9b-5f24-4ce8-baf2-829b2f28ceca)
![CleanShot 2024-04-22 at 13 48 31](https://github.com/SAP/ui5-language-assistant/assets/26802754/af931550-e0c5-4e60-b047-79bab1662085)

marufrasully:
Dear @mateusz-matyjaszczyk,

thanks for reporting this issue. The issue is fixed and can be test with `.vsix` from https://github.com/SAP/ui5-language-assistant/releases/tag/v4.0.61. This fix will be part of up coming release.

Thanks and best regards

MN-public:
Fix is provided, closing the issue

PR #708, opened 2024-04-22
fix: relax diagnostic for special property

marufrasully:
Issue: https://github.com/SAP/ui5-language-assistant/issues/707

## Description
Some property e.g `tooltip` can be used with both `aggregation binding info` or `property binding info`.  Aggregation binding in this case requires mandatory property e.g `path` where property binding requires no mandatory property. Application user can decide to use either `aggregation binding info` or `property binding info`.  This PR relax diagnostics check. If there is any element used, diagnostic for mandatory property is not checked.

### DIFF
--- a/packages/binding/src/services/diagnostics/validators/check-required.ts
+++ b/packages/binding/src/services/diagnostics/validators/check-required.ts
@@ -1,4 +1,4 @@
-import { findRange } from "../../../utils";
+import { findRange, getAltTypesPrime } from "../../../utils";
 import {
   BindingIssue,
   BINDING_ISSUE_TYPE,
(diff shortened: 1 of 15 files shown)


voicis (approved):

packages/binding/src/services/diagnostics/validators/check-required.ts
```diff
@@ -1,24 +1,32 @@
...
-    const usedRequiredEl = element.elements.find(
+    let usedRequiredEl = element.elements.find(
       /* istanbul ignore next */
       (i) => i.key?.text === reqEl.name
     );
+    const altTypes = getAltTypesPrime(aggregation);
+    if (!usedRequiredEl && altTypes) {
+      // some property e.g `tooltip` can be used with both `aggregation binding info` or `property binding info`. Therefore `altTypes` is defined in design time.
+      // if `altTypes` is present, check if any element is used. This is a very broad check to avoid false diagnostic.
+      usedRequiredEl = element.elements[0];
```
vadson71, line 28: Does elements array always contain at least one element?
Maybe better here just exit the validation function with empty result?

marufrasully: the exception is to check if there is at least one element present from either `aggregation property binding` or `property binding info`. If that element is not present, we still want to show diagnostic for mandatory property.

PR #710, opened 2024-04-25
fix: support cc and diagnostics for absolute path in meta path

marufrasully:


vadson71 (approved):
Code changes look good. Tested locally, works as expected. All changes covered in tests

Issue #646, opened 2023-08-07
enhancement: Generate ID's: make this unique for a whole project (~Flexibility Services)


jvanattedev:
If you do not have IDs for your UI5 components in the view, the UI5 Language Assistant allows you to generate IDs for the whole file. This occurs when "flexEnabled": true, in manifest.json. Unfortunately, this still generates errors in a UI5 project, since the IDs are unique in the file, but not across the whole project.

This results in errors such as:
`Error: Error: adding element with duplicate id 'application-app-tile-component---worklist--_IDGenText1'`

The request: is it possible to have the tools generate unique IDs across the whole project. This would be really useful for Flexibility Services.

marufrasully:
Dear @jvanattedev. Your request is implemented and can be tested using `.vsix` from https://github.com/SAP/ui5-language-assistant/releases/tag/v4.0.72 
Can you please check and close this issue if everything is working as you expecting?

Thanks and regards

jvanattedev:
Hi, I tested it with a small project. The ID's seems unique across different files now, from my test project:
- _IDGenTitle2
- _IDGenTitle1
- _IDGenLink1
- _IDGenLink

One more convenience would be to be able to generate them for the entire project in one go. Currently there are only 2 options in the Quick Fix menu:
- Generate a unique ID
- Generate unique IDs for the entire file

MN-public:
@jvanattedev Thanks for verifying and providing the feedback.
Please create new issue for each enhancement request. I would like to close this one now as implemented

PR #719, opened 2024-08-07
feat: Generate unique ID across view files in an application

marufrasully:
Issue: https://github.com/SAP/ui5-language-assistant/issues/646
dependent on this PR: https://github.com/SAP/ui5-language-assistant/pull/721


packages/language-server/test/unit/quick-fix.test.ts
```diff
@@ -33,6 +35,11 @@ describe("The @ui5-language-assistant/language-server diagnostics quick fix func
     appContext.manifestDetails.flexEnabled = true;
   });
 
+  afterEach(() => {
+    appContext = getDefaultContext(ui5SemanticModel);
```
voicis, line 39: This seems a bit problematic and error-prone, that tests leave a dirty environment after execution.

packages/xml-views-validation/src/validators/document/non-unique-id.ts
```diff
@@ -12,62 +12,75 @@ import {
...
+}
 
-export function validateNonUniqueID(xmlDoc: XMLDocument): NonUniqueIDIssue[] {
+export function validateNonUniqueID(
+  xmlDoc: XMLDocument,
+  context: Context
+): NonUniqueIDIssue[] {
   const idCollector = new IdsCollectorVisitor();
-  accept(xmlDoc, idCollector);
+  const files = Object.keys(context.viewFiles);
```
voicis, line 30: Do I understand this correctly that every time we validate a document we go through all the XML templates and collect all the IDs? This will have poor performance scaling, we should only do this at most once per diagnostic cycle. It is probably fine when editing a single file, but on start up with many open files we could see performance degradation.

packages/xml-views-quick-fix/src/quick-fix-stable-id.ts
```diff
@@ -18,12 +19,16 @@ export type QuickFixStableIdInfo = {
...
 
 export function computeQuickFixStableIdInfo(
-  xmlDoc: XMLDocument,
+  context: Context,
   errorOffset: OffsetRange[]
 ): QuickFixStableIdInfo[] {
   const biggestIdsByElementNameCollector =
     new BiggestIdsByElementNameCollector();
-  accept(xmlDoc, biggestIdsByElementNameCollector);
+  const files = Object.keys(context.viewFiles);
```
voicis, line 27: See the comment for `validateNonUniqueID`, the same applies here.

packages/xml-views-validation/src/validators/document/non-unique-id.ts
```diff
@@ -80,11 +93,19 @@ class IdsCollectorVisitor implements XMLAstVisitor {
       isPossibleCustomClass(attrib.parent)
     ) {
       if (this.idsToXMLElements[attrib.value] === undefined) {
-        // @ts-expect-error - TSC does not understand: `attrib.syntax.value !== undefined` is a type guard
-        this.idsToXMLElements[attrib.value] = [attrib];
+        this.idsToXMLElements[attrib.value] = [
+          {
+            // @ts-expect-error - TSC does not understand: `attrib.syntax.value !== undefined` is a type guard
```
voicis, line 98: Did you try assigning `this.idsToXMLElements[attrib.value]` to a local variable before the check? That usually works.

voicis (commented):
There seems to be a lot of `toHaveBeenCalledTimes` checks in tests, which makes the tests too focused on the implementation details and not actually testing the output. Majority of the times it adds very little value to the test and just makes it more difficult to refactor the tested function, because you will need to update the tests.

packages/context/src/utils/view-files.ts
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @param base base path to a folder or a file
+ * @param files xml files with its XMLDocument
+ * @returns xml files with its XMLDocument
+ */
+async function processViewFiles(
+  base: string,
+  files: Record<string, XMLDocument> = {}
+): Promise<Record<DocPath, XMLDocument>> {
+  return new Promise((resolve) => {
```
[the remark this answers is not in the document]
marufrasully: with a bit of modification, we can avoid it. I adapt it.

packages/context/src/utils/project.ts
```diff
@@ -175,3 +176,18 @@ export async function getLocalMetadataForService(
...
+ *
+ * The webapp folder contains the manifest.json file
+ *
+ * @param docPath document path of a file in current project
+ * @returns webapp path
+ */
+export const getWebappPath = function (docPath: string): string {
+  const pathSegments = docPath.split(DirName.Webapp);
+  pathSegments.pop();
+  const webappPath = join(...pathSegments, DirName.Webapp);
```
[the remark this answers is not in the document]
marufrasully: I get ride of webApp, we relays on`manifestPath` which already retrieved.

packages/context/src/utils/project.ts
```diff
@@ -175,3 +176,18 @@ export async function getLocalMetadataForService(
...
+
+/**
+ * Returns the path of the corresponding webapp folder for a given file
+ *
+ * The webapp folder contains the manifest.json file
+ *
+ * @param docPath document path of a file in current project
+ * @returns webapp path
+ */
+export const getWebappPath = function (docPath: string): string {
```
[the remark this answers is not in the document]
marufrasully: see bellow.

packages/context/test/unit/utils/project.test.ts
```diff
@@ -210,4 +211,28 @@ describe("project", () => {
...
+        "root",
+        "project",
+        "webapp",
+        "ext",
+        "my-test-file.view.xml"
+      );
+      // act
+      const result = getWebappPath(uri);
+      // assert
+      expect(join("root", "project", "webapp")).toStrictEqual(result);
```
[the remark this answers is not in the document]
marufrasully: I see. In essence it taste. Anyway, since method is removed no further action to take.

packages/context/src/watcher.ts
```diff
@@ -224,7 +226,18 @@ export const reactOnXmlFileChange = async (
...
   });
+
   const documentPath = URI.parse(uri).fsPath;
+  if (isXMLView(documentPath)) {
+    if (
+      changeType === FileChangeType.Created ||
+      changeType === FileChangeType.Deleted
+    ) {
+      // reset cached view files
+      cache.setViewFiles(getWebappPath(documentPath), {});
```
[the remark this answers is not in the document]
marufrasully: file has dependency to each other due to `id`, but yet we could invalidate a single file. 
Previous solution was to avoid micro programming. Furthermore these operations are not very often performed and building 105 xml files with roughly 240 lines of xml, takes less than a second. Anyway, now adapted to invalidate a single file.

packages/language-server/src/server.ts
```diff
@@ -321,6 +344,12 @@ documents.onDidChangeContent(async (changeEvent): Promise<void> => {
       );
       return;
     }
+
+    const documentText = document.getText();
+    const { cst, tokenVector } = parse(documentText);
+    const xmlDocAst = buildAst(cst as DocumentCstNode, tokenVector);
+    context.viewFiles[documentPath] = xmlDocAst;
```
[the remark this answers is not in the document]
marufrasully: good point. Side effect is not handled in `cache`. New methods e.g `setViewFile` and `setControlIdsForViewFile`are introduced to handle side effect

packages/language-server/src/server.ts
```diff
@@ -363,6 +393,12 @@ connection.onCodeAction(async (params) => {
     connection.sendNotification("UI5LanguageAssistant/context-error", context);
     return;
   }
+
+  // rebuild XMLDocument and assign it to viewFils cache. In quick fix, sometimes there is un-sync content in TextDocument and its XMLDocument version especially in case of QuickFix
+  const { cst, tokenVector } = parse(textDocument.getText());
```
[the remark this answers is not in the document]
marufrasully: see above.

packages/language-server/src/xml-view-diagnostics.ts
```diff
@@ -124,8 +120,8 @@ function baseDiagnostic(
           (_) => ({
             message: validations.NON_UNIQUE_ID_RELATED_INFO.msg,
             location: {
-              uri: document.uri,
-              range: offsetRangeToLSPRange(_, document),
+              uri: URI.file(_.documentPath).toString(),
```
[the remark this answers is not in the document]
marufrasully: yes we could. But till now we have been using `vscode-uri` to handle file uri. To clean up and get ride of dependency  (if needed), that should be done in a different PR.

packages/language-server/test/unit/quick-fix.test.ts
marufrasully: well. This is because we do manipulate `appContext` see line 91 
```typescript
appContext.viewFiles[appContext.documentPath] = ast;
```
after each is basically clean up that dirty env. We could avoid this by using `befreEach` instead of `beforeAll` but this `generateModel` seems to be a heavy operation.

packages/xml-views-validation/src/validators/document/non-unique-id.ts
marufrasully: yes. Performance wise did not notice any, but this can be optimized. Now control ids are part of context and reused in validation and quick fix.

packages/xml-views-validation/src/validators/document/non-unique-id.ts
marufrasully: seems obsolete after refactoring.

packages/xml-views-quick-fix/src/quick-fix-stable-id.ts
marufrasully: see above.

marufrasully:
> toHaveBeenCalledTimes

It depends what we want to check. Sometimes we are only interested in return signatures of a method e.g `getContext` In this case mocking make sense as underlaying methods have their respective unit tests.

packages/language-server/src/server.ts
```diff
@@ -361,16 +346,7 @@ documents.onDidChangeContent(async (changeEvent): Promise<void> => {
...
-    const diagnostics = getXMLViewDiagnostics({
-      document,
-      context,
-    });
-    getLogger().trace("computed diagnostics", { diagnostics });
-    connection.sendDiagnostics({
-      uri: changeEvent.document.uri,
-      diagnostics,
-    });
+    await validateOpenDocuments();
```
voicis, line 349: Do we really need to revalidate all documents on each change? We have different type of validators and currently we only have one that needs to be run for every change - unique ids. But we do not need to run other validators for every single open file.

packages/xml-views-validation/src/api.ts
```diff
@@ -42,6 +42,9 @@ export function validateXMLView<ExternalXMLViewIssue>(opts: {
   const validatorVisitor = new ValidatorVisitor(opts.context, opts.validators);
   accept(opts.xmlView, validatorVisitor);
   const issues = validatorVisitor.collectedIssues;
+
+  const nonUniqueIdIssues = validateNonUniqueID(opts.xmlView, opts.context);
```
voicis, line 46: We still revalidate all the documents multiple times here. We should only do it once.

packages/context/src/watcher.ts
```diff
@@ -265,6 +255,70 @@ export const reactOnXmlFileChange = async (
...
+  const documentPath = URI.parse(uri).fsPath;
+  if (!isXMLView(documentPath)) {
+    return;
+  }
+
+  const manifestPath = await findManifestPath(documentPath);
+  if (!manifestPath) {
+    return;
+  }
+  if (changeType === FileChangeType.Deleted) {
```
voicis, line 289: Both blocks look exactly the same, just with different operation. We can just pass the `operation` based on the `changeType`.

packages/context/src/utils/control-ids.ts
```diff
@@ -0,0 +1,73 @@
...
+  const allDocumentsIds = cache.getControlIds(manifestPath);
+  const keys = Object.keys(allDocumentsIds);
+
+  const mergedIds: Map<string, ControlIdLocation[]> = new Map();
+  for (const doc of keys) {
+    const ids = allDocumentsIds[doc];
+    for (const [id, location] of ids) {
+      const existing = mergedIds.get(id);
+      if (existing) {
+        mergedIds.set(id, [...existing, ...location]);
```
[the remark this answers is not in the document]
marufrasully: This avoid unexpected side effect. Pushing mutate array here we are overriding it.

packages/context/src/utils/ids-collector.ts
```diff
@@ -0,0 +1,56 @@
...
+      // @ts-expect-error - we already checked that xmlElement.name is not null
+      isPossibleCustomClass(attrib.parent)
+    ) {
+      const existing = this.ids.get(attrib.value);
+      const offsetRange = {
+        start: attrib.syntax.value?.startOffset ?? 0,
+        end: attrib.syntax.value?.endOffset ?? 0,
+      };
+      if (existing) {
+        this.ids.set(attrib.value, [
```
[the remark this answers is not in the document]
marufrasully: see below.

packages/language-server/src/server.ts
marufrasully: yes it is needed. The scenario changed with validation of unique ID across xml view files. Now we need to validate other opened documents as uniqueness of ID depends on other file. Now when user change content of xml view file, we need to collect diagnostics for all open documents and send these diagnostics to client for each xml document view. Here all validation should be involved as user may have other diagnostics apart from ID.  Short video illustrate basic follow


Here we change `id` of `Text` in `App.view.xml` file. This change affect `InvoiceList.view.xml` file and vise versa. Furthermore validation of `Button` e.g.`The "Button" class can't have an empty ID attribute when flexEnabled is "true" in manifest.json.` need to be collected which involves other validations.

packages/xml-views-validation/src/api.ts
marufrasully: We validate once per file. But now number files all opened documents. See above for more details.

voicis (approved):

PR #726, opened 2024-08-21
fix: correct change set

marufrasully:
In this PR https://github.com/SAP/ui5-language-assistant/pull/719 it looks like we had a comma in change set. Therefore there is a suspicion that pipe line build https://github.com/SAP/ui5-language-assistant/actions/runs/10490758568/job/29066687441 fails. This is because packages could not be incremented. With this PR which only include  correct change set file, it should fix that issue.


vadson71:
@marufrasully 
The damaged changeset file saved in the repo and will disturb the next build. In normal situation all changeset files are deleted and new release version is generated.
I'd suggest to correct the existing broken file instead of creating the new one

vadson71 (changes requested):
The damaged changeset file saved in the repo and will disturb the next build. In normal situation all changeset files are deleted and new release version is generated.
I'd suggest to correct the existing broken file instead of creating the new one

vadson71 (approved):
Comment addressed, Let's go.

Issue #739, opened 2024-12-19
Mandatory property "vSorterInfo" must be defined -  UI5 Language Assistant


AramSol:
Hi colleagues,

Recently, we encountered an error in the list binding related to the sorter, with the following message: `"Mandatory property 'vSorterInfo' must be defined."`
the Sorter it self work as expected

![Image](https://github.com/user-attachments/assets/9cfc8b39-b4a4-4021-bc3c-9b44721eed33)

MN-public:
@AramSol 
What is the minUI5Version set in the manifest of your project?
Best Regards
Mariana

AramSol:
Hi @AramSol ,

` "minUI5Version": "1.131.0"`

marufrasully:
@AramSol I tried to create a sample scenario to reproduce the issue. It works as expected. I attached project. Can you please adapt attached project to reproduce your issue? 
[sap.m.tutorial.walkthrough.38.zip](https://github.com/user-attachments/files/18258653/sap.m.tutorial.walkthrough.38.zip)

Thanks and best regards

AramSol:
@MN-public ,
I just test the provided example, and as soon as the minversion set to 1.131 the notification appeared.
![Image](https://github.com/user-attachments/assets/f4a876f5-4d75-44d4-b447-d2dea79d6f2d)

PR #741, opened 2025-01-02
fix: do not show vSorterInfo as completion item and show correct diagnostic

marufrasully:
Issue: https://github.com/SAP/ui5-language-assistant/issues/739


packages/binding/src/services/completion/providers/binding.ts
```diff
@@ -17,6 +17,7 @@ import {
 import { BindContext } from "../../../types";
 import { createInitialSnippet } from "./create-initial-snippet";
 import {
+  getAltTypesPrime,
```
voicis, line 20: This function name is very confusing. In general I would avoid using abbreviations especially if matches a commonly used word, because you do not always have the context to figure out what they mean.

packages/binding/src/definition/definition.ts
```diff
@@ -144,8 +194,17 @@ const getPossibleElement = (param: {
       if (!constParam.type) {
         continue;
       }
-      const reference = getReference(constParam.type);
       const name = sorterMap.get(constParam.name) ?? constParam.name;
+      if (name === "vSorterInfo" && constParam.parameterProperties) {
```
voicis, line 198: Having custom logic based on a name doesn't seem right. Why do we have problems handling `string|object` type? I suspect we would have similar issues elsewhere, because I think it is somewhat common pattern used in UI5.

marufrasully: The issue here is how sorter is initialized. In Javascript, user can pass `vSorterInfo` as param to constructor of Sorter. e.g 
```javascript
new sap.ui.model.Sorter(vSorterInfo)
```
where `vSorterInfo` can have properties e.g `comparator`. This syntax is fine in Javascript, but not ok in XML syntax. This `vSorterInfo` information is added to  [design time api.json ](https://ui5.sap.com/1.131.0/test-resources/sap/ui/core/designtime/api.json) which is the root cause of this issue. 

It would be ideal to have clean data to work with, but this is not always the case.

packages/binding/src/services/completion/providers/binding.ts
marufrasully: Done in this patch https://github.com/SAP/ui5-language-assistant/pull/741/commits/17c9fd1c8474f8d2f5f5d7046e37841f27e43a1b

packages/binding/src/definition/definition.ts
voicis: ok, I see. There is a specific logic how the constructor is called when the XML template is processed and it does not necessarily match the generic way how it can be called from JS.

voicis (approved):

PR #762, opened 2026-01-26
fix: manual latest release github action

marufrasully:
Currently when [github action](https://github.com/SAP/ui5-language-assistant/actions/workflows/deploy_maual.yaml) is executed, it fails at [publish step](https://github.com/SAP/ui5-language-assistant/actions/runs/21354577012).

The root cause seems to be missing `libnpmpack` dependency.

To verify:
execute this command `yarn release:publish --no-verify-access 2>&1 | head -50`  without installing `libnpmpack` dependency and re-run that command after installing `libnpmpack` dependency to see difference.

### DIFF
--- a/.changeset/rare-turkeys-think.md
+++ b/.changeset/rare-turkeys-think.md
@@ -1,0 +1,6 @@
+---
+"@ui5-language-assistant/vscode-ui5-language-assistant-bas-ext": patch
+"vscode-ui5-language-assistant": patch
+---
+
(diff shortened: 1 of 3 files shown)


Klaus-Keller (approved):
Thanks @marufrasully!
Adding `libnpmpack` module as devDependency looks good.

donal-tobin-sap (approved):
👍🏽 Looks good to me.

PR #763, opened 2026-01-26
fix: Lerna publish failure with tar 7.5.4

marufrasully:
## Problem
The manual deployment workflow was failing during the `lerna publish` step with the following error:
```
lerna ERR! TypeError: Cannot read properties of undefined (reading 'create')
lerna ERR!     at packDirectory (/home/runner/work/ui5-language-assistant/ui5-language-assistant/node_modules/lerna/dist/index.js:6429:39)
```

## Root Cause
- Lerna 9.0.3 was designed to work with tar 6.2.1, which exports `tar.default.create()`
- Our project uses tar 7.5.4 (upgraded for security reasons via `resolutions`)
- Tar 7.x introduced breaking API changes where `tar.default` no longer exists - only `tar.create()` is available
- This caused Lerna's `packDirectory` function to fail when trying to access `import_tar2.default.create`

## Solution
Applied a compatibility patch using `patch-package` that:
- Modifies Lerna's code to use `(import_tar2.default || import_tar2).create()` instead of `import_tar2.default.create()`
- This works with both tar 6.x (via `.default`) and tar 7.x (via direct export)
- Maintains the security upgrade to tar 7.5.4

## Locally verifying
```
yarn lerna publish from-package --no-push --no-git-tag-version --yes --loglevel=silly 2>&1 | tail -20
```


Klaus-Keller (approved):
Thanks @marufrasully!
Patch for tar module looks promising, hope it solves the issue reliably.

PR #765, opened 2026-02-05
Fix/tar security alert

marufrasully:
Issue: https://github.com/SAP/ui5-language-assistant/security/dependabot/152


donal-tobin-sap (approved):
👍🏽 Looks good to me.

PR #799, opened 2026-05-26
Fix/ci cd

marufrasully:
## Summary

Addresses security violations identified by **Orca** and **Zizmor** scanners across the CI/CD workflow files.

### `release.yaml`
- **Least-privilege permissions** — added `permissions: {}` at workflow level and `contents: read` per job; removed previously excessive `actions: write` / `actions: read`
- **Secret protection via environments** — `version` and `release` jobs now use `environment: production` so `ACCESS_PAT` is gated behind master-only deployment branches
- **Sonar decoupled from build** — removed inline `Run SonarCloud scan` step from the `build` job (which runs on PRs and cannot safely hold secrets); replaced with a `sonar-artifact` upload that feeds the standalone `sonar.yaml` workflow (PR #800)
- **Pinned action SHAs** — all `uses:` references pinned to full commit SHA with tag as comment
- **`persist-credentials: false`** — added to all `actions/checkout` steps to prevent credential leakage into artifacts (Zizmor Artipacked finding)
- **Removed `martinbeentjes/npm-get-version-action@main`** — replaced with an inline `node -p` shell command, eliminating an unpinned third-party action on `main`
- **Template injection fix** — moved `${{ steps.package-version.outputs.current-version }}` out of inline shell commands into an `env` variable
- **Modern `$GITHUB_OUTPUT` syntax** — replaced deprecated `::set-output` command

### `deploy_maual.yaml`
- **Least-privilege permissions** — added `permissions: {}` at workflow level and `contents: read` on `deploy-npm` job
- **Secret protection via environments** — `deploy-npm` job now uses `environment: production`
- **Pinned action SHAs** — all `uses:` references pinned to full commit SHA
- **`persist-credentials: false`** — added to `actions/checkout`

### `dependabot.yml`
- **Cooldown added** — `default-days: 7` / `semver-major-days: 30` to prevent dependency PR floods (Zizmor supply-chain finding)


donal-tobin-sap (approved):
No issue seen in this change.
Looks good to fix the OSPO compliance findings.

PR #821, opened 2026-06-24
fix: sanitize adaptive card schema URL

Artemis-Coder17:
#822 
Fixed untrusted schema warning in `manifest.json` by replacing the external `adaptivecards.io` reference with the bundled local schema after fetching from GitHub, aligning the online fetch path with the existing bundled fallback behaviour


marufrasully:
Thanks @Artemis-Coder17 

- Please align on reg  https://github.com/SAP/ui5-language-assistant/blob/master/packages/vscode-ui5-language-assistant/scripts/manifest/run-update.ts#L80. You may re-use it as function.
- Please add unit test for this case too
- Please add change file. It must include:
    - "vscode-ui5-language-assistant"
    - "@ui5-language-assistant/vscode-ui5-language-assistant-bas-ext"

Please also note that CI/CD pipline is broken due to remove of ui5 version where our tests depends on it.

Artemis-Coder17:
> Thanks @Artemis-Coder17
> 
> * Please align on reg  https://github.com/SAP/ui5-language-assistant/blob/master/packages/vscode-ui5-language-assistant/scripts/manifest/run-update.ts#L80. You may re-use it as function.
> * Please add unit test for this case too
> * Please add change file. It must include:
>   
>   * "vscode-ui5-language-assistant"
>   * "@ui5-language-assistant/vscode-ui5-language-assistant-bas-ext"
> 
> Please also note that CI/CD pipline is broken due to remove of ui5 version where our tests depends on it.

Thanks @marufrasully for the review

I've addressed all the review comments:

- **Regex alignment** — extracted the `adaptivecards.io` replacement into a shared `sanitizeAdaptiveCardUrl()` function in `src/utils/index.ts`, using the same regex as `run-update.ts`. Both files now call the shared function.
- **Unit tests** — added 3 tests for `sanitizeAdaptiveCardUrl` in `utils/index.test.ts` and 1 test covering the sanitization flow in `manifest-schema-provider.test.ts`.
- **Changeset Added** — Including both `vscode-ui5-language-assistant` and `@ui5-language-assistant/vscode-ui5-language-assistant-bas-ext`.

Regarding the CI/CD pipeline failure — as you noted, this is caused by the removal of a UI5 version that the `semantic-model` tests depend on.

marufrasully (commented):
@Artemis-Coder17 thanks.
- code looks good
- review comment is addressed

The pipeline is broken due to removal of UI5-version `"1.71` which our tests are based on. See https://github.com/SAP/ui5-language-assistant/blob/master/packages/constant/src/constant.ts#L9 we should adapt and use 1.120.  See this for available UI5 versions https://ui5.sap.com/versionoverview.html

Artemis-Coder17:
> @Artemis-Coder17 thanks.
> 
> * code looks good
> * review comment is addressed
> 
> The pipeline is broken due to removal of UI5-version `"1.71` which our tests are based on. See https://github.com/SAP/ui5-language-assistant/blob/master/packages/constant/src/constant.ts#L9 we should adapt and use 1.120. See this for available UI5 versions https://ui5.sap.com/versionoverview.html

@marufrasully Could I please get a re-review

I have updated the UI5 version from `1.71` to `1.120`. The tests are now adapted to use this version

.changeset/two-dodos-drum.md
```diff
@@ -0,0 +1,6 @@
+---
+"vscode-ui5-language-assistant": patch
```
marufrasully: please include all packages  which are modified in the change set.

packages/fe/test/unit/services/diagnostics/validators/unknown-annotation-target.test.ts
```diff
@@ -146,7 +146,6 @@ describe("contextPath attribute value validation", () => {
         `<macros:Form contextPath="/Travel"></macros:Form>`
       );
       expect(result.map((item) => issueToSnapshot(item))).toStrictEqual([
-        "kind: ContextPathBindingNotRecommended; text: Context path for Form is usually defined if binding for the object is different than that of the page; severity:info; offset:346-354",
```
marufrasully, line 149: why these are removed? @vadson71 fyi

marufrasully (changes requested):
See two comments.

packages/fe/test/unit/services/diagnostics/validators/unknown-annotation-target.test.ts
Artemis-Coder17: ## Why `ContextPathBindingNotRecommended` Was Removed From Tests

## Background

The `ContextPathBindingNotRecommended` info message is emitted by `unknown-annotation-target.ts` when the validator determines that `contextPath` is "not recommended" for a given control. The condition that triggers it is:

```javascript
if (expectedAnnotations.length + expectedTypes.length === 0) {
    isNotRecommended = true;
    result.push({ kind: "ContextPathBindingNotRecommended", /* ... */ });
}
```

In other words: if the UI5 model has no information about what `contextPath` should point to for this control, the validator assumes it's unusual and warns the user.

## What Changed

This PR bumps the default UI5 version from **1.71** to **1.120**.

- In **1.71**, the `contextPath` property on `sap.fe.macros.Form`, `sap.fe.macros.Field`, and `sap.fe.macros.Table` had no `ui5-metadata` block — meaning `expectedTypes` and `expectedAnnotations` were both empty. The validator fell back to the hardcoded spec, which also had no `contextPath` constraints for these controls, so the not-recommended message fired.

- In **1.120**, SAP updated the metadata. The `contextPath` property on these controls now carries a full `ui5-metadata` block. For example, `sap.fe.macros.Form.contextPath` in **1.120** contains:

    ```json
    "ui5-metadata": {
      "sap.fe": {
        "expectedAnnotations": [],
        "expectedTypes": [
          "EntitySet",
          "NavigationProperty",
          "Singleton",
          "EntityType"
        ]
      }
    }
    ```

The `expectedTypes` is now non-empty, so `expectedAnnotations.length + expectedTypes.length === 0` is false → `isNotRecommended` stays false → `ContextPathBindingNotRecommended` no longer fires.

## Investigation

1. **CI failed** with 4 tests in `unknown-annotation-target.test.ts` all complaining that `ContextPathBindingNotRecommended` was expected but not received.
2. We **traced the condition** in the source — `isNotRecommended` is only set when both `expectedAnnotations` and `expectedTypes` are empty.
3. `getPathConstraintsForControl` in `misc.ts` **checks property.metadata first** — if populated, it uses that instead of the hardcoded spec.
4. We **inspected the actual downloaded 1.120 model file** at `test-packages/test-utils/.model-cache/1.120.48/sap.fe.macros.designtime.api.json` and confirmed that `sap.fe.macros.Form.contextPath` now has `expectedTypes: ["EntitySet", "NavigationProperty", "Singleton", "EntityType"]`.

.changeset/two-dodos-drum.md
Artemis-Coder17: @marufrasully Thanks for the review, I have updated the changeset

marufrasully (approved):
@Artemis-Coder17 thanks
- changes looks good to me
- review comment addressed
- did not verify it locally

@Artemis-Coder17 Please verify (both manifest + annotation macro part) final artifact `.vsix` when this PR is merged. You can find final artifact here https://github.com/SAP/ui5-language-assistant/tags

Artemis-Coder17:
### Before Fix
image

### Tested with Latest VSIX
image

